PR #12, opened 2018-06-04
Fix typo in upgrade manual

Rukenshia:
By the way, I think this should not have been a patch version. My build process just broke :) Appreciate the namespace changes though.


TimonPost:
Thanks for the contribution

PR #29, opened 2018-08-28
Fix set_size for ansi terminals

iCiaran:
Height should come first in the escape code as shown [here](https://apple.stackexchange.com/questions/33736/can-a-terminal-window-be-resized-with-a-terminal-command) or more formally [here](http://invisible-island.net/xterm/ctlseqs/ctlseqs.html) (ctrl+f for `CSI Ps ; Ps ; Ps t` and scroll down to `Ps = 8`).


TimonPost:
Thanks if this is really wrong I can't believe I have never noticed it, going to check it out tomorrow (wensday) and merge. Thanks !

PR #31, opened 2018-09-16
Fix `AnsiOutput` write methods claiming zero bytes were written

hawkw:
Currently, the implementation of `IStdout` for `AnsiOutput`always returns `Ok(0)` when a write completes successfully. 

This behaviour is incorrect, as these methods are used elsewhere to implement `std::io::Write`, which [expects  the writer to return the number of bytes written](https://doc.rust-lang.org/std/io/trait.Write.html#tymethod.write). This means that the implementation of `Write` for `crossterm::Terminal` doesn't play nice with the `write!` macro (and other consumers of the `Write` trait), causing uses of `write!` to panic with the message "failed to write whole buffer". (see mgattozzi/ysh#12).

This branch changes these methods to correctly return the number of bytes written, fixing the panics.

### DIFF
--- a/src/modules/output/ansi_output.rs
+++ b/src/modules/output/ansi_output.rs
@@ -14,25 +14,23 @@ pub struct AnsiOutput {
 
 impl IStdout for AnsiOutput {
    fn write_str(&self, string: &str) -> io::Result<usize> {
-       let out = &self.handle;
-       let mut handle = out.lock();
-        write!(handle, "{}", string)?;
-       handle.flush();
-        Ok(0)
+        let out = &self.handle;
+        let mut handle = out.lock();
+        let amt = handle.write(string.as_bytes())?;
+        handle.flush()?;
+        Ok(amt)
     }
 
     fn write(&self, buf: &[u8]) -> io::Result<usize> {
         let out = &self.handle;
         let mut handle = out.lock();
-        handle.write(buf)?;
-        Ok(0)
+        handle.write(buf)
     }
 
     fn flush(&self) -> io::Result<()> {
         let out = &self.handle;
         let mut handle = out.lock();
-        handle.flush();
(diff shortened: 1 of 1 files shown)


src/modules/output/ansi_output.rs
```diff
@@ -14,25 +14,23 @@ pub struct AnsiOutput {
 
 impl IStdout for AnsiOutput {
    fn write_str(&self, string: &str) -> io::Result<usize> {
-       let out = &self.handle;
-       let mut handle = out.lock();
-        write!(handle, "{}", string)?;
-       handle.flush();
-        Ok(0)
+        let out = &self.handle;
```
jojolepro: For some reason there are 5 spaces instead of 4 now.

TimonPost:
Thanks, my fault :( ! I'll merge it an do a release. 

I have on question could the release wait for a couple of days? This because there maybe come another PR so I could have two patches in one release. If that doesn't work out let me know!

You could then use the GitHub repo with master branch in your cargo.toml

src/modules/output/ansi_output.rs
peddermaster2: no, the line 16 has only 3 spaces so it looks like 5.

TimonPost: Thanks for reviewing! I'll cargo fmt and all spaces and tabs will be right ;)

hawkw: Sorry about that, the original function was missing one space of indentation, so I tried to fix it, but I think I missed line 16. My bad.

TimonPost: Don't worry already fixed it! But thanks for your pull request!

TimonPost:
I published the new version with the changes.

PR #35, opened 2018-10-05
clean up

TimonPost:
clean up

### DIFF
--- a/examples/cursor/mod.rs
+++ b/examples/cursor/mod.rs
@@ -21,7 +21,7 @@ pub fn pos() {
     let screen = Screen::default();
     let mut cursor = cursor(&screen);
     // get the cursor position.
-    let (_x, _y) = cursor.pos();
+    let (x, y) = cursor.pos();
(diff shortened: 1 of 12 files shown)


TimonPost:
I ran `cargo fix` and this PR was the result. I agree that you could have the name there since we than know that those variables exists but they could be removed. @jojolepro have you any specific reason why having `_` is better.  I once heard that "when using `_` no allocation" is made you know if that statement is right?

jojolepro:
Since those are in example files, _name makes more sense. If it was in crossterm's implementation then _ could fit better. Its mostly preferences though.

TimonPost:
@jojolepro A I see indeed it is just preference I removed the `_` before variables in the examples. In the lib I left it as it is.

TimonPost:
Yea I get your point. It looks cleaner without `_` maybe if I do `#![allow(dead_code)]` for the examples it will be fixed. And I can have normal variable names.

PR #61, opened 2018-12-16
Make example less confusing by using the described colours.

Jezza:
The main example said it was printing out 'Red Text on a White background', but the whole thing was green.
I thought the terminal encoding was broken...


TimonPost:
This file is not directly meant to have example code in. So no need to update it. It is more of a local playground to test your made changes or the crate.

PR #62, opened 2018-12-16
Implement support for some basic attributes for the windows terminal.

### LINKED ISSUE #60: Basic attribute support for Windows systems.
AFAIK, the windows terminal does support attributes, such as bold, but it doesn't support a lot of *nix ones.

https://docs.microsoft.com/en-us/windows/console/console-virtual-terminal-sequences

Is it a good idea to see if we can implement partial attributes?

Jezza:
Closes #60.

### DIFF
--- a/examples/color/mod.rs
+++ b/examples/color/mod.rs
@@ -207,6 +207,15 @@ pub fn print_font_with_attributes() {
     println!("{}", style("Crossed out font").crossed_out());
 }
 
+/// Print font with all available attributes. Note that this can only be used at unix systems and that some are not supported widely | demonstration..
+#[cfg(windows)]
+pub fn print_font_with_attributes() {
+    println!("{}", style("Normal text"));
+    println!("{}", style("Bold text").bold());
+    println!("{}", style("Underlined text").underlined());
+    println!("{}", style("Negative text").negative());
+}
+
 /// Print all supported RGB colors  | demonstration.
 #[cfg(unix)]
 pub fn print_supported_colors() {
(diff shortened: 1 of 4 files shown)


Jezza:
You'll probably notice that it doesn't have the full list of values that windows supports as methods, and that's because of how crossterm constructs styles and what the other windows' values are used for.

The values I've omitted, but added explicit members in the enum for are "reset" values, and due to how the styles are constructed, it you can't really construct a line that contains multiple "on"s and "off"s.
Eg, you can't say `style("Introducing ").bold().text("Me").reset().text(" and I.")` which would output something like "Introducing **Me** and I."
The ObjectStyles in crossterm don't seem to be built that way, so it makes no sense to add explicit support for the values as methods.
The values themselves are there in the enum, so you could conceivably, chain together multiple object styles, printing them out one by one, and then in that case, the reset would come in handy, but at that point, it might be easier just to wrap that with your own thing....

Jezza:
Look back, I think a better solution to this might be doing what Std does.  
Have system specific modules, and then just export those based on the system.  
That way the user can still access attributes for other systems, as it is possible that someone can use this library for sending terminal seqs over the wire.

Jezza:
Yeah, but do it in a controlled fashion.  

`format!("{}Inverted{}Normal", Attribute::Inverted, Attribute::Reset)`

It doesn't stop people from being stupid, but it didn't do that before.
It just makes composing things easier.

TimonPost:
Jup that totally does it, however, for now, I am fine whit those changes. This should become another PR.

PR #78, opened 2019-01-13
feat: make error module public

fdehau:
With the recent introduction of the explicit error handling #69, `Result` and `ErrorKind` types where added as part of the signature of some functions of the public api. In order to use this crate within an other one (e.g tui in my case), the error might need to be converted which is to my understanding impossible in the current state since those types are private. Thus this change makes the error module public.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -342,5 +342,6 @@ pub use self::style::{
     TerminalColor,
 };
 pub use self::terminal::{terminal, Terminal};
+pub use common::error;
 pub use common::screen::{AlternateScreen, Screen};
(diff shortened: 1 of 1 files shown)


TimonPost:
Thanks, yes this might be helpful indeed and this should be public. I'll be working on it!

TimonPost:
Let's merge this! Thanks for the change.

TimonPost:
Crossterm 5.5 fixes this. In case you where using GitHub link as dependencies move it over to crossterm 5.5, the coming days I'll have API breaking changes for 6.0

PR #89, opened 2019-02-12
Raw (blocking) read_char for unix

scauligi:
The current implementation of read_char seems only suited for non-raw mode.
This PR adds a read_char_raw() function so that raw terminals can call read_char without backed-up characters printing, and allowing the calling code to directly handle ctrl-c input.

### DIFF
--- a/crossterm_input/src/input/unix_input.rs
+++ b/crossterm_input/src/input/unix_input.rs
@@ -1,7 +1,7 @@
 //! This is a UNIX specific implementation for input related action.
 
 use super::*;
-use crate::sys::unix::{get_tty, read_char};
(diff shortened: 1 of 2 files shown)


TimonPost:
Thanks for doing the PR! It looks good. One question though... So this code does not enable raw mode itself since we can expect that the terminal is in a raw mode right?

scauligi:
Yes, it should only be called from UnixInput::read_char, which (with this PR) check whether the screen is in raw mode or not before calling the appropriate function.

TimonPost:
Alright, I'll give it a test soon, whereafter it could be merged.

TimonPost:
Thanks, could be merged!

PR #129, opened 2019-04-28
Backtab support

scauligi:
Adds support for "backwards tab" (shift+tab) which has a special ansi keycode


scauligi:
Sorry; took me a bit to get my windows environment set up.
I made some silly mistakes with the windows code, but they should be fixed with the latest commit.

TimonPost:
Awesome and thanks for the effort!

PR #131, opened 2019-04-28
Fixes white/grey and adds dark grey

scauligi:
The ANSI codes for white and grey seem to have been switched.
I also added the ANSI code for dark grey for completeness.
I saw you have an open PR against fdehau/tui-rs; if this change makes it in then the [corresponding line there](https://github.com/fdehau/tui-rs/blob/5f8dd3813509cd409dc51202e41d7d5a6f52246b/src/backend/crossterm.rs#L172) should be updated as well to correctly translate dark grey for the CrosstermBackend.


scauligi:
Tested on windows and linux. I added dark grey to the style.rs example as well.

TimonPost:
Awesome, thanks!

PR #144, opened 2019-05-18
Prefetching buffer size for Windows console reads.

svartalf:
This should reduce memory usage when there is only small amount of events to read is available.

I'm not sure yet if it is working properly, since I do not have Windows machines available except one VM.
Would be nice to test it properly before merging.


TimonPost:
Awesome, this works on Windows 7. I will have an antoher test on Windows 10. After that, I will merge and release it.

PR #170, opened 2019-06-26
Fix Unix terminal size

### LINKED ISSUE #162: Change terminal size results to be 1-based
# Task Description
`Terminal::new().terminal_size()` reports a height and width smaller by one compared to the number of characters you can count (and the one reported by ncurses or termion).

The reason for this is that Windows is 0-based, and thus starts to count from 0 when it comes to terminal size, cursor position. Until now, cross-term maintained 0-based for everything. But it makes more sense to have it 1-based.

# Task Todo
- [ ] [Here ](https://github.com/TimonPost/crossterm/blob/master/crossterm_terminal/src/sys/unix.rs#L22)remove the '-1' from both `ws_col` and `ws_row`.
- [ ] [Here](https://github.com/TimonPost/crossterm/blob/master/crossterm_terminal/src/sys/winapi.rs#L12) add one to width and height, because windows starts counting at 0, and we need it to be 1.

twe4ked:
In #162 you mentioned needing to add 1 to the windows code. The struct we're dealing with from Windows, [`CONSOLE_SCREEN_BUFFER_INFO`](https://docs.microsoft.com/en-us/windows/console/console-screen-buffer-info-str), doesn't mention anything about their `dwSize` returning anything special. I would assume it's going to be counting “from 0” meaning that a terminal with a width of 10 should report 10. I don't have access to Windows easily to test. The [original report](https://github.com/TimonPost/crossterm/issues/155) only mentioned Unix systems.

Fixes #162.

### DIFF
--- a/crossterm_terminal/src/sys/unix.rs
+++ b/crossterm_terminal/src/sys/unix.rs
@@ -16,10 +16,7 @@ pub fn get_terminal_size() -> (u16, u16) {
     let r = unsafe { ioctl(STDOUT_FILENO, TIOCGWINSZ.into(), &mut size) };
 
     if r == 0 {
-        // because crossterm works starts counting at 0
-        // and unix terminal starts at cell 1
-        // you have subtract one to get 0-based results.
-        (size.ws_col - 1, size.ws_row - 1)
+        (size.ws_co, size.ws_row)
     } else {
         (0, 0)
     }


TimonPost:
Thanks for the fix, you are right, as mentioned in issue #162 windows is 0-based, should be increased as well. Are you able to change that as well, otherwise I merge this one and check the one checkbox in #162.

twe4ked:
Sorry, I won't be able to as I can't test it on Windows. I also don't see any documentation saying that the values for width/height are 0-based. Being 0-based is just normal counting when you're talking about values like width/height.

TimonPost:
This can be merged, windows can be done in an other PR.

PR #173, opened 2019-07-01
Improve cursor (fix panics)

sigmaSd:
reproducible case: Holding arrow up key on linux triggers the panic

This is due to parsing logic this pr should improve it, I tested it for a fair time with no panics


sigmaSd:
needs more testing

Issue #171, opened 2019-06-26
Performance Enhancement


TimonPost:
# Task Description
Crossterm flushes the buffer each time action with an ANSI-code is performed. This is not realistic in the case of moving the cursor a lot and meanwhile print characters. When calling those types of functions a lot, cross-term might behave a bit slow. 

- Flushing is done [here](https://github.com/TimonPost/crossterm/blob/master/crossterm_utils/src/macros.rs#L9)
- Locking on stdout is done as well on each ANSI write to the terminal, however, flushing is way more performance bottleneck then the lock.

# Task TODO
- [ ] Research if it is possible to let the user control the flushing so that the user can control when to flush the ANSI-commands to the terminal. Issue #167 can be related to this issue. More explanation about this in the comments.
- [ ] If we found out a way to make ANSI-codes more performant, we are left with one problem. WINAPI. The nice think about ANSI codes is that you can queue them up somewhere and push them into the stdout later on. With WinApi the user interacts directly with the terminal, and you can't queue them in the same way of ANSI-codes. 

Example of what I mean:

_winapi scenario of moving the cursor_
```
Move cursor (10.10)
println!("abc");
Move cursor (10,10)
```

_ansi scenario to move the cursor_
```
println!("\x1B[{10};{10}H abc \x1B[{5};{5}H");
```

As you can see, in the case of ANSI codes we can store the codes and text of the user inside on string, however, with WinApi we don't store actions inside a string. But those are calls that need to be made to the console individually. 

Somehow, we need to be able to store those WinApi commands in order ass well. So that if the user does a manual flush to the terminal, in the case of windows, we walk through all the WinApi commands queued and execute those, as well print the user input, done between those two cursor move events. 

# Notice
I do notice, Linux is way faster with console actions than windows is. I am not sure whether this is the problem of rust using winapi to write to the console window, [winapi ANSI interpretation layer](https://docs.microsoft.com/en-us/previous-versions/tn-archive/cc722862(v=technet.10)), crossterm performance issue. Windows is often forgotten in the Rust library, and would not be amazed if there is some performance issue there somewhere. But this needs to be proven of course.

TimonPost:
So related to #167. 

[...]

PR #175, opened 2019-07-07
Command API experiment

TimonPost:
This pull request is an experiment with issue: #171, it introduces a macro which writes commands onto the given `Write` type. The use can manually call `flush` to clear the buffer. By using this API, we can improve the performance bottleneck we are currently experimenting because there is a lot of flushing going on.

```rust
fn main() {
    let mut stdout = ::std::io::stdout();

    for x in 0..30 {
        for y in 0..30 {
            schedule!(stdout,Goto(x, y), Output(String::from("#")));
        }
    }

    thread::sleep(Duration::from_millis(3000));
    stdout.flush();
}
```

- [ ] Write documentation in book
- [ ] Write example: /examples/command.rs 
- [ ] Implement the command below
```
✔ Goto
✔UP
✔Down
✔Left
✔Right
✔SavePos
✔ResetPos
✔Hide
✔Show
✔Blink

✔ SetFg
✔ SetBg
✔ SetAttr
✔ Print Styled Text

❌ Reset
✔ Clear
✔ Scroll Up
✔ Scroll Down
❌ SetSize

- [ ] Make sure all commands are documented

Please checkout `crossterm_cursor/example/cursor.rs` for the actual implementation

### CODE REVIEW

**Comment by zesterer** [comment]
Hi. As someone looking to use this library, I'd be quite saddened if macros were introduced into the public API for this. In my humble opinion, macros should be used specifically to extend Rust's syntax, not for something as trivial as variadic functions. May I suggest using a builder method instead?

```rs
write_chain(stdout)
    .and(Goto(x, y))
    .and(Output(String::from("#")))
    .finish();
```

In my opinion, this is far more intuitive and doesn't undermine the rich semantics and visual guarantees that make Rust such a pleasant language to work with.

TimonPost:
Thanks for your input. Indeed macro's should be used for cases where they are meant for. Although macros are also super powerful in rust, and it is a feature a lot of people use. Also in public API. 

So we either do:
```
write_chain(stdout)
    .and(Goto(x, y))
    .and(Output(String::from("#")))
    .finish();
```

Or
```
schedule!(stdout, Goto(x, y), Hide, Output("#"));
```

I have several problems with the builder idea

1. If we do the builder approach we have to queue commands into some vector, store them as traits, which will result in the dynamic dispatch.
2. We can't less easily move the chain builder around. It needs to be a mutable type, because of the add method. When working in a multithreaded application we are probably going to get mutexes and a lot of arcs around this type. 
3. Boilerplate code
We can choose to create a new chain every time we want to queue another command. As a workaround of point 2, this would look like:

```rust
write_chain(stdout)
    .and(Goto(x, y));
// some code
write_chain(stdout)
    .and(Goto(x, y));
// some code
write_chain(stdout)
    .and(Goto(x, y))
   finishe();
```

This would work way better with the macro where we do:

```rust
 schedule!(stdout, Goto(x, y);
// some code
 schedule!(stdout, Goto(x, y);
// some code
 schedule!(stdout, Goto(x, y);
stdout.flush()
```

4. In most cases, the user only wants to use the default stdout.
With the builder solution, we would have to create an other function that creates the builder which would have a default argument with `std::io::stdout()`.
```rust
default_write_chain(stdout)
    .and(Goto(x, y));
```
And with macro's we can just leave a way that argument:

```rust
 schedule!(Goto(x, y);
```


So all with all, I still prefer the macro functions, 
1) they can have different arguments and are more flexible, 
2) are stateless, and thus no mutexes or atomic types are needed, 
3) less boilerplate code

zesterer:
To address your concerns:

1) There is absolutely no need to do dynamic dispatch, in the same way that Rust's iterator API doesn't do dynamic dispatch. Encoding this stuff at the type level with a builder pattern is quite trivial.

2) Why not? Iterators can be moved around, and this would have a similar internal representation

3) Given the circumstance you mention, I suggest a better API:

A trait implemented for all `Write` types, that looks permits this:

```rs
stdout
    .queue(Goto(x, y))
    .queue('!')
    .queue(Goto(a, b))
    .flush();
```

In the single case, it's just a matter of doing `stdout.queue(Goto(x, y))` which keeps the API very simple.

4) You seem to be forgetting that macros come with a lot of disadvantages. I'm sure as the author you do not see these disadvantages, but they tend to be an indicator of 'magic' code. When working with macros, it's hard to predict their behaviour and Rust developers have been conditioned to abandon existing rules about how code runs when seeing a macro. For this reason, I'd really advise avoiding them.

TimonPost:
I'll take your concern into consideration, I will make a test that works with the type system as well.

TimonPost:
I currently implemented via the function as well:

We have two traits:

```rust
pub trait QueueableCommand<T: Display> {
    fn queue(mut self, command: impl Command<AnsiType = T>) -> Self;
}

pub trait ExecutableCommand<T: Display> {
    fn execute(mut self, command: impl Command<AnsiType = T>) -> Self;
}
```

I implement those traits for all types implementing `Write`
```rust
impl<T, A> QueueableCommand<A> for T
where
    A: Display,
    T: Write,
{
    fn queue(mut self, command: impl Command<AnsiType = A>) -> Self {
        schedule!(self, command);
        self
    }
}

impl<T, A> ExecutableCommand<A> for T
where
    A: Display,
    T: Write,
{
    fn execute(mut self, command: impl Command<AnsiType = A>) -> Self {
        execute!(self, command);
        self
    }
}
```

Now I am able to call the queue and executing functionality on Stdout:

```rust
```
 let mut stdout = ::std::io::stdout();

    stdout
        .queue(Goto(5, 5))
        .queue(Output("#".to_string()))
        .flush();
```
```

TimonPost:
We have two options:
1) Keep the macros for users that want to have less boilerplate code and don't bother working with macros.
2) Remove the macro's and only have support for the builder functions.

_macros_
```rust
schedule!(Goto(x, y), Output("#"));
::std::io::flush();
```rust

Builder:
```rust
 ::std::io::stdout()
     .queue(Goto(5, 5))
     .queue(Output("#".to_string()))
     .flush();
```

zesterer:
Nice job! I much prefer this design, thanks for working on it.

TimonPost:
Will be merged now, documentation, examples, and comments are all finished.  I implemented this branch for both TUI and cursive, to test the functionality. Will be released soon under version 10

PR #176, opened 2019-07-08
Fix Windows terminal size (#170)

natbraun:
This change is [Reviewable](https://reviewable.io/reviews/timonpost/crossterm/176)


TimonPost:
Thanks for your work! Looks good to me, I forgot about that other place that was getting the screen size, good that you noticed that place as well. Actually `WinApiTerminal::terminal_size()` should use `get_terminal_size` function as well, instead of having duplicated code. Would be nice, if you could fix that in this PR, otherwise we can just merge it, and I make an issue for it.

natbraun:
Just pushed a new revision

natbraun:
By the way, in order to be complete, we also have to make **setting** the size consistent across platforms and 1-based.

TimonPost:
Right, let me create a new issue for that

PR #177, opened 2019-07-14
fix typos and tests

andy128k:


crossterm_terminal/src/terminal/terminal.rs
```diff
@@ -52,18 +52,19 @@ impl Terminal {
     ///
     /// # Example
     /// ```rust
+    /// # use crossterm_terminal::terminal;
     /// let mut term = terminal();
     ///
     /// // clear all cells in terminal.
-    /// term.clear(terminal::ClearType::All);
```
TimonPost: I do understeand this change, but I am in a dilema about this. Because if the user used `crossterm` crate, he will see those docs but is not able to use them as `crossterm_terminal` because all types in this crate are re exported in crossterm and therefore they can be: `crossterm`

TimonPost:
I do understand this change, but I am in a dilemma about this. Because if the user used `crossterm` crate, he will see those with `crossterm_terminal` but is not able to use them as `crossterm_terminal` because all types in this crate are re exported in crossterm and therefore they are: `crossterm::_some_type`

andy128k:
@TimonPost  done

TimonPost:
Thanks for the improvements!

Issue #179, opened 2019-07-16
Add serde support


zesterer:
It would be very useful to be able to serialise things like the colour and attribute types in `crossterm`.

TimonPost:
Good idea, however, it needs to be behind a feature flag.

arnej:
This would be very useful for me. I need KeyEvent to implement Deserialize and Serialize.
Can you tell us why this needs to be behind a feature flag?

I would like to help with this if I can.

TimonPost:
I have two reasons for that:
1) Because I want to have as less external deps as possible. Currently, I only have WinApi and Libc. 
2) Not all users are going to use serde serialize and deserialize and the dependencie to serde is going to be unnecessarily in most cases.

If you have questions on how to implement this please ask away here :)

arnej:
That sounds reasonable :)

I will have a look at this soon and will come back to you when I can't find out how to implement this as a feature.

arnej:
I finally had the time to work on this. See my pull request here: [#190](https://github.com/TimonPost/crossterm/pull/190)

TimonPost:
Thanks!

PR #190, opened 2019-08-06
Add optional derives for Serialize and Deserialize for public enums

arnej:
This adds optional derives for serde Serialize and Deserialize to all public enums except ErrorKind. Please let me know if you have any objections against this approach.
Reference issue: [#179](https://github.com/TimonPost/crossterm/issues/179)


TimonPost:
Thanks for the PR, looks good!

PR #207, opened 2019-09-13
crossterm_input test/docstrings fixes

zrzka:
Fixes `crossterm_input` test & doc string examples.

`cd crossterm_input && cargo test` is passing.

### DIFF
--- a/crossterm_input/src/input/input.rs
+++ b/crossterm_input/src/input/input.rs
@@ -44,11 +44,11 @@ impl TerminalInput {
     ///
     /// # Example
     /// ```rust
-    /// let input = input();
-    ///  match input.read_line() {
+    /// let input = crossterm_input::input();
+    /// match input.read_line() {
     ///     Ok(s) => println!("string typed: {}", s),
     ///     Err(e) => println!("error: {}", e),
-    ///  }
+    /// }
     /// ```
     pub fn read_line(&self) -> io::Result<String> {
         let mut rv = String::new();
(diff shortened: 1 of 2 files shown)


crossterm_input/src/input/input.rs
```diff
@@ -44,11 +44,11 @@ impl TerminalInput {
     ///
     /// # Example
     /// ```rust
-    /// let input = input();
-    ///  match input.read_line() {
+    /// let input = crossterm_input::input();
```
TimonPost: So this was another concern of me, most of the time people will be using `crossterm_input` through crossterm. All types and functions are re-exported. In that case, this namespace is not right and people might get confused because they don't have access to that namespace.

zrzka: These reexports ... What do you propose? Remove these two examples? Move them to the book / root crate?

zrzka: What do you think [about this](https://github.com/TimonPost/crossterm/pull/207/commits/945e1ba7d13c0cbf7202bf0531f0feee7183bd90)? Tests are passing, these two examples marked as ignored (even in the tests output), ...

TimonPost: Ignoring examples might cause old invalid code in the feature. But I much more like it to ignore the compilation then having wrong imports. So that looks good to me

zrzka: At least it's visible in the test output and can be solved in a better way later. Marking as resolved.

Issue #205, opened 2019-09-12
Travis - Not all tests are executed


zrzka:
`.travis.yml` does contain following lines:

```bash
cargo fmt -- --check
cargo test -- --nocapture --test-threads 1
```

But these commands check formatting / test the `crossterm` crate only. An example - modify `crossterm_terminal/examples/terminal.rs` (make it not compilable) and then run `cargo test -- --nocapture --test-threads 1`. It passes. Same applies to formatting, etc.

zrzka:
```bash
warning: output filename collision.
The example target `terminal` in package `crossterm_terminal v0.2.5 (/Users/robertvojta/Projects/personal/crossterm/crossterm_terminal)` has the same output filename as the example target `terminal` in package `crossterm v0.10.1 (/Users/robertvojta/Projects/personal/crossterm)`.
Colliding filename is: /Users/robertvojta/Projects/personal/crossterm/target/debug/examples/terminal.dSYM
The targets should have unique names.
Consider changing their names to be unique or compiling them separately.
This may become a hard error in the future; see <https://github.com/rust-lang/cargo/issues/6313>.
```

zrzka:
I saw that you did add GH Actions support, but here's my Travis CI change proposal anyway:

* Replace `rustup component add rustfmt-preview` with `rustup component add rustfmt`
* If `$TRAVIS_OS_NAME == "windows"` (just add `--all` to both)
  * `cargo fmt --all -- --check`
  * `cargo test --all -- --nocapture --test-threads 1`
* If `$TRAVIS_OS_NAME != "windows"` (add `--all` & `--exclude ...`)
  * `cargo fmt --all --exclude crossterm_winapi -- --check`
  * `cargo test --all --exclude crossterm_winapi -- --nocapture --test-threads 1`
* plus examples collisions (see above)
* plus remove `.travis.yml` files from nested folders

TimonPost:
I will take that and move it into GH actions. Travis is probably going to be disabled.

TimonPost:
https://github.com/TimonPost/crossterm/pull/215

PR #215, opened 2019-09-13
Update crossterm_test.yml

TimonPost:
#205

### DIFF
--- a/.github/workflows/crossterm_test.yml
+++ b/.github/workflows/crossterm_test.yml
@@ -1,49 +1,47 @@
 name: Crossterm Test
-on:
-  push:
-    branches:
-    - master
-  pull_request:
-    branches:
-    - master
+on: [push, pull_request]
 jobs:
   crossterm_rust:
     name: Running on ${{ matrix.os }}
+
     runs-on: ${{ matrix.os }}
+
     strategy:
       matrix:
         os: [ubuntu-latest, windows-2019, macOS-10.14]
         rust: [stable, nightly]
         include:
-          - os: ubuntu-latest, windows-2019, macOS-10.14
-            rust: 'stable'
-            components: 'rustfmt'
+          - rust: stable
+            can-fail: false
+          - rust: nightly
+            can-fail: true
     steps:
-    - uses: hecrj/setup-rust-action@master
-      with:
-        rust-version: ${{ matrix.rust }}
-        components: ${{ matrix.components || '' }}
-    - uses: actions/checkout@master
-    
-    - name: Debug Information
-      run: | 
-        cargo +stable fmt --version
-        rustup --version
-        rustc --version
-      
-    - name: Run Build
-      run: cargo build
-      
-    - name: Run fmt, tests windows
-      if: matrix.os == 'windows-2019'
-      run: | 
-        cargo test --all -- --nocapture --test-threads 1
-        cargo +stable fmt --all -- --check
-      
-    - name: Run fmt, tests if not Windows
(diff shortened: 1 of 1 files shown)


.github/workflows/crossterm_test.yml
```diff
@@ -17,9 +20,20 @@ jobs:
       with:
         rust-version: ${{ matrix.rust }}
     - uses: actions/checkout@master
+    - name: Add Component
```
zrzka: See [this](https://github.com/hecrj/setup-rust-action/blob/master/action.yml#L8-L10). You can use `with.components: 'rustfmt'`.

.github/workflows/crossterm_test.yml
```diff
@@ -1,25 +1,52 @@
...
+    - name: Debug Information
+      run: | 
+        cargo fmt --version
+        rustup --version
+        rustc --version
+      
     - name: Run Build
       run: cargo build
+      
     - name: Check Examples
```
zrzka: This step is redundant, `cargo test --all` covers this.

zrzka:
What are you trying to achieve? Check [this workflow](https://github.com/zrzka/crossterm/blob/zrzka/gh-actions/.github/workflows/crossterm_test.yml). It:

* uses stable & nightly
* uses rustfmt on both
* checks formatting on both
* run tests on both

And what's more important, it requires stable to pass and nightly is allowed to fail.

TimonPost:
Thanks, I will try that.

zrzka:
Seems that GH actions are kind of broken. I've got lot of canceled builds without a good reason. No idea what's going on, still beta, ...

TimonPost:
It broken on macos fromat check

zrzka:
Hmm ...

```bash
(zrzka/fix-cargo-fmt ✓) crossterm /bin/bash -e cargo fmt --all -- --check
/Users/robertvojta/.cargo/bin/cargo: /Users/robertvojta/.cargo/bin/cargo: cannot execute binary file
(zrzka/fix-cargo-fmt ✓) crossterm cargo fmt --all -- --check
(zrzka/fix-cargo-fmt ✓) crossterm echo $?
0
(zrzka/fix-cargo-fmt ✓) crossterm 
```

zrzka:
It's weird, because when you check _Check code formatting_ [here](https://github.com/zrzka/crossterm/runs/222137493), it was started, but then canceled.

TimonPost:
From what I see here, is that it stops after a few errors when checking for fmt.
 

https://github.com/TimonPost/crossterm/pull/215/checks?check_run_id=222174585#step:5:4

zrzka:
Yeah, if I understand it correctly, the build was canceled by GH Actions, not that rustfmt fails or anything like this.

zrzka:
We have to sign up again for GH actions to continue with testing. Because we moved the repo to the organization, the whole `crossterm-rs` org should ask for it.

TimonPost:
Done, I think that once everything is merged we should try this PR again.

zrzka:
Closing this PR. Changes:

* reenabled GitHub actions
* made the job name shorter to fit the column width
* added packaging test (stable only)
* synced steps with `.travis.yml`

Seems that it works properly. Check [this test suite](https://github.com/crossterm-rs/crossterm/commit/5904a2a10677a36084cbbc08e9589ea6b06b47cf/checks?check_suite_id=271677333).

GitHub Actions are way faster than Travis CI, but still beta. So, keeping the `.travis.yml` there, which should be still treated as a mandatory test for a PR.

PR #222, opened 2019-09-16
2018 edition

### LINKED ISSUE #213: 2018 edition
Some crates are 2015, some are 2018. Are you okay with converting all of them to 2018?

zrzka:
### 2018 edition changes

* All crates (`Cargo.toml`) do contain `edition = "2018"`
* All occurrences of `extern crate ...` removed
* Macros imported like any other types

### Warnings

* Replaced `ONCE_INIT` with `Once::new`
  * Will [be deprecated](https://travis-ci.org/zrzka/crossterm/jobs/585436132#L79-L92) in the new Rust version
* Prefixed all subcrate examples with subcrate abbreviation
  * Example `crossterm_cursor/examples/cursor.rs` -> `crossterm_cursor/examples/cc_cursor.rs`
  * I agree that it's ugly and that [it has to be solved somehow](https://github.com/TimonPost/crossterm/issues/221), but it's polluting build logs and will be a hard error in the future, take it as a temporary _fix_

### Imports

All imports are semantically ordered. The order is:

* Standard library
* External crates
* `crate`
* `super`
* `self`
* `mod`

I do not expect that everyone is going to maintain this order, but I consider it as a good practice to make the code more readable.

Fixes #213


TimonPost:
Awesome! Yea I had an other order I maintained import order for a while. Thanks for putting in the work for the arranging them to an other order. I will try to keep that as default, I might want to create some code_style.md.

I'm really doubting that we might want to remove the examples from the sub crates and instead use those in /examples, because all those examples are mostly duplicated with the only difference being the namespaces

zrzka:
> I'm really doubting that we might want to remove the examples from the sub crates and instead use those in /examples, because all those examples are mostly duplicated with the only difference being the namespaces.

Yeah, that's the reason why I just prefixed them with an abbreviation (to _fix_ warnings) and made a [separate issue](https://github.com/TimonPost/crossterm/issues/221) for this. I'm not aiming to solve the examples in this PR.

zrzka:
Don't merge yet, I've just found that I missed the `examples/program_examples`. Going to fix them right now.

zrzka:
Okay, I think it's ready for review. Couple of new things:

* Missed some `extern crate ...`
* Converted `examples/` as well
* Removed `extern crate` from docs
* Added build & formatting check (`stable` only) for `examples/program_examples`

crossterm_style/src/color.rs
```diff
@@ -4,13 +4,14 @@
...
-use super::*;
-use crate::{Color, ITerminalColor};
-
 #[cfg(windows)]
 use crossterm_utils::supports_ansi;
 use crossterm_utils::{impl_display, Command, Result};
 
+use crate::{Color, ITerminalColor};
+
+use super::*;
```
TimonPost: Eventually, I like to get rid of those as well, doesn't need to be done in this PR tough

crossterm_style/src/lib.rs
```diff
@@ -1,6 +1,19 @@
...
+pub use crossterm_utils::{execute, queue, Command, ExecutableCommand, QueueableCommand, Result};
+
+use self::ansi_color::AnsiColor;
+pub use self::color::{color, PrintStyledFont, SetAttr, SetBg, SetFg, TerminalColor};
+pub use self::enums::{Attribute, Color, Colored};
+pub use self::objectstyle::ObjectStyle;
+pub use self::styledobject::StyledObject;
+pub use self::traits::{Colorize, Styler};
+#[cfg(windows)]
+use self::winapi_color::WinApiColor;
```
TimonPost: This probably isn't supposed to be here. I see that it already was there, but wondering why....

zrzka:
```rust
pub use self::styledobject::StyledObject;
```

What about `pub use self::objectstyle::ObjectStyle`? Same category, no? You'd like to remove these two reexports, because they're accessible via public `objectstyle` & `styledobject` modules? If yes, examples needs to be updated as well, because they do use crate root reexport.

```rust
pub use self::traits::{Colorize, Styler};
```

These traits are used in examples and this reexports them, because the `traits` module isn't public (`mod traits`).

```rust
#[cfg(windows)]
use self::winapi_color::WinApiColor;
```

Seems useless, no usage in the `lib.rs`. Wondering why Travis CI log doesn't say unused import. Will remove it.

crossterm_style/src/color.rs
zrzka: Can you elaborate more? Why to remove `crate::...`? Replace `super::*` with specific imports to avoid `*`?

TimonPost: Yea, * is unclear to which types your importing and can create confusion

crossterm_style/src/lib.rs
TimonPost: It think that `WinApiColor` should be imported directly. Now it is re-exported at the top level of the crate so that all files have it when doing `use super::*`

zrzka: `WinApiColor` import already moved the `src/color.rs` where it's used.

crossterm_style/src/color.rs
zrzka: Okay, I'll do this. But if you don't mind, in a separate PR, cause this one is mainly about 2018 edition.

TimonPost: Yea totally fine

TimonPost:
Just give the sign and I merge it

zrzka:
There's no more things to do in this PR. I mean, 2018 edition related. I think it can be merged and comments can be addressed in another PRs.

PR #224, opened 2019-09-16
Wildcard import cleanup

zrzka:
* No `::*` imports
* Some of the imports reorganized a bit
* Reorganized tests
  * Moved tests to proper modules & removed `tests.rs` modules
* `#![deny(unused_imports)]` (hard error) added to all `lib.rs` to force contributors to keep them clean


zrzka:
Sorry for a lot of commits, but as I don't have Windows machine around I have to check Travis CI output :)

I'm not planning to do more in this PR, if you find it okay, feel free to merge.

crossterm_cursor/src/cursor/ansi_cursor.rs
```diff
@@ -104,3 +104,62 @@ impl ITerminalCursor for AnsiCursor {
         Ok(())
     }
 }
+
```
TimonPost: I like the tests being here

crossterm_cursor/src/lib.rs
```diff
@@ -1,3 +1,5 @@
+#![deny(unused_imports)]
```
TimonPost: Strange wy it complains about unused import like it is quite obvious that they are re-exported for external use

zrzka: Where it complains? Maybe you just swapped `allow` and `deny` for a moment. `allow` says that unused imports are okay. `deny` says that it's an error and Travis will fail. Did add it so now new unused imports are introduced (CI check).

PR #226, opened 2019-09-17
Command API improvements: don’t ignore errors, don’t require reassigning stdout back to itself

andersk:
* `QueueableCommand`, `ExecutableCommand`: don’t ignore errors. Return `io::Result<Self>` from `QueueableCommand::queue` and `ExcutableCommand::execute`. This lets us get rid of the `let _ =` pattern that ignores errors.
* `QueueableCommand`, `ExecutableCommand`: Use references to the writer. This lets us dispense with the noise of reassigning `stdout` back to itself.

### DIFF
--- a/examples/command.rs
+++ b/examples/command.rs
@@ -39,10 +39,10 @@ fn later_execution_command_using_functions() -> Result<()> {
     let mut sdout = stdout();
 
     // single command
-    sdout = sdout.queue(Output("Text1 ".to_string()))?;
+    sdout.queue(Output("Text1 ".to_string()))?;
 
     // multiple commands
-    sdout = sdout
+    sdout
         .queue(Clear(ClearType::All))?
         .queue(Goto(5, 5))?
         .queue(Output(
--- a/examples/command.rs
+++ b/examples/command.rs
@@ -3,73 +3,81 @@
 use std::io::{stdout, Write};
 
 use crossterm::{
-    execute, queue, Clear, ClearType, ExecutableCommand, Goto, Output, QueueableCommand,
+    execute, queue, Clear, ClearType, ExecutableCommand, Goto, Output, QueueableCommand, Result,
 };
 
 /// execute commands by using normal functions
-fn execute_command_directly_using_functions() {
+fn execute_command_directly_using_functions() -> Result<()> {
     // single command
-    let _ = stdout().execute(Output("Text1 ".to_string()));
+    stdout().execute(Output("Text1 ".to_string()))?;
 
     // multiple commands
-    let _ = stdout()
-        .execute(Output("Text2 ".to_string()))
-        .execute(Output("Text3 ".to_string()));
+    stdout()
+        .execute(Output("Text2 ".to_string()))?
+        .execute(Output("Text3 ".to_string()))?;
+
+    Ok(())
 }
 
 /// execute commands by using macro's
-fn execute_command_directly_using_macros() {
+fn execute_command_directly_using_macros() -> Result<()> {
     // single command
-    let _ = execute!(stdout(), Output("Text1 ".to_string()));
+    execute!(stdout(), Output("Text1 ".to_string()))?;
 
     // multiple commands
-    let _ = execute!(
+    execute!(
         stdout(),
         Output("Text2 ".to_string()),
         Output("Text 3".to_string())
-    );
+    )?;
+
+    Ok(())
 }
 
 /// queue commands without executing them directly by using normal functions
-fn later_execution_command_using_functions() {
+fn later_execution_command_using_functions() -> Result<()> {
     let mut sdout = stdout();
 
     // single command
-    sdout = sdout.queue(Output("Text1 ".to_string()));
+    sdout = sdout.queue(Output("Text1 ".to_string()))?;
 
     // multiple commands
     sdout = sdout
-        .queue(Clear(ClearType::All))
-        .queue(Goto(5, 5))
+        .queue(Clear(ClearType::All))?
+        .queue(Goto(5, 5))?
         .queue(Output(
             "console cleared, and moved to coord X: 5 Y: 5 ".to_string(),
-        ));
+        ))?;
 
     ::std::thread::sleep(std::time::Duration::from_millis(2000));
 
     // when you call this all commands will be executed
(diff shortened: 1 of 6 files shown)


examples/terminal.rs
```diff
@@ -134,6 +134,8 @@ pub fn exit() {
     terminal.exit();
 }
 
-fn main() {
-    let _ = scroll_down();
+fn main() -> io::Result<()> {
+    scroll_down()?;
```
zrzka: You can remove `?;` and `Ok(())`. The only line here will be `scroll_down()`.

examples/command.rs
```diff
@@ -1,75 +1,83 @@
...
         Goto(5, 5),
         Output("console cleared, and moved to coord X: 5 Y: 5 ".to_string())
-    );
+    )?;
 
     ::std::thread::sleep(std::time::Duration::from_millis(2000));
 
     // when you call this all commands will be executed
-    let _ = stdout.flush();
+    stdout.flush()?;
```
zrzka: You can remove `?;` and `Ok(())`.

examples/command.rs
```diff
@@ -1,75 +1,83 @@
...
-    let _ = stdout().execute(Output("Text1 ".to_string()));
+    stdout().execute(Output("Text1 ".to_string()))?;
 
     // multiple commands
-    let _ = stdout()
-        .execute(Output("Text2 ".to_string()))
-        .execute(Output("Text3 ".to_string()));
+    stdout()
+        .execute(Output("Text2 ".to_string()))?
+        .execute(Output("Text3 ".to_string()))?;
```
zrzka: You can remove `?;` and `Ok(())`.

examples/command.rs
```diff
@@ -1,75 +1,83 @@
...
+    execute!(stdout(), Output("Text1 ".to_string()))?;
 
     // multiple commands
-    let _ = execute!(
+    execute!(
         stdout(),
         Output("Text2 ".to_string()),
         Output("Text 3".to_string())
-    );
+    )?;
```
zrzka: You can remove `?;` and `Ok(())`.

examples/command.rs
```diff
@@ -1,75 +1,83 @@
...
         .queue(Output(
             "console cleared, and moved to coord X: 5 Y: 5 ".to_string(),
-        ));
+        ))?;
 
     ::std::thread::sleep(std::time::Duration::from_millis(2000));
 
     // when you call this all commands will be executed
-    let _ = sdout.flush();
+    sdout.flush()?;
```
zrzka: You can remove `?;` and `Ok(())`.

examples/terminal.rs
andersk: Of course I can, but should I? I think eliding the `?; Ok(())` is less clear, especially in functions that do multiple things where the last thing is mysteriously missing a `?` just because it’s last. It also prevents error conversion, e.g. from `io::Result` to `crossterm::Result`.

If you still disagree, I’ll change it (or you can).

TimonPost: I am neutral in this :)

TimonPost:
Sorry, but I broke your branch :)

examples/terminal.rs
zrzka: My thoughts ...

* What is the reason to return `io::Result` instead of `crossterm::Result`?

... nitpicks ...

* Why to add another import `use std::io`?
* There's nothing mysterious about how Rust works in terms of the last expression and a function result
* The example is 20 lines shorter

zrzka: P.S. I do not want to argue much about this, if Timon is okay with this, let be it. I'm just trying to make it shorter, especially in such a simple case where there's a function like `set_terminal_size` which calls just `set_size`. I'm also aware that you didn't introduce `io::Result`, ... in this PR.

andersk: > * What is the reason to return `io::Result` instead of `crossterm::Result`?

`io::Result` is more specific than `crossterm::Result`; if that’s the only kind of error that can happen, we shouldn’t force callers to handle the other unused cases.

zrzka: > if that’s the only kind of error that can happen

As user of the `crossterm` crate, you don't know this, do you?

Look at this example (`examples/terminal.rs`).

```rust
pub fn set_terminal_size() -> io::Result<()> {
    let terminal = terminal();
    terminal.set_size(10, 10)
}
```

* As a user, I see that the `terminal.set_size` returns `Result<(), ErrorKind>`
* What if the `set_size` will return `ErrorKind::ResizingTerminalFailure`?
* How can I, as a user, know that it wont return `ResizingTerminalFailure`?
* And if it will, you'll end up with the _can not convert error to IO error_ error, which is a bit unfriendly

```rust
impl From<ErrorKind> for io::Error {
    fn from(e: ErrorKind) -> io::Error {
        match e {
            ErrorKind::IoError(io) => return io,
            _ => io::Error::new(io::ErrorKind::Other, "can not convert error to IO error"),
        }
    }
}
```

zrzka: Let's not pollute this PR more. It can be merged and discussed elsewhere (Discord, GH issue).

andersk: That is indeed a mistake. (Perhaps `crossterm::Error` should not be convertable to `io::Error`?)

zrzka: As this particular issue (and main reason for my comments) is out of scope of this PR (you didn't introduce this), I'm going to create an issue for this. Sounds okay?

andersk: Trying to remove that conversion `impl` reveals that the existing code is not consistent, so I’m not sure what the right thing is. I’ve updated my PR to use `crossterm::Result`, except in the one example that already uses `io::Result` everywhere.

zrzka: I made [a generic issue](https://github.com/TimonPost/crossterm/issues/229) for this problem as it's unclear to me why the conversions exists, why we're using it in examples, etc.

zrzka:
@TimonPost can you merge this one? There was an ongoing discussion, which was basically ended with the #229 issue. I'm working on the #229 now and it's conflicting (merge point of view) with this PR. Either I or Anders will be forced to rebase & solve the conflicts again.

TimonPost:
Merged!

PR #227, opened 2019-09-17
ITerminalCursor::pos() - sync platforms behaviour

### LINKED ISSUE #223: WinApiCursor::pos call unwrap internally
Compare `WinApiCursor::pos` and `AnsiCursor::pos`.

* `WinApiCursor` does call `unwrap` -> possible panic
* `AnsiCursor` returns `(0, 0)` in case of error -> no panic

Shouldn't the behaviour be synced? Return `(0, 0)` in case of error in the `WinApiCursor::pos`?

`WinApiCursor::pos`:

```rust
fn pos(&self) -> (u16, u16) {
    let cursor = Cursor::new().unwrap();
    cursor.position().unwrap().into()
}
```

`AnsiCursor::pos`:

```rust
fn pos(&self) -> (u16, u16) {
    get_cursor_position()
}

pub fn get_cursor_position() -> (u16, u16) {
    if unsafe { RAW_MODE_ENABLED } {
        if let Ok(pos) = pos_raw() {
            pos
        } else {
            (0, 0)
        }
    } else {
        if let Ok(pos) = pos() {
            pos
        } else {
            (0, 0)
        }
    }
}
```

zrzka:
Fixes #223

### DIFF
--- a/crossterm_cursor/src/cursor.rs
+++ b/crossterm_cursor/src/cursor.rs
@@ -27,7 +27,9 @@ mod winapi_cursor;
 trait ITerminalCursor: Sync + Send {
     /// Goto some location (x,y) in the context.
     fn goto(&self, x: u16, y: u16) -> Result<()>;
-    /// Get the location (x,y) of the current cursor in the context
+    /// Get the location (x,y) of the current cursor in the context.
+    ///
+    /// `(0, 0)` is returned in case of an error.
     fn pos(&self) -> (u16, u16);
     /// Move cursor n times up
     fn move_up(&self, count: u16) -> Result<()>;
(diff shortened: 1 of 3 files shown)


crossterm_cursor/src/cursor.rs
```diff
@@ -27,7 +27,9 @@ mod winapi_cursor;
 trait ITerminalCursor: Sync + Send {
     /// Goto some location (x,y) in the context.
     fn goto(&self, x: u16, y: u16) -> Result<()>;
-    /// Get the location (x,y) of the current cursor in the context
+    /// Get the location (x,y) of the current cursor in the context.
```
TimonPost: "in the context"? Sounds a bit weird, can we change that to
```suggestion
    /// Get the cursor location (x,y) in the current terminal window.
```

zrzka: Updated with couple of other changes.

PR #228, opened 2019-09-17
A bit nicer + up to date contribution guide

zrzka:


docs/Contributing.md
```diff
@@ -1,50 +1,89 @@
...
+* [style](https://github.com/TimonPost/crossterm/tree/master/crossterm_style)
+* [terminal](https://github.com/TimonPost/crossterm/tree/master/crossterm_terminal)
+* [screen](https://github.com/TimonPost/crossterm/tree/master/crossterm_screen)
+* [utils](https://github.com/TimonPost/crossterm/tree/master/crossterm_utils)
+* [winapi](https://github.com/TimonPost/crossterm/tree/master/crossterm_winapi)
+
+### Module structure
+
+If you would like to contribute, then please follow the existing structure. For
+example, a module like cursor has the following file structure:
```
TimonPost:
```suggestion
example, the cursor crate has the following file structure:
```

docs/Contributing.md
```diff
@@ -1,50 +1,89 @@
...
-
-- standard library
-- external crates
-- crate
-- super
-- self
-- mod
\ No newline at end of file
+
+Crossterm is using ANSI escape codes by default for both Unix and Windows systems. It is a bit more
```
TimonPost: Crossterm is using ANSI escape codes by default for both Unix and for Windows systems except for Windows versions lower than 10. Crossterm uses WinAPI in this case.

docs/Contributing.md
zrzka: Updated.

docs/Contributing.md
zrzka: Updated.

PR #236, opened 2019-09-19
feat(input): Add KeyEvent::Enter

### LINKED ISSUE #234: Consider adding KeyEvent::Enter
Hello !

Just began using the library and I searched for more than 5 minutes if I was having a bug with Enter not being present in the enum.

 ```rust 
// Pressing Enter in terminal apps is pretty recurrent, so I really think
// this is easier to use and read
KeyEvent::Enter
// Than this
KeyEvent::Char('\n')
```

I can send a PR if you're interested

MonsieurMan:
Add KeyEvent::Enter.

Also remove b'\r' from #[cfg(unix)] -> verify this is correct behavior.

Closes #234

### DIFF
--- a/crossterm_input/src/input.rs
+++ b/crossterm_input/src/input.rs
@@ -89,6 +89,7 @@ pub enum MouseButton {
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 pub enum KeyEvent {
     Backspace,
+    Enter,
     Left,
     Right,
     Up,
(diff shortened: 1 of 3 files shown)


MonsieurMan:
I just realised, this is **breaking change** as matching KeyEvent::Char('\n') will no longer work, it would be the same for tab.

I think I'm going to take a look at other libraries that provides enums for Keyboard input and look if there is more common key we should add. This way we could change it all in one go.

zrzka:
Breaking change is fine, minor version will be bumped. There's a plan to release new crossterm by the end of this week / beginning of the next week. So, if you'd like to see more `KeyEvent` variants, it would be nice if we can fit all these breaking changes into one release. But it's not mandatory, it's a nice to have.

TimonPost:
Jub release  0.11 will be the next one

MonsieurMan:
Nice ! I should have some time tommorow and this weekend 👍

PR #237, opened 2019-09-19
Drop implementations cleanup

zrzka:
* Never panic (`.unwrap()`) in the `drop` function
* `RawScreen`
  * `drop` field renamed to `disable_raw_mode_on_drop`
  * `disable_drop` fn renamed to `disable_raw_mode_on_drop()`
  * Both were misleading and kind of implied that the whole value (`RawScreen`) won't be dropped. Actually, the value will be dropped, but only the raw mode won't be disabled.

### DIFF
--- a/crossterm_screen/src/screen/alternate.rs
+++ b/crossterm_screen/src/screen/alternate.rs
@@ -75,6 +75,6 @@ impl AlternateScreen {
 impl Drop for AlternateScreen {
     /// This will switch back to the main screen on drop.
     fn drop(&mut self) {
-        self.to_main().unwrap();
+        let _ = self.to_main();
     }
 }
--- a/crossterm_screen/src/screen/raw.rs
+++ b/crossterm_screen/src/screen/raw.rs
@@ -24,7 +24,7 @@ use crate::sys;
 ///
 /// Please note that if this type drops, the raw screen will be undone. To prevent this behaviour call `disable_drop`.
 pub struct RawScreen {
-    drop: bool,
+    disable_raw_mode_on_drop: bool,
 }
 
 impl RawScreen {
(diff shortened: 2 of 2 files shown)


crossterm_screen/src/screen/alternate.rs
```diff
@@ -75,6 +75,6 @@ impl AlternateScreen {
 impl Drop for AlternateScreen {
     /// This will switch back to the main screen on drop.
     fn drop(&mut self) {
-        self.to_main().unwrap();
+        let _ = self.to_main();
```
TimonPost: But if we do this we swallow an error, is that correct behavior?

zrzka: The rule we're violating here is not to panic inside the `drop` fn. We have couple of options here ...

* Leave it as it is in this PR
* Add a new dependency ([log](https://crates.io/crates/log) crate) and use `error!` macro to log the error

... I'd say that we should proceed with the 2nd option. Swallowing errors is not a good practice as well.

zrzka: Imagine that you're user of this crate. You have lot of `Result`s everywhere, you're handling them and now you've got sudden panic and it's not your (user) fault. My plan is to remove all dangerous `unwrap()` calls, but not in this PR and not in the incoming release. Take it as a note what's in my queue.

TimonPost: Okay although I don't like adding extra dependenies. I did an unwrap cleanup my self as well a while ago, and moved all unwraps to the top level api. When I did that I  already had made a lot of changes to the user api and decided to just panic on error and add result types in later releases. So that's were most of the library is at now. I will merge this PR

PR #239, opened 2019-09-20
Add \r back for the KeyEvent::Enter

zrzka:
Actually, that's my mistake. When I said that the `\r` is useless on Unix, I was right & wrong. In raw mode, the Enter key sends CR and the terminal driver translates this to the new line. I've just tested this and without the `\r`, the input event is CTRL-M, which is Enter. When added back, it's correctly translated to the `KeyEvent::Enter`. Sorry for introducing this bug.

Introduce `KeyEvent::Tab` mentioned in the #236.

### DIFF
--- a/crossterm_input/src/input.rs
+++ b/crossterm_input/src/input.rs
@@ -98,6 +98,7 @@ pub enum KeyEvent {
     End,
     PageUp,
     PageDown,
+    Tab,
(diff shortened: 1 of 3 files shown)


TimonPost:
Somehow I don't see the `/r` added back, you probably did something wrong with your commits

PR #242, opened 2019-09-21
Unit tests

TimonPost:
- Increased get terminal size by one (corrected [change](https://github.com/TimonPost/crossterm/commit/41de9a63a1a1ac768c225fa2328fdc2890f3171d#diff-f6a97d614303969c5f5116bdd1f0fac4L9))

- Decreased set terminal size windows by 1. Windows starts counting from 0, when setting the size to 5, 5 windows will see that as 6,6, therefore we need to subtract 1.

- removed ignores from test

### DIFF
--- a/crossterm_cursor/src/cursor/ansi_cursor.rs
+++ b/crossterm_cursor/src/cursor/ansi_cursor.rs
@@ -110,7 +110,6 @@ mod tests {
 
     // TODO - Test is ingored, because it's stalled on Travis CI
     #[test]
-    #[ignore]
     fn test_ansi_save_restore_position() {
         if try_enable_ansi() {
             let cursor = AnsiCursor::new();
@@ -130,7 +129,6 @@ mod tests {
 
     // TODO - Test is ingored, because it's stalled on Travis CI
     #[test]
-    #[ignore]
     fn test_ansi_goto() {
         if try_enable_ansi() {
             let cursor = AnsiCursor::new();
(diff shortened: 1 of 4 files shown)


crossterm_cursor/src/cursor/ansi_cursor.rs
```diff
@@ -130,7 +129,6 @@ mod tests {
 
     // TODO - Test is ingored, because it's stalled on Travis CI
     #[test]
-    #[ignore]
```
zrzka: These don't work on TCI, should be disabled. See the comment above (in the code).

crossterm_terminal/src/terminal/ansi_terminal.rs
```diff
@@ -85,20 +85,19 @@ mod tests {
     use super::{AnsiTerminal, ITerminal};
 
     /* ======================== ANSI =========================== */
-    // TODO - Test is disabled, because it's failing on Travis CI
     #[test]
-    #[ignore]
```
zrzka: This doesn't work on TCI, should be disabled. See the comment above (in the code).

crossterm_terminal/src/terminal/winapi_terminal.rs
```diff
@@ -128,8 +128,12 @@ impl ITerminal for WinApiTerminal {
             resize_buffer = true;
         }
 
+        println!("{:?}", new_size);
```
zrzka: Should be removed.

crossterm_cursor/src/cursor/ansi_cursor.rs
TimonPost: Yea I saw the comments, just wanted to try it again. Unfortunately...

zrzka:
BTW you can run ignored tests locally by adding `--ignored` to `cargo test` (after `--`).

zrzka:
On mobile, can't find the Approve button, so, approved :D

PR #243, opened 2019-09-22
Crossterm 0.11

TimonPost:
# 1

- [X] Updated crossterm Changelog 
- [x] Updated crossterm_cursor Changelog 
- [x] Updated crossterm_screen Changelog 
- [x] Updated crossterm_terminal Changelog 
- [x] Updated crossterm_style Changelog 
- [x] Updated crossterm_utils Changelog 
- [x] Updated crossterm_winapi Changelog 

# 2
- [X] Increased version number crossterm
- [X] Increased version number crossterm_cursor 
- [X] Increased version number crossterm_screen 
- [X] Increased version number crossterm_terminal 
- [X] Increased version number crossterm_style 
- [X] Increased version number crossterm_utils 
- [X] Increased version number crossterm_winapi 
- Updated version numbers sub dependencies
    Will happen when crossterm is going to be released, needs to be done while releasing step by step.

# 3
- [x] Update README crossterm
- [x] Update README crossterm_cursor 
- [x] Update README crossterm_screen 
- [x] Update README crossterm_terminal 
- [x] Update README crossterm_style 
- [x] Update README crossterm_utils 
- [x] Update README crossterm_winapi

PR #248, opened 2019-09-23
Cursor refactoring

zrzka:
* Internal refactoring
  * No breaking changes except one - `sys` module is private
* Forbids `unused_must_use` warning -> error
* Improves documentation
  * Only the part available at docs.rs
  * Internal documentation not touched

You can check the documentation via `cd crossterm_cursor && cargo doc --open`.


crossterm_cursor/src/cursor/windows.rs
```diff
@@ -4,28 +4,28 @@
...
 
-use crate::sys::winapi::{Cursor, Handle};
+use crate::sys::windows::ScreenBufferCursor;
 
-use super::ITerminalCursor;
+use super::Cursor;
 
 /// This struct is a windows implementation for cursor related actions.
-pub struct WinApiCursor;
+pub(crate) struct WindowsCursor;
```
TimonPost: I would like this to be changed back. `WinApiCursor` was named like that because I like to higlight 'winapi' . On the `AnsiCursor` i'd like to highlight that it used 'ansi'. Maybe in the future, we might have another implementation. And this name would be misleading because `AnsiCuror' works on windows as well, and my naming the cursor of this window you might get the idea that this cursor is always used by windows.

crossterm_cursor/src/cursor/ansi.rs
```diff
@@ -6,47 +6,47 @@ use crossterm_utils::{csi, write_cout, Result};
...
     format!(csi!("{}D"), count)
 }
 
-pub static SAVE_POS_ANSI: &'static str = csi!("s");
-pub static RESTORE_POS_ANSI: &'static str = csi!("u");
-pub static HIDE_ANSI: &'static str = csi!("?25l");
-pub static SHOW_ANSI: &'static str = csi!("?25h");
-pub static BLINK_ON_ANSI: &'static str = csi!("?12h");
-pub static BLINK_OFF_ANSI: &'static str = csi!("?12l");
+pub(crate) static SAVE_POS_CODE: &'static str = csi!("s");
```
TimonPost: I am a bit unsure about this change, like what do we mean with 'code'? I would like to call it or `ansi_code` or `code`. What do you think of that?

crossterm_cursor/src/lib.rs
```diff
@@ -1,13 +1,449 @@
...
+            Box::new(WindowsCursor::new()) as Box<(dyn Cursor + Sync + Send)>
+        };
+
+        #[cfg(unix)]
+        let cursor = AnsiCursor::new();
+
+        TerminalCursor { cursor }
+    }
+
+    /// Moves the cursor to the given position.
```
TimonPost: take it or leave it

```suggestion
    /// Moves the cursor to the given position (column, row).
```

crossterm_cursor/src/lib.rs
```diff
@@ -1,13 +1,449 @@
...
+    pub fn goto(&self, x: u16, y: u16) -> Result<()> {
+        self.cursor.goto(x, y)
+    }
+
+    /// Gets the cursor position.
```
TimonPost:
```suggestion
    /// Gets the cursor position (column, row)
```

crossterm_cursor/src/lib.rs
```diff
@@ -1,13 +1,449 @@
...
+    /// See the [hide](struct.TerminalCursor.html#method.hide) method.
+    pub fn show(&self) -> Result<()> {
+        self.cursor.show()
+    }
+
+    /// Enables or disables the cursor blinking.
+    ///
+    /// # Notes
+    ///
+    /// Not all terminals do support this functionality. Windows versions lower than
```
TimonPost:
```suggestion
    /// Windows versions lower than
```

crossterm_cursor/src/lib.rs
```diff
@@ -1,13 +1,449 @@
...
+    /// Windows 10 do not support this functionality.
+    pub fn blink(&self, blink: bool) -> Result<()> {
+        self.cursor.blink(blink)
+    }
+}
+
+/// Creates a new terminal cursor.
```
TimonPost:
```suggestion
/// Creates a new `Terminal Cursor`.
```

crossterm_cursor/src/cursor/windows.rs
zrzka: Okay

crossterm_cursor/src/lib.rs
zrzka: Why this one? TerminalCursor says _A terminal cursor._  And a `new` says _Creates a new terminal cursor.` Since we're documenting `TerminalCursor`, why to use backticks here?

crossterm_cursor/src/cursor/windows.rs
zrzka: Fixed.

crossterm_cursor/src/lib.rs
zrzka: Fixed by renaming arguments to `column` & `row`.

crossterm_cursor/src/lib.rs
zrzka: Fixed here & on other places where the same sentence was used.

crossterm_cursor/src/lib.rs
zrzka: Fixed.

crossterm_cursor/src/cursor/ansi.rs
zrzka: What about now? Pushed a change. They end with `_CSI_SEQUENCE` & `_csi_sequence`.

TimonPost: Yea good alternatative

crossterm_cursor/src/lib.rs
TimonPost: Because that is returned from this function, I see this function as a constructor/factor function that returns some instance

crossterm_cursor/src/lib.rs
TimonPost: actually that is way better than x, y

crossterm_cursor/src/lib.rs
zrzka: Pushed.

crossterm_terminal/src/terminal/winapi_terminal.rs
```diff
@@ -226,7 +226,7 @@ pub fn clear_entire_screen(buffer_size: Size, current_attribute: u16) -> Result<
     clear(start_location, cells_to_write, current_attribute)?;
 
     // put the cursor back at cell 0,0
-    let cursor = Cursor::new()?;
+    let cursor = TerminalCursor::new();
```
TimonPost: We can use `ScreenBufferCursor` directly because we are using winapi here and thus there is no need to go through the abstraction of ansi/winapi

TimonPost:
Should I wait for the series of refactors concerning the 'I' before doing the 0.11 release?

crossterm_terminal/src/terminal/winapi_terminal.rs
zrzka: Disagree here as it's an implementation detail of the `crossterm_cursor` crate (private `sys` module).

zrzka:
> Should I wait for the series of refactors concerning the 'I' before doing the 0.11 release?

No need to wait because this refactoring is internal and doesn't change public API (except closing `sys` modules). Assuming similar situation is in other crates.

TimonPost:
Let's move this over to 0.11.1, see #243, let's do the 0.11 release tomorrow.

zrzka:
It's impossible to say if anyone is using it or not. It's a breaking change of the `crossterm_cursor` crate only, not the `crossterm` crate, because it doesn't reexport sys modules.

Anyway, this wasn't meant to be included in 0.11 release. It's mainly about cursor doc & refactor and to say if it's okay to use the same schema for other sub crates or if there's anything you'd like to change before I proceed with remaining sub crates.

zrzka:
Closing. Same PR will be against separate repo once I finish separation tmw.

PR #251, opened 2019-09-25
Sub-crates separatation

zrzka:


Cargo.toml
```diff
@@ -12,37 +12,21 @@ readme = "README.md"
...
-    "crossterm_input",
-    "crossterm_screen"
-]
-
-exclude = [
-    "examples/program_examples"
-]
-
 [dependencies]
-crossterm_screen =   { path = "./crossterm_screen", version = "0.3.0"  , optional = true  }
```
TimonPost: I have the following file structure for crossterm:

```
- crossterm-rs
   - crossterm
   - crossterm_cursor
   - etc..
```

Having a path as a reference is easier for development it allows you to change things on the fly. I would prefer to keep `path = "../crossterm_cursor"`, so that the reference to this crate is in the folder `crossterm-rs` and point to `crossterm_cursor`. If we don't add it back, you need to add the path here each time you change something in a subcrate and remove it when creating a PR.

Issue #265, opened 2019-09-30
Merge sub-crates to the crossterm crate


zrzka:
## Goal

Provide all the features we do provide now through the `crossterm` crate only. Archive sub-crates and inform all users.

This does not include `crossterm_winapi`, which can still be used by anyone outside the `crossterm` ecosystem.

We will end up with three active repositories:

* `crossterm-rs/crossterm`
* `crossterm-rs/crossterm-winapi`
* `crossterm-rs/examples`

## Motivation

There're dependencies between sub-crates we would like to avoid. Some of them are also using `sys` module, which needs to be public, but it shouldn't be. It's not part of the public API contract.

There're inconsistencies in examples, book, readmes, ... Some of the use `crossterm` crate with features, some of the use `crossterm-*` sub-crates, ... User can be puzzled what to use. In fact, it doesn't matter what the user will use, but the question is why we do provide two ways to achieve same thing.

And the last one is `crossterm_utils`, which is published as a crate, but it's not useful outside of them `crossterm` ecosystem.

## Timeframe

There's no date set. I would like to focus on phase 1 now to finish the cleanup process I started. Once I finish it, I'll focus on bugs. Then, can be even distant future, discuss with Timon if it's the right time to merge.

## Phase 1 - Cleanup

Improve & polish crates public (docs.rs & README) documentation (@zrzka):

- [x] `cursor` - https://github.com/crossterm-rs/crossterm-cursor/pull/2 & https://github.com/crossterm-rs/crossterm-cursor/pull/7
- [x] `terminal` - https://github.com/crossterm-rs/crossterm-terminal/pull/3
- [x] `screen` - https://github.com/crossterm-rs/crossterm-screen/pull/5
- [x] `style` - https://github.com/crossterm-rs/crossterm-style/pull/2 & https://github.com/crossterm-rs/crossterm-style/pull/3
- [x] `input` - https://github.com/crossterm-rs/crossterm-input/pull/3 & https://github.com/crossterm-rs/crossterm-input/pull/4
- [x] `utils` - https://github.com/crossterm-rs/crossterm-utils/pull/3
- [x] `crossterm` - https://github.com/crossterm-rs/crossterm/pull/269
- [x] Publish new book with info that the book is deprecated & no longer maintained

This includes copying [the book](https://github.com/crossterm-rs/crossterm/tree/master/docs/mdbook) chapters to their respective `lib.rs` modules.

## Phase 2 - Book removal

Then remove the book references from:

[...]

PR #266, opened 2019-10-01
Fix an URL to example file

Detegr:
Fix an invalid URL in README.md. The current version points to `key_events.rs` example as it used to do. Should it point to the example repo instead?


zrzka:
Hi, thanks! Yes, you're right.

Currently, we're in the middle of the documentation update (API & README). You can check what's going on in this issue https://github.com/crossterm-rs/crossterm/issues/265. Documentation of the `crossterm` is in the queue and will be updated this week. Depends on how fast I can update other sub-crates.

But it won't hurt if I merge this even now. Thanks!

PR #280, opened 2019-10-22
Merge crates

zrzka:
## Changes

* GH Actions & Travis CI
  * Test all features
  * Test feature by feature to check that `Cargo.toml` feature dependencies works
* Copy & paste of all the crates code to a module
  * Example `crossterm_cursor` -> `src/cursor` module
  * All these modules are public because of the modules documentation
  * All import paths updated
* `src/lib.rs` contains the Command API documentation from the trashed book
* No public API breaking changes
* README updated
  * Removed links to repos, API documentation, ... for all the sub-crates
* Version bumped to 0.12.1
* Changelog updated

## Tested

- [x] All examples on macOS
- [x] All examples on Linux
- [x] All examples on Windows

```diff
diff --git a/examples/Cargo.toml b/examples/Cargo.toml
index f363119..71ee701 100644
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -12,4 +12,4 @@ readme = "README.md"
 edition = "2018"
 
 [dependencies]
-crossterm = { version = "0.12" }
+crossterm = { path = "../../crossterm" } 
diff --git a/first-depth-search/Cargo.toml b/first-depth-search/Cargo.toml
index 2a5e03f..a944357 100644
--- a/first-depth-search/Cargo.toml
+++ b/first-depth-search/Cargo.toml
@@ -12,5 +12,5 @@ readme = "README.md"
 edition = "2018"
 
 [dependencies]
-crossterm = { version = "0.12" }
+crossterm = { path = "../../crossterm" }
 rand = "0.4"
diff --git a/snake/Cargo.toml b/snake/Cargo.toml
index 236d196..8fccd64 100644
--- a/snake/Cargo.toml
+++ b/snake/Cargo.toml
@@ -12,5 +12,5 @@ readme = "README.md"
 edition = "2018"
 
 [dependencies]
-crossterm = { version = "0.12" }
+crossterm = { path = "../../crossterm" }
 rand = "0.4"
```

### DIFF
--- a/.github/workflows/crossterm_test.yml
+++ b/.github/workflows/crossterm_test.yml
@@ -47,8 +47,23 @@ jobs:
     - name: Test Build
       run: cargo build
       continue-on-error: ${{ matrix.can-fail }}
-    - name: Run Tests
+    - name: Test all features
       run: cargo test --all-features -- --nocapture --test-threads 1
+      continue-on-error: ${{ matrix.can-fail }}
+    - name: Test cursor Feature
+      run: cargo test --features cursor -- --nocapture --test-threads 1
+      continue-on-error: ${{ matrix.can-fail }}
+    - name: Test style Feature
+      run: cargo test --features style -- --nocapture --test-threads 1
+      continue-on-error: ${{ matrix.can-fail }}
+    - name: Test terminal Feature
+      run: cargo test --features terminal -- --nocapture --test-threads 1
+      continue-on-error: ${{ matrix.can-fail }}
+    - name: Test screen Feature
+      run: cargo test --features screen -- --nocapture --test-threads 1
+      continue-on-error: ${{ matrix.can-fail }}
+    - name: Test input Feature
+      run: cargo test --features input -- --nocapture --test-threads 1
       continue-on-error: ${{ matrix.can-fail }}
     - name: Test Packaging
       if: matrix.rust == 'stable'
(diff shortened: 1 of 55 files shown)


.github/workflows/crossterm_test.yml
```diff
@@ -47,9 +47,24 @@ jobs:
     - name: Test Build
       run: cargo build
       continue-on-error: ${{ matrix.can-fail }}
-    - name: Run Tests
+    - name: Test all features
       run: cargo test --all-features -- --nocapture --test-threads 1
```
TimonPost: Why do we test all features separately after testing all of them here?

.travis.yml
```diff
@@ -34,3 +34,8 @@ script:
   - if [ "$TRAVIS_RUST_VERSION" = "stable" ]; then cargo fmt --all -- --check; fi
   - cargo build
   - cargo test --all-features -- --nocapture --test-threads 1
```
TimonPost: Same question?

.github/workflows/crossterm_test.yml
zrzka: Terminal auto enables Cursor feature for example. With all features test, we enable everything and test it. But if anyone makes a mistake in Cargo.toml, terminal wont auto enable cursor, it will still pass for all features, but will fail if user wants a terminal only. In other words, we have to test all features and one by one to catch these kind of errors.

TimonPost:
Ran all tests on windows

PR #281, opened 2019-10-22
Make `KeyEvent` implement Copy

Canop:
There's no reason against it and it eases passing key events around.


TimonPost:
Thanks!

TimonPost:
LGTM!

Issue #276, opened 2019-10-18
Determine the terminal size in a subshell (unix)


Canop:
Like all other libraries I've checked, crossterm fails to read the terminal size when the crossterm based program is ran in a subshell.

I've found a solution which works in my tests: call `tput cols` and `tput lines` and parse the result:
```
fn tput_value(arg: &str) -> Option<u16> {
    match process::Command::new("tput").arg(arg).output() {
        Ok(process::Output{stdout, ..}) => {
                let value = stdout.iter()
                    .map(|&b| b as u16)
                    .take_while(|&b| b>=48 && b<=58)
                    .fold(0, |v, b| v*10 + (b-48));
                if value > 0 { Some(value) } else { None }
        }
        _ => None,
    }
}

pub fn terminal_size() -> (u16, u16) {
    let mut size = Terminal::new().size();
    #[cfg(unix)]
    {
        if size.is_err() {
            if let (Some(w), Some(h)) = (tput_value("cols"), tput_value("lines")) {
                return (w, h);
            }
        }
    }
    ...
```

I'm willing to post a PR.

Check-list before the PR:
- [ ] wait for the various crossterm-rs subcrate to be merged
- [ ] try to find if there's a similar problem (and solution) on Windows
- [ ] decide a name and return type for this function (which should probably be consciously called as it starts new processes)

(In the meantime I've included the function in termimad which will let me run more tests)

zrzka:
I know that the `tput` is with us for a long time. As a safety measure, are we aware of any UNIX like platform (`#[cfg(unix)]`), which doesn't have `tput`?

This approach is pragmatic and I'm fine with it. Although, do we know how the `tput` works? I mean, is there're an easy way how to achieve the same thing without `process::Command::new("tput")`?

joshtriplett:
Invoking a separate program has *many* problems, not least of which that it won't work if the program runs in a container or chroot.

`tput` typically invokes the `TIOCGWINSZ` ioctl, which returns the terminal size.

Canop:
@joshtriplett I agree that invoking a separate program is problematic (maybe use it in a separate function) but you can easily check in a subshell that this solution works while the TIOCGWINSZ ioctl doesn't (and doesn't work either in a container).

TimonPost:
This issue is fixed with #283, if not, re-open.

joshtriplett:
> you can easily check in a subshell that this solution works while the TIOCGWINSZ ioctl currently doesn't

[...]

PR #283, opened 2019-10-22
add a tput based computation of terminal size

Canop:
It's useful, and automatically called, when in a subshell (in linux).

Related to issue https://github.com/crossterm-rs/crossterm/issues/276

It has been tested in a "normal" terminal and in a subshell, but *only on my linux/ubuntu computer*.

It would probably be wise to have more tests before merging this into master.


zrzka:
Sorry for the delay. It works for me.

Although, I like it as a pragmatic solution, but I'm not really sure if crossterm should spin another process because of the `tput`.

Leaving this for @TimonPost as he's the code owner. If you are okay with it, feel free to merge.

Canop:
The concern is legitimate. It could be another function, if we prefer to fail rather than start a process.

An alternative would be to try read the code of tput and understand why it works and how to emulate it without a process. But IMO this could be done later.

TimonPost:
I am not familiar with TPUT, if you say that it works better in more scenarios I am fine with merging it.  One question, it works based on stdout as I can see in your PR, how does this work when raw mode is enabled?

Canop:
I successfully use this function to get the terminal size in my applications running on stderr in a subshell. There's no leak to stdout.

Issue #285, opened 2019-10-23
Remove old the old API and keep the command API


TimonPost:
Related: #268, https://github.com/crossterm-rs/crossterm/issues/249#issuecomment-535023202

# Task
Only keep the Command API, and remove the structures: `TerminalCursor`, `TerminalStyle`, `Terminal`, `Crossterm`.

Motivations behind this:
- User has control over when to execute a command (flush ANSI-sequence to buffer) 
- Better Performance (because of manual flush)
- Complete control over where the ANSI escape commands are executed to
- Less maintenance
- Reducing the number of ways how you can use crossterm (`cursor().goto(), TerminalCursor::new().goto(), println!("{}", Goto()); io::stdout.execute(Goto()`); by focusing on one API we can create better code reduce the maintenance and duplicated documentation and will make it easier for the user.
- This API was made because of old use cases that don't apply anymore

Functions that can't be transformed into an command:

```
Cursor::pos()
RawScreen::into_raw_mode()
RawScreen::disable_raw_mode()
Terminal::size()
```

Let's do the removal one module per PR. 

# Todo
- [x] Terminal Cursor
     - issue: #292 
    - PR: #289 
- [x] Terminal Style
    - issue: #286 
    - PR: #291
- [x] Terminal
       PR: #293 
- [x] Remove `Crossterm`

PR #289, opened 2019-10-23
Removed old Cursor Api

TimonPost:
Fixes task in #285 and #292 

- Removed `TerminalCursor`, `cursor`, `Crossterm::cursor()`
- Introduces static function `crossterm::cursor::position` in place of `TerminalCursor::pos`
- Rename `Goto` to `MoveTo`
- Rename `Up` to `MoveLeft`
- Rename `Right` to `MoveRight`
- Rename `Down` to `MoveDown`
- Rename `BlinkOn` to `EnableBlinking`
- Rename `BlinkOff` to `DisableBlinking`
- Rename `ResetPos` to `ResetPosition`
- Rename `SavePos` to `SavePosition`
- Removed re-export cursor module types at root level, are now accessible from `crossterm::cursor`
- Removed: `/cursor/cursor/winapi`, `/cursor/cursor/ansi`, `/cursor/cursor/cursor.rs`
- Refactored documentation to have a plural beginning and is action-oriented. "Disables ..."
- Removed some unnecessarily system interaction code. 
- Added unit tests (for the winapi logic we have extra unit tests because winapi will not be used trough our commants with the CI. 

-/cursor
|--  sys.rs
|--  ansi.rs
|-- /sys
|---- windows.rs
|---- unix.rs

## Questionable things
- Maybe we should move all ANSI codes into a single file.


zrzka:
This PR and the Terminal one was made in a hurry IMHO. There're couple of things that weren't discussed or decision wasn't made.

For example ... We have a command `Down`. In the `crossterm::Down` context, it doesn't make sense. Down what exactly? It doesn't make sense even in the `crossterm::cursor::Down` context. Slightly better, but not good. It should be either:

* `crossterm::cursor::MoveDown` & no root reexport,
* `crossterm::MoveCursorDown` & private `cursor` module.

These root reexports are not good IMHO. They don't help much, they are prone to collisions, ...

Same applies to the function `crossterm::pos`. What is it? Screen buffer (scroll) position? Cursor position? Mouse position?

I think we should stop for a while and think about the API. Those quick changes make it worse IMHO.

From what I saw so far, I think that we should do the following:

* No root reexports
  * Just public module like `cursor`
  * Commands/functions are accessible via `crossterm::cursor::...`, but not via `crossterm::...`
* Naming
  * All commands should start with the verb - `MoveDown`, `MoveUp`, `GoTo` or `MoveTo`, `RestorePosition`, `SavePosition`, `MoveLeft`, `MoveRight`, `Show`, `Hide`, `EnableBlinking`, `DisableBlinking`
  * All commands should not use abbreviation because of the readability & clarity
    * Everyone has an autocompletion and the length is not important
  * Same applies for the function like `pos` -> `position`

In the end, user can do the following:

```rust
use crossterm::cursor;
let pos = cursor::position()?;
```

... or ...

```rust
use crossterm::cursor::position;
let pos = position()?;
```

... if the user goes with the 2nd option, it's not our problem. We do provide consistent naming and access via module. So, it's clear that it's a `cursor::position`.

This way allows to keep the `cursor` module public, which means clear boundary, module documentation at the docs.rs site (not everything in the `lib.rs` in case of private module), etc.

TimonPost:
The scope of this PR wasn't to fix the know naming concerns discussed in the other issue. This PR only removes the old api. I will make this wip since it is not done yet, forgot to add that. Or do you think we should do all the renaming, export changes in this PR as well?

TimonPost:
After discussing in discord, this PR is going to introduce command refactor/rename as well. Will update this tomorrow. If we agree ton changes in this PR we can apply the same changes to other prs. In the same time I'm going to construct a list of tasks that are done for this PR to make it easier to apply the same changes to the other modules.

TimonPost:
@zrzka it's ready for review

CHANGELOG.md
```diff
@@ -1,7 +1,19 @@
-# Next Version
+# Next Version 13.0
```
zrzka: Would be nice if we can ...

* Use just master in the changelog version, we never know what's going to happen and the version can be misleading, we can make a branch from the latest release tag, fix a bug, introduce new thing and release as 0.13 and then the 0.13 in the master's changelog must be update to something else, like 0.14, etc.
* Split changes to logical group (by modules for example).
* Do not mention documentation changes.
* Do not use past tense, just `Remove` instead of `Removed`, etc.

Proposal ...

```text
# master

- `input` module
  - Derive 'Copy' for 'KeyEvent'
- `cursor` module
  - Remove `TerminalCursor`, `cursor`, `Crossterm::cursor()`
  - Remove root re-exports of the `cursor` module types, accessible via `crossterm::cursor` only
  - Introduce `crossterm::cursor::position` as a `TerminalCursor::pos` replacement
  - Rename `Goto` command to `MoveTo`
  - Rename `Up` command to `MoveLeft`
  - Rename `Right` command to `MoveRight`
  - Rename `Down` command to `MoveDown`
  - Rename `BlinkOn` command to `EnableBlinking`
  - Rename `BlinkOff` command to `DisableBlinking`
  - Rename `ResetPos` command to `ResetPosition`
  - Rename `SavePos` command to `SavePosition`
```

src/cursor.rs
```diff
@@ -1,247 +1,82 @@
...
-//!     cursor.blink(true)?;
-//!
-//!     // Be a good citizen, cleanup
-//!     cursor.blink(false)?;
-//!     cursor.restore_position()
-//! }
-//! ```
-//!
-//! Commands:
+//! Cursor actions can be performed with commands.
```
zrzka: Add empty line after the heading. It's more readable aka the heading is more visible. It also follows already existing documentation style.

src/cursor.rs
```diff
@@ -1,247 +1,82 @@
...
+//! Please have a look at [command documention](../index.html) for a more detailed documentation.
 //!
 //! ```no_run
 //! use std::io::{stdout, Write};
 //!
-//! use crossterm::{BlinkOff, BlinkOn, execute, Goto, ResetPos, Result, SavePos};
-//!
+//! use crossterm::{ExecutableCommand, execute, Result, cursor::{DisableBlinking, EnableBlinking, MoveTo, RestorePosition, SavePosition}};
```
zrzka: Line too long, see https://github.com/crossterm-rs/crossterm/blob/master/docs/CONTRIBUTING.md#max-line-length

src/cursor.rs
```diff
@@ -1,247 +1,82 @@
...
 //! }
 //! ```
-use cursor::ansi::{self, AnsiCursor};
-#[cfg(windows)]
-use cursor::windows::WinApiCursor;
-use cursor::Cursor;
+//!
+//! For manual execution control check out [crossterm::queue](../macro.queue.html).
+
+pub use sys::position;
```
zrzka: Wrong semantic ordering ...

```diff
@@ -41,11 +41,10 @@
 pub use sys::position;
 
 use crate::impl_display;
+use crate::utils::Command;
 #[cfg(windows)]
 use crate::utils::Result;
 
-use crate::utils::Command;
-
 mod ansi;
 pub(crate) mod sys;
 ```

File: src/cursor.rs [hunk 5]
```diff
@@ -447,14 +274,90 @@ impl Command for BlinkOff {
...
 impl_display!(for Hide);
 impl_display!(for Show);
-impl_display!(for BlinkOn);
-impl_display!(for BlinkOff);
+impl_display!(for EnableBlinking);
+impl_display!(for DisableBlinking);
+
+#[cfg(test)]
+mod tests {
+    use super::{
```
zrzka: Wrong semantic ordering ...

```diff
@@ -288,11 +287,13 @@ impl_display!(for DisableBlinking);
 
 #[cfg(test)]
 mod tests {
+    use std::io::{self, stdout, Write};
+
+    use crate::execute;
+
     use super::{
         position, MoveDown, MoveLeft, MoveRight, MoveTo, MoveUp, RestorePosition, SavePosition,
     };
-    use crate::execute;
-    use std::io::{self, stdout, Write};
 
     // Test is disabled, because it's failing on Travis
     #[test]
```

src/cursor.rs
```diff
@@ -1,247 +1,82 @@
...
-}
-
-/// A command to move the cursor to the given position.
+/// A command that moves the terminal cursor to the given position (column, row).
 ///
 /// # Notes
-///
-/// Commands must be executed/queued for execution otherwise they do nothing.
-pub struct Goto(pub u16, pub u16);
+/// - The counting of the given dimensions starts from 0 were column 0 and row 0 is the top left.
```
zrzka: * Empty line after the heading.
* Just a nitpick, we do use `*` for the unnumbered lists in the code.
* There's typo `were` -> `where`

Can we make it shorter?

* The counting of the given dimensions starts from 0 were column 0 and row 0 is the top left.

... to ...

* Top left cell is represented as `0, 0`.

src/cursor.rs
```diff
@@ -290,18 +125,27 @@ impl Command for Left {
...
 /// # Notes
 ///
 /// Commands must be executed/queued for execution otherwise they do nothing.
-pub struct Right(pub u16);
+pub struct MoveRight(pub u16);
+
+/// A command that saves the current terminal cursor position.
+///
+/// See the [RestorePosition](./struct.RestorePosition.html) command.
+/// # Notes
```
zrzka: Empty line before heading.

src/cursor.rs
```diff
@@ -333,18 +167,19 @@ impl Command for SavePos {
...
 
-/// A command to restore the saved cursor position.
+/// A command that restores the saved terminal cursor position.
 ///
+/// See the [SavePosition](./struct.SavePosition.html) command.
 /// # Notes
-///
-/// Commands must be executed/queued for execution otherwise they do nothing.
-pub struct ResetPos;
+/// - The cursor position is stored globally.
```
zrzka: Empty line before & after the heading.

src/cursor.rs
```diff
@@ -353,18 +188,15 @@ impl Command for ResetPos {
 
     #[cfg(windows)]
     fn execute_winapi(&self) -> Result<()> {
-        WinApiCursor::new().restore_position()
+        sys::restore_position()
     }
 }
 
-/// A command to hide the cursor.
+/// A command that hides the terminal cursor indicator.
```
zrzka: Can we remove the `indicator` word? We don't use it anywhere in the docs. It was added in this PR for just two commands `Hide` & `Show`.

src/cursor.rs
```diff
@@ -376,18 +208,15 @@ impl Command for Hide {
 
     #[cfg(windows)]
     fn execute_winapi(&self) -> Result<()> {
-        WinApiCursor::new().hide()
+        sys::show_cursor(false)
     }
 }
 
-/// A command to show the cursor.
+/// A command that shows the terminal cursor indicator.
```
zrzka: Can we remove the `indicator` word? We don't use it anywhere in the docs. It was added in this PR for just two commands `Hide` & `Show`.

src/cursor/sys/unix.rs
```diff
@@ -1,42 +1,34 @@
...
 use crate::input::{InputEvent, TerminalInput};
 use crate::utils::{
     sys::unix::{disable_raw_mode, enable_raw_mode, is_raw_mode_enabled},
     Result,
 };
-use crate::{csi, write_cout};
 
-pub(crate) fn get_cursor_position() -> Result<(u16, u16)> {
+/// Returns the cursor position (column, row).
+/// The counting starts from 0 were column 0 and row 0 is the top left.
```
zrzka: Empty doc comment line between these two lines. Also I'd use the same wording from the previous comment:

Top left cell is represented as `0, 0`.

Short and clear enough.

src/cursor/sys/windows.rs
```diff
@@ -14,7 +14,13 @@ use lazy_static::lazy_static;
 
 use crate::utils::Result;
 
-pub(crate) fn get_cursor_position() -> Result<(u16, u16)> {
+lazy_static! {
+    static ref SAVED_CURSOR_POS: Mutex<Option<(i16, i16)>> = Mutex::new(None);
+}
+
+/// Returns the cursor position (column, row).
```
zrzka: Empty doc comment line between these two lines. Also I'd use the same wording from the previous comment:

Top left cell is represented as `0, 0`.

Short and clear enough.

src/lib.rs
```diff
@@ -117,7 +117,7 @@
 //!
 //! ```no_run
 //! use std::io::{stdout, Write};
-//! use crossterm::{ExecutableCommand, QueueableCommand, Color, PrintStyledFont, Colorize, Clear, ClearType, Goto, Result};
+//! use crossterm::{ExecutableCommand, QueueableCommand, Color, PrintStyledFont, Colorize, Clear, ClearType, cursor::MoveTo, Result};
```
zrzka: Long line. Horizontal scrollbar in the documentation, because it shows just:

```
use crossterm::{ExecutableCommand, QueueableCommand, Color, PrintStyledFont, Colorize, Clear, Cl
```

src/lib.rs
```diff
@@ -142,7 +142,7 @@
 //!
 //! ```no_run
 //! use std::io::{stdout, Write};
-//! use crossterm::{execute, queue, Color, PrintStyledFont, Colorize, Goto, Clear, ClearType, Result};
+//! use crossterm::{execute, queue, Color, PrintStyledFont, Colorize, cursor::MoveTo, Clear, ClearType, Result};
```
zrzka: Long line. Horizontal scrollbar in the documentation, because it shows just:

```
use crossterm::{execute, queue, Color, PrintStyledFont, Colorize, cursor::MoveTo, Clear, ClearTy
```

src/cursor.rs
```diff
@@ -1,247 +1,82 @@
...
-//!
-//!     // Be a good citizen, cleanup
-//!     cursor.blink(false)?;
-//!     cursor.restore_position()
-//! }
-//! ```
-//!
-//! Commands:
+//! Cursor actions can be performed with commands.
+//! Please have a look at [command documention](../index.html) for a more detailed documentation.
```
zrzka: `../index.html#command-api`

CHANGELOG.md
zrzka: Unresolved, check my proposal. Backticks in around Copy, ... rename vs renamed vs refactored vs ..., ... Also the doc note is probably not useful.

PR #296, opened 2019-10-29
introduce the EnableMouseMode and DisableMouseMode commands

Canop:
Those commands allow using the mouse in applications running in stderr without leaking on stdout.

They have been tested on linux (most notably in broot).

No test has been done yet on windows, and the windows specific code should be reviewed and maybe improved.

Global documentation of the input module should probably be augmented too.


CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 - `input` module
     - Derive 'Copy' for 'KeyEvent'
+    - add the EnableMouseMode and DisableMouseMode commands
```
TimonPost:
```suggestion
    - Add the EnableMouseMode and DisableMouseMode commands
```

src/input.rs
```diff
@@ -423,3 +424,37 @@ impl TerminalInput {
 pub fn input() -> TerminalInput {
     TerminalInput::new()
 }
+
+/// A command that enables mouse mode
```
TimonPost:
```suggestion
/// A command that enables mouse event's to be monitored. 
///
/// Mouse events will be produced by the
/// [`AsyncReader`](struct.AsyncReader.html)/[`SyncReader`](struct.SyncReader.html).
```

src/input.rs
```diff
@@ -423,3 +424,37 @@ impl TerminalInput {
...
+        ansi::enable_mouse_mode_csi_sequence()
+    }
+
+    #[cfg(windows)]
+    fn execute_winapi(&self) -> Result<()> {
+        input().enable_mouse_mode()
+    }
+}
+
+/// A command that disables mouse mode
```
TimonPost:
```suggestion
/// A command that disables mouse event's to be monitored. 
///
/// Mouse events will be produced by the
/// [`AsyncReader`](struct.AsyncReader.html)/[`SyncReader`](struct.SyncReader.html).
```

src/input/ansi.rs
```diff
@@ -0,0 +1,23 @@
+//! This module provides input related ANSI escape codes.
+
+use crate::csi;
+
+pub(crate) fn enable_mouse_mode_csi_sequence() -> String {
```
TimonPost: This might as well be a constant right? Then we can remove the string alocation completely. I am not sure if we can make `AnsiType` in command be a `&' static str` then we can get rid of this allocation completely.

src/input.rs
```diff
@@ -423,3 +424,40 @@ impl TerminalInput {
 pub fn input() -> TerminalInput {
     TerminalInput::new()
 }
+
+/// A command that enables mouse mode
+///
+pub struct EnableMouseMode;
```
TimonPost: I would like this to be renamed to `EnableMouseEvents`.

src/input.rs
```diff
@@ -423,3 +424,40 @@ impl TerminalInput {
...
+        input().enable_mouse_mode()
+    }
+}
+
+/// A command that disables mouse event's to be monitored. 
+///
+/// Mouse events will be produced by the
+/// [`AsyncReader`](struct.AsyncReader.html)/[`SyncReader`](struct.SyncReader.html).
+///
+pub struct DisableMouseMode;
```
TimonPost: I would like this to be renamed to `DisableMouseEvents`.

Canop:
It looks like the best names would be `EnableMouseCaputure` and `DisableMouseCapture`.

Opinions welcome.

zrzka:
I'd say that these commands should be named:

* `EnableMouseEvents`
* `DisableMouseEvents`

Reasoning - we will have `Event` in the `input`, `Event::Mouse(MouseEvent)`, ... It appears to me that these names are best. At least, all the names are synced.

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 - `input` module
     - Derive 'Copy' for 'KeyEvent'
+    - Add the EnableMouseMode and DisableMouseMode commands
```
TimonPost:
```suggestion
    - Add the `EnableMouseCapture` and `EnableMouseCapture` commands
```

PR #307, opened 2019-11-03
read_async: join background thread and avoid looping forever on windows

udoprog:
This fixes the issue seen in crossterm-rs/examples#7.

There's an inner loop when creating the thread which will cause it to loop forever, even when it shouldn't. Like when the cancellation token is set or the sender channel errors.

Also joining the thread to more closely tie the `AsyncReader` to the life cycle of the background thread.

### DIFF
--- a/src/input/input/windows.rs
+++ b/src/input/input/windows.rs
@@ -273,6 +273,7 @@ impl Iterator for SyncReader {
 pub struct AsyncReader {
     event_rx: Receiver<InputEvent>,
     shutdown: Arc<AtomicBool>,
+    thread: Option<thread::JoinHandle<()>>,
 }
 
 impl AsyncReader {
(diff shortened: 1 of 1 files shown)


src/input/input/windows.rs
```diff
@@ -289,13 +290,14 @@ impl AsyncReader {
         let (event_tx, event_rx) = mpsc::channel();
         let thread_shutdown = shutdown_handle.clone();
 
-        thread::spawn(move || loop {
```
TimonPost: Very good one! Such a sneaky loop here.

src/input/input/windows.rs
```diff
@@ -273,6 +273,7 @@ impl Iterator for SyncReader {
 pub struct AsyncReader {
     event_rx: Receiver<InputEvent>,
     shutdown: Arc<AtomicBool>,
+    thread: Option<thread::JoinHandle<()>>,
```
TimonPost: Why is this an option?

udoprog: `JoinHandle::join` takes `self`, and `stop` `&mut self` since it's called on drop. This is the easiest way.

PR #309, opened 2019-11-04
Remove reader thread in AsyncReader

udoprog:
We also store the `Console` instance in `AsyncReader`, which means we no
longer have to create it every millisecond.

Note that this won't build until we have a released version of crossterm-rs/crossterm-winapi#7 since it's highly desirable for AsyncReader to be `Send` + `Sync`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# Unreleased
+
+- Remove thread from AsyncReader on Windows ([PR #309](https://github.com/crossterm-rs/crossterm/pull/309))
+
 # Version 0.13.2
 
 - New `input::stop_reading_thread()` function
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,7 @@ serde = { version = "1.0.0", features = ["derive"], optional = true }
 
 [target.'cfg(windows)'.dependencies]
 winapi = "0.3.8"
-crossterm_winapi = { version = "0.3.0", path = "../crossterm-winapi" }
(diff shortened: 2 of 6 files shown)


udoprog:
This is just to give an idea of what this change accomplishes in combination with the change to crossbeam-winapi. Below I've run the same application for a couple of seconds and I'm viewing the results in process explorer (A more detailed task manager application for Windows). On the left, we have the current version of crossterm, and on the right we have this patch;
image

The number of I/O operations is much higher, because the thread runs the polling in a very tight loop (1ms) which is then buffered over the channel.
And the number of handles keep increasing indefinitely because we are not closing them properly.

TimonPost:
That's is a very great improvement! I totally like the improvement of handling the HANDLE in this PR. Although what I was highlighting is that this breaks the way async reader 'supposes' to work and it requires UNIX input to be modified as well.

Instead of removing the thread and keep the same behavior across Windows and Linux. We could increase the thread sleep. The average typing speed is 200 characters a minute which is every 0.3 seconds or 300 milliseconds. This means that if we want to cover the average speed we should not go above 300 ms sleeping because it could cause delays. Although fast typers might have a higher rate so I think we would be fine with 100ms sleep.

TimonPost:
Nice work! Make sure to update the CHANGELOG. I will merge the other PR and do a release.

udoprog:
The build failure seems to be related to cargo using a cached, non-git version of the crossterm-winapi crate. It should fix itself once there is a release to pin it to.

PR #313, opened 2019-11-08
Replace 'An' with 'A' in docs.

cedric-h:
Use "an" if the word following the article begins with a vowel sound, i.e. "early" or "hour". Use "a" if the word following the following the article begins with a consonant sound, i.e. "command" or "hero".

(this is because the 'h' in "hour" is pronounced as if it didn't exist at all, making the word similar to "our", but the 'h' in "hero" is actually emphasized.)


TimonPost:
Thanks!

PR #314, opened 2019-11-08
Allow macro trailing comma

cedric-h:
All of the other Rust macros allow trailing commas, so the fact that crossterm's wouldn't let me have an extra useless comma at the end was bothering me. It really does make it much easier to work with when you have a lot of things being passed to the macro where each are on their own line.

### DIFF
--- a/src/utils/macros.rs
+++ b/src/utils/macros.rs
@@ -114,20 +114,10 @@ macro_rules! queue {
 
 #[test]
 fn test_queue() {
-    use std::io::{stdout, Write};
     use super::command::Output;
-    assert!(
-        queue!(
-            stdout(),
-            Output("hi"),
-        ).is_ok()
-    );
-    assert!(
-        queue!(
-            stdout(),
-            Output("hi"),
-        ).is_ok()
-    );
+    use std::io::{stdout, Write};
+    assert!(queue!(stdout(), Output("hi"),).is_ok());
+    assert!(queue!(stdout(), Output("hi")).is_ok());
 }
 
 /// Execute one or more command(s)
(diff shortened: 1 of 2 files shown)


src/utils/macros.rs
```diff
@@ -201,3 +201,22 @@ macro_rules! impl_from {
...
     };
 }
+
+#[cfg(test)]
+mod test {
+    #[test]
+    fn test_queue() {
+        use crate::utils::command::Output;
+        use std::io::{stdout, Write};
+        assert!(queue!(stdout(), Output("hi"),).is_ok());
```
zrzka: Please, do not use `Output`. We do not want to print anything on `stdout`. Instead of this, create a fake command in the `test` module. Something like:

```rust
pub struct FakeCommand;

impl Command for FakeCommand {
    type AnsiType = &'static str;

    fn ansi_code(&self) -> Self::AnsiType {
        ""
    }

    #[cfg(windows)]
    fn execute_winapi(&self) -> Result<(), ErrorKind> {
        Ok(())
    }
}
```

src/utils/macros.rs
```diff
@@ -201,3 +201,22 @@ macro_rules! impl_from {
         }
     };
 }
+
+#[cfg(test)]
+mod test {
```
zrzka: `mod test` -> `mod tests`

src/utils/macros.rs
```diff
@@ -201,3 +201,37 @@ macro_rules! impl_from {
...
     };
 }
+
+#[cfg(test)]
+mod tests {
+    use crate::utils::command::Command;
+    #[cfg(windows)]
+    use crate::utils::error::ErrorKind;
+    use std::io::{stdout, Write};
+    pub struct FakeCommand;
```
zrzka: Also, please reorder imports in the following way and add an empty line before `FakeCommand`.

```rust
use std::io::{stdout, Write};

use crate::utils::command::Command;
#[cfg(windows)]
use crate::utils::error::ErrorKind;

pub struct FakeCommand;
```

zrzka:
And the last thing, because it's a public API "change", can you please add the following line ...

```
- `queue!` & `execute!` macros allow trailing comma
```

... to the CHANGELOG.md.

TimonPost:
Thanks for the work!

PR #322, opened 2019-11-19
Remove all feature flags except event-stream

### LINKED ISSUE #317: Remove all feature flags
When I look at the feature list ...

```
default = ["cursor", "style", "terminal", "screen", "event"]
cursor = ["lazy_static", "event", "winapi/wincon", "winapi/winnt", "winapi/minwindef"]
style = ["lazy_static", "winapi/wincon"]
terminal = ["cursor"]
screen = ["lazy_static", "winapi/wincon", "winapi/minwindef"]
event = ["mio", "lazy_static", "parking_lot", "screen", "terminal", "winapi/winnt", "winapi/winuser", "winapi/wincon", "winapi/synchapi", "winapi/namedpipeapi", "winapi/impl-default"]
```

... it's getting complicated, feature A requires feature B, ...

I propose to remove all features and make them always available.

Cons?

* Slower compile time
  * Really? How many users do disable default features and enables only selected ones?
* Does not affect size, linker eliminates dead/not used code.

Or less radical change:

```
default = []
event = []
```

Just one feature named `event`. What are the use cases?

* User is either drawing, writing, changing colors, getting size, cursor position, ...
* User is doing the previous stuff + consuming events

In this way we will avoid downloading & compiling `mio`, `parking_lot`, ... if user does not want to consume events.

zrzka:
Fixes #317

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,7 +44,7 @@ serde = { version = "1.0.0", features = ["derive"], optional = true }
 #
 [target.'cfg(windows)'.dependencies.winapi]
 version = "0.3.8"
-features = ["wincon", "minwindef", "winnt", "winuser", "synchapi", "namedpipeapi", "impl-default"]
+features = ["winuser"]
 
 [target.'cfg(windows)'.dependencies]
 crossterm_winapi = "0.5.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,34 +16,50 @@ categories = ["command-line-interface", "command-line-utilities"]
 name = "crossterm"
 path = "src/lib.rs"
 
+#
+# Build documentation with all features -> EventStream is available
+#
 [package.metadata.docs.rs]
 all-features = true
 
+#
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -16,34 +16,50 @@ categories = ["command-line-interface", "command-line-utilities"]
...
+parking_lot = "0.9"
 serde = { version = "1.0.0", features = ["derive"], optional = true }
-futures = {version = "0.3", optional = true }
+
+#
+# Windows dependencies
+#
+[target.'cfg(windows)'.dependencies.winapi]
+version = "0.3.8"
+features = ["wincon", "minwindef", "winnt", "winuser", "synchapi", "namedpipeapi", "impl-default"]
```
TimonPost: Will have to validate this, I think that "synchapi", "namedpipeapi", "impl-default" can be removed.

TimonPost: Updated your PR.

PR #327, opened 2019-11-22
Introduced more cursor commands.

TimonPost:
- Added `MoveToColumn`, `MoveToPreviousLine`, `MoveToNextLine` commands

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
     documentation
 * Replace `docs/UPGRADE.md` with the [Upgrade Paths](https://github.com/crossterm-rs/crossterm/wiki#upgrade-paths)
   documentation
+ - Added `MoveToColumn`, `MoveToPreviousLine`, `MoveToNextLine` commands
  
 # Version 0.13.3
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
     documentation
 * Replace `docs/UPGRADE.md` with the [Upgrade Paths](https://github.com/crossterm-rs/crossterm/wiki#upgrade-paths)
   documentation
- - Added `MoveToColumn`, `MoveToPreviousLine`, `MoveToNextLine` commands
+ - Add `MoveToColumn`, `MoveToPreviousLine`, `MoveToNextLine` commands
  
 # Version 0.13.3
 
--- a/src/cursor.rs
+++ b/src/cursor.rs
@@ -92,6 +92,26 @@ impl Command for MoveUp {
     }
 }
 
+/// A command that moves the terminal cursor a given number of columns to the right.
+///
+/// # Notes
+///
+/// Commands must be executed/queued for execution otherwise they do nothing.
+pub struct MoveRight(pub u16);
+
+impl Command for MoveRight {
+    type AnsiType = String;
+
+    fn ansi_code(&self) -> Self::AnsiType {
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
     documentation
 * Replace `docs/UPGRADE.md` with the [Upgrade Paths](https://github.com/crossterm-rs/crossterm/wiki#upgrade-paths)
   documentation
+ - Added `MoveToColumn`, `MoveToPreviousLine`, `MoveToNextLine` commands
```
zrzka: Past tense ...

src/cursor.rs
```diff
@@ -72,6 +72,66 @@ impl Command for MoveTo {
...
+        ansi::move_to_next_line_csi_sequence(self.0)
+    }
+
+    #[cfg(windows)]
+    fn execute_winapi(&self) -> Result<()> {
+        sys::move_to_next_line(self.0)
+    }
+}
+
+/// A command that moves the terminal cursor down the given number of lines, and resets it to the first column at the beginning of the line.
```
zrzka: * Contributing guidelines violation - line is too long
* _resets it to the first column at the beginning of the line_ - sounds like the same thing is said twice. What about just `and moves it to the first column.`?

src/cursor.rs
```diff
@@ -72,6 +72,66 @@ impl Command for MoveTo {
     }
 }
 
+/// A command that moves the terminal cursor up the given number of lines, and resets it to the first column at the beginning of the line.
```
zrzka: * Contributing guidelines violation - line is too long
* _resets it to the first column at the beginning of the line_ - sounds like the same thing is said twice. What about just `and moves it to the first column.`?

src/cursor/sys/windows.rs
```diff
@@ -206,6 +224,33 @@ mod tests {
         assert_eq!(position().unwrap(), (0, 0));
     }
 
+    #[test]
+    fn test_move_to_next_line_winapi() {
+        move_to(0, 2).unwrap();
```
zrzka: Not necessarily in this PR, because other tests do the same thing. We should wrap all tests with a macro/fn which saves the cursor position, executes the test and then it restores the position. I can imagine that the current behaviour can create a mess from the tests output. Not saying that unit tests shouldn't do these things at all.

TimonPost:
Thanks for the comments, I'll do the updates, sorry I'm not perfect with the English grammar. I certainly care about good documentation, and I try to get better every time. It's just super hard to write good, short and clear comments. Especially since I have Dyslexia.

zrzka:
> Thanks for the comments, I'll do the updates, sorry I'm not perfect with the English grammar.

Me neither :)

> I certainly care about good documentation, and I try to get better every time. It's just super hard to write good, short and clear comments.

I mainly meant stuff like very long lines, etc.

> Especially since I have Dyslexia.

Welcome to the club. The good news is that it's better with age :) I remember all the issues I've got in the school, but they're mostly gone.

TimonPost:
O hahah, I believe that it is common for IT people.  I notice the slow shift and improvement over the last year, at least in my own language, I start to become very good at grammar.

src/cursor/sys/windows.rs
TimonPost: We need to establish a baseline, we can't expect the cursor to be somewhere. I agree that those tests do more then they should to do and like the idea of those macro's. However, we can do those things in other PR's.

PR #328, opened 2019-11-22
Make use of cancellation token when polling

udoprog:
This is just to showcase the API I'm proposing to solve the apparent deadlock in EventStream. I've only implemented it for Windows for now, but the same concept should be possible on unix as well.

This api will allow other threads to cancel a read, which is necessary to properly clean up resources for certain operations.

If it looks good, I can build the unix impl as well.

Needs crossterm-rs/crossterm-winapi#9 before it can be merged.

### DIFF
--- a/src/event/stream.rs
+++ b/src/event/stream.rs
@@ -15,7 +15,8 @@ use std::{
 use crate::Result;
 
 use super::{
-    filter::EventFilter, poll_internal, read_internal, Event, InternalEvent, INTERNAL_EVENT_READER,
+    cancellation, filter::EventFilter, poll_internal, read_internal, CancelRx, CancelTx, Event,
+    InternalEvent,
 };
 
 /// A stream of `Result<Event>`.
@@ -29,19 +30,24 @@ use super::{
 ///
 /// Check the [examples](https://github.com/crossterm-rs/crossterm/tree/master/examples) folder to see how to use
 /// it (`event-stream-*`).
-#[derive(Default)]
 pub struct EventStream {
     wake_thread_spawned: Arc<AtomicBool>,
     wake_thread_should_shutdown: Arc<AtomicBool>,
+    cancel_tx: Option<CancelTx>,
+    cancel_rx: CancelRx,
 }
 
 impl EventStream {
     /// Constructs a new instance of `EventStream`.
-    pub fn new() -> EventStream {
-        EventStream {
+    pub fn new() -> Result<EventStream> {
+        let (cancel_tx, cancel_rx) = cancellation()?;
+
+        Ok(EventStream {
             wake_thread_spawned: Arc::new(AtomicBool::new(false)),
             wake_thread_should_shutdown: Arc::new(AtomicBool::new(false)),
-        }
+            cancel_tx: Some(cancel_tx),
+            cancel_rx,
+        })
     }
 }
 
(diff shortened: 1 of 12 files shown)


src/event/stream.rs
```diff
@@ -29,27 +30,36 @@ use super::{
 ///
 /// Check the [examples](https://github.com/crossterm-rs/crossterm/tree/master/examples) folder to see how to use
 /// it (`event-stream-*`).
-#[derive(Default)]
```
TimonPost: Why do you remove default?

src/event/sys/unix.rs
```diff
@@ -11,6 +12,45 @@ use crate::{
...
+        self.0.write(&[0x57])?;
+        Ok(())
+    }
+}
+
+#[derive(Clone)]
+pub struct CancelRx(pub(crate) Arc<FileDesc>);
+
+/// Construct a cancellation pair.
+pub fn cancellation() -> Result<(CancelTx, CancelRx)> {
```
TimonPost: Is this supposed to call this from outside the scope, maybe pub(crate)?

src/event/stream.rs
udoprog: Had a version where `EventStream` was constructed differently. Fixed!

udoprog: Ah, actually. Construction is no longer fallible, so we can't provide a `Default` impl :(

src/event/sys/unix.rs
udoprog: Not exposing this raises a warning that `CancelRx` is never constructed since it's present in another pub API, albeit in a `Option<&CancelRx>`, so it's typically used with `None`.

udoprog: I bumped the branch, feel free to try it out yourself and see if you can find a better solution!

zrzka:
@udoprog can we please rebase this one? So we can finish the review and have it merged? Also CI is failing on Windows stable.

udoprog:
@zrzka I'm blocked on feedback and a release of `crossterm-winapi` with this commit:
https://github.com/crossterm-rs/crossterm-winapi/commit/f53938b88ed6249aa9a4eabb353d21bec9a5b6bc

zrzka:
@udoprog ah, that's why the packaging fails. @TimonPost can we please release `crossterm-winapi`, so, we can finish this PR?

zrzka:
Okay, I reviewed this one, tested couple of changes locally, played with `cfg-if`s and we have to organise this whole thing in a different way to:

* hide public reexports (`CancelRx` in `event` & docs, ...),
* make it nice for both `event-stream` and no-`event-stream` (minimize `cfg` usage),
* "fix" unused imports, dead code, ... for different situations/features,
* ...

I'll think about it till tomorrow and will provide an alternative based on this PR. The idea is good, but I do not recommend approving & merging it now.

TimonPost:
I released crossterm_winapi 0.5.1

zrzka:
Here's my experiment with this:

* [PR](https://github.com/crossterm-rs/crossterm/pull/339)
* UNIX only
* Windows NOT implemented yet

It's similar to the upcoming [mio::Waker](https://github.com/tokio-rs/mio/blob/master/src/waker.rs).

//cc @TimonPost @udoprog

zrzka:
Did add Windows implementation (**not tested it at all and can be wrong**). See https://github.com/crossterm-rs/crossterm/pull/339.

PR #326, opened 2019-11-21
Refactored the command macro's and rewrote documentation.

TimonPost:
- Removed duplicated logic from queue/execute commands
- Added a macro which contains the duplicated logic.
- Renamed Output to Print and moved it to the sytle module (#320)
- Improved documentation
- Enable/disable mouse capture mouse mode #315

PR #332, opened 2019-11-23
Moved examples into crossterm 

TimonPost:


TimonPost:
- There is one issue, with the cursor commands, that needs to be resolved first. I will open an PR for that. Namely, the fields of those commands are private, and therefore I can't access them. 
- #326 has to be merged first for the `style::Print` change.

Issue #325, opened 2019-11-20
Switch old `AlternateScreen` api to command API.


TimonPost:
We still write to the stdout in the [AnsiAlternateScreen](https://github.com/crossterm-rs/crossterm/blob/master/src/screen/alternate/ansi.rs#L10), we already created [Enter](https://github.com/crossterm-rs/crossterm/blob/master/src/screen.rs#L159)/[LeaveAlternateScreen](https://github.com/crossterm-rs/crossterm/blob/master/src/screen.rs#L195) commands to fix this.

We should update [this ](https://github.com/crossterm-rs/crossterm/blob/master/src/screen.rs#L89)place with commands and remove the `AlternateScreen` trait and both the [windows ](https://github.com/crossterm-rs/crossterm/blob/master/src/screen/alternate/windows.rs) and [ansi ](https://github.com/crossterm-rs/crossterm/blob/master/src/screen/alternate/ansi.rs#L10) implementation.

PR #336, opened 2019-11-27
Refactored Screen Module

TimonPost:
fixes: #325 

 - Rename `screen::Rawscreen::into_raw_mode` to `screen::Rawscreen::enable_raw_mode`
 - Remove `screen::AlternateScreen`


zrzka:
Not sure if I'm late to the party, but here's my take on this PR.

## `screen` module

* I'd remove it completely and move all the functionality to the `terminal` module
* Both these modules do work with the terminal/screen/buffer, whatever name you pick
  * Let's pick buffer for this comment
  * `terminal` allows to scroll buffer up, down and set/get text area size
  * `screen` allows to enable/disable raw mode
  * `screen` allows to enter/leave alternate screen buffer

It seems that we're keeping these two modules for historical reasons. Both provide such a small amount of functionality, both are about +- same thing, ... why not merge them to the `terminal` module?

## `RawScreen`

While the crossterm 0.14 is a major overhaul, let's touch this one as well.

* Is `RawScreen` right name?
* Is `enable_raw_mode` right method name?

To make it clear for the user, there's nothing like raw screen, there's just terminal raw mode which can be enabled and disabled. What about?

```rust
mod terminal {
    pub struct RawMode {
        pub fn enable() -> Result<()>;
        pub fn disable() -> Result<()>;
    }
}
```

Drop semantics gone, no return value to keep, simple. Minimal API user can build anything on top of it.

TimonPost:
> While the crossterm 0.14 is a major overhaul, let's touch this one as well.

Yeah sure, let's rewrite this module. It is a good idea to move the screen into the terminal module. I prefer keeping 

> To make it clear for the user, there's nothing like raw screen, there's just terminal raw mode which can be enabled and disabled. What about?

```rust
   pub struct RawMode {
       pub fn enable() -> Result<()>;
       pub fn disable() -> Result<()>;
   }
```rust

Is there anything against making these functions static  I feel like the `RawMode` is just like a place holder here and does not add any extra value.

**Comment by zrzka** [comment]
> Yeah sure, let's rewrite this module. It is a good idea to move the screen into the terminal module. I prefer keeping either terminal or screen. Because buffer is not the same as terminal or screen. Over screen vs terminal I prefer to keep terminal, because it contains the most types, and will cause fewer API breaking changes.

I prefer `terminal` as well.

> Is there anything against making these functions static I feel like the RawMode is just like a place holder here and does not add any extra value.

Nice, even simpler. We can provide just two functions like `enable_raw_mode` & `disable_raw_mode`.

**Comment by TimonPost** [comment]
> Nice, even simpler. We can provide just two functions like enable_raw_mode & disable_raw_mode.

I will name them enter / leave raw_mode. Then they are in sync with enter/leave alternate mode

**Comment by zrzka** [comment]
> I will name them enter / leave raw_mode. Then they are in sync with enter/leave alternate screen

Why they should be in sync? No strong opinion here, but I feel that the `enable`/`disable` prefix is more appropriate here.

**Comment by TimonPost** [comment]
Why do you think that? I do agree that they don't have to be in sync, however, it is nice to use the same naming if we can, to me enter/leave sounds just as fine as enable/disable. I also don't have a strong opinion here. Let's leave it like this. We should only do things we are sure about.

**Comment by zrzka** [comment]
> Why do you think that?

As I said, no strong opinion, it just sounds better (to me) and also we have `RawScreen::enable_raw_mode` now, so, the conversion will be mainly about removing `RawScreen::` only and it will be more familiar to the current version users.

**Comment by TimonPost** [comment]
Oke, let's keep enable/disable

File: src/terminal.rs [hunk 1]
```diff
@@ -33,21 +81,123 @@
...
 
 mod ansi;
-mod sys;
+pub(crate) mod sys;
+
+/// Enables raw mode.
+///
+/// Please have a look at the [raw mode](./#raw-mode) section.
+pub fn enable_raw_mode() -> Result<()> {
+    sys::disable_raw_mode()
```
zrzka: `enable_raw_mode` calls `sys::disable_raw_mode`?

src/terminal.rs
```diff
@@ -33,21 +81,123 @@
...
+}
+
+/// Disables raw mode.
+///
+/// Please have a look at the [raw mode](./#raw-mode) section.
+pub fn disable_raw_mode() -> Result<()> {
+    sys::enable_raw_mode()
```
zrzka: `disable_raw_mode` calls `sys::enable_raw_mode`?

src/terminal.rs
```diff
@@ -33,21 +81,123 @@
...
+}
+
+/// Exits the current application.
+pub fn exit() {
+    sys::exit();
+}
+
+/// Returns the terminal size `(columns, rows)`.
+///
+/// The top left cell is represented `1,1`.
```
zrzka: Nitpick - `1,1` -> `(1, 1)`.

src/terminal.rs
TimonPost: lol

src/terminal.rs
TimonPost: lol

PR #340, opened 2019-12-10
Timon/poll wake fix

TimonPost:
- Break loop in `InternalEventReader` when waking thread up with `Waker`

### DIFF
--- a/src/cursor.rs
+++ b/src/cursor.rs
@@ -44,9 +44,9 @@
 
 pub use sys::position;
 
+use crate::{impl_display, utils::Command};
 #[cfg(windows)]
 use crate::utils::Result;
-use crate::{impl_display, utils::Command};
 
 mod ansi;
 pub(crate) mod sys;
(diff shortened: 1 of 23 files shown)


TimonPost:
#[cfg(test)]
mod tests {
    use super::INTERNAL_EVENT_READER;
    use crate::event::filter::EventFilter;
    use crate::event::poll_internal;
    use crate::terminal::{disable_raw_mode, enable_raw_mode};
    use std::thread;
    use std::time::Duration;

    #[test]
    pub fn awake_stream() {
        let waker = INTERNAL_EVENT_READER.read().waker();

        let handle = thread::spawn(|| {
            let event = poll_internal(None, &EventFilter);

            println!("result: {:?}", event);
        });

        enable_raw_mode().unwrap();

        println!("waiting ...");
        thread::sleep(Duration::from_millis(2000));
        println!("waking up ...");
        waker.wake().unwrap();
        println!("joining ...");
        handle.join().unwrap();

        disable_raw_mode().unwrap();
    }
}

PR #363, opened 2020-01-12
Added basic trait implementations to all Commands

Lucretiel:
Added a small set of reasonable trait implementations (`Debug`, `Copy`, `Clone`, `PartialEq`, and `Eq`) to all of the commands. The largest command is a pair of `u16`, so they're easily copyable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 - Fixed minor syntax bug in `execute!` and `queue!`.
 - Cleaned up implementation of `execute!` and `queue!`.
 - **breaking change** Changed `ContentStyle::apply` to take self by value instead of reference, to prevent an unnecessary extra clone.
+- Added basic trait implementations (`Debug`, `Clone`, `Copy`, etc) to all of the command structs
+- `ResetColor` uses `&'static str` instead of `String`
 
 # Version 0.14.2
 - Fix TIOCGWINSZ for FreeBSD
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,7 +47,7 @@ version = "0.3.8"
 features = ["winuser"]
 
 [target.'cfg(windows)'.dependencies]
-crossterm_winapi = "0.5.1"
+crossterm_winapi = "0.6.0"
 
 #
 # UNIX dependencies
--- a/src/cursor.rs
+++ b/src/cursor.rs
@@ -57,6 +57,7 @@ pub(crate) mod sys;
 ///
 /// * Top left cell is represented as `0,0`.
 /// * Commands must be executed/queued for execution otherwise they do nothing.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub struct MoveTo(pub u16, pub u16);
 
 impl Command for MoveTo {
@@ -78,6 +79,7 @@ impl Command for MoveTo {
 /// # Notes
 ///
 /// Commands must be executed/queued for execution otherwise they do nothing.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub struct MoveToNextLine(pub u16);
 
 impl Command for MoveToNextLine {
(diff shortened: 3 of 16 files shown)


TimonPost:
Thanks! I would like to have it merged. Can you add a changelog entry?

```  
 |          ^^^^ could not find `main` in `tokio`

error[E0277]: `main` has invalid return type `impl core::future::future::Future`
##[error]  --> examples/event-stream-tokio.rs:59:20
   |
59 | async fn main() -> Result<()> {
   |                    ^^^^^^^^^^ `main` can only return types that implement `std::process::Termination`
   |
   = help: consider using `()`, or a `Result`
```

I am not sure why this error is happening. I think it has to do with some changes I made yesterday. Just forget about it for now.

Lucretiel:
- Added changelog entry
- Added [Debug, Clone] to a few other structs
- A couple obvious minor code changes

TimonPost:
Great, I wonder whether I should crate just one huge macro that generates all those commands. Because comments, derives and impls are becoming a bit repetitive now.

TimonPost:
Another thing, if we do derive all those traits, why don't we just let de `Command` trait inherit those derivatives.

Lucretiel:
> Another thing, if we do derive all those traits, why don't we just let de `Command` trait inherit those derivatives.

What do you mean? We do do that. Rust prefers to be explicit, so if you don't `#[derive(Debug)]`, there's no debug implementation.

Unless you're asking why we don't do `impl Derive for T: Command`, in which case my answer is that Debug should be implemented directly on structs, because it provides debugging information about their internals that is not available to the `Command` trait.

Lucretiel:
Note: we're getting some failures due to missing debug implementations in crossterm_winapi. I'll file a pull request over there, and once that's resolved we can finish this one.

TimonPost:
Ah indeed. It was a thought, you're correct. 

I  can merge and release winapi pretty fast. So that shouldn't be the issue. Thanks again!

Lucretiel:
Looks like winapi omits Debug implementations (😖) so I'm gonna add some custom impls for the time being. Looks like they're working on it here: https://github.com/retep998/winapi-rs/pull/760

Lucretiel:
I've started work on adding `Debug` implementations to `crossterm_winapi`, here: https://github.com/crossterm-rs/crossterm-winapi/pull/10. This PR is on hold until that one lands.

TimonPost:
if you rebase/merge then the errors should be resolved.

Lucretiel:
Done

TimonPost:
As I look through the code I question whether we should add debug and clone derivitives to non-public internal types. What are the usecases for this? And isn't it overhead?

Lucretiel:
Adding debug and clone to internal structs is (almost) always good practice, because the traits are derived recursively, which means that embedding them in public structs requires them to have the traits. It can also be useful for all the same reasons that clone and debug are useful in the first place.

It does add a small amount of compile time overhead– negligible compared to everything else the compiler needs to do, especially in complex terminal applications. And, if that code is never used, it doesn't get added to the final binary at all, which means there's no runtime size or speed overhead.

That being said, if you'd rather wait and add these traits to internal types "as needed", especially the ones that aren't embedded, I'm happy to do that, too.

Lucretiel:
After considering further, this Pull Request is starting to experience scope creep. While I still think that it's worth adding Debug + Clone to nearly all structs, we can save that for a different PR. I'll revert the additions to the internal types for now, and make this PR just the traits on public types (and the minor code refactorings)

Lucretiel:
Squashed my changes. I left `Debug` on a few internal structs, in cases where:

- It's embedded in a public struct
- It just seemed reasonable, because the type is simple.

TimonPost:
Okay, thanks for elaborating on the motivation, now it makes more sense indeed.

TimonPost:
Once that we merge the winapi PR we can merge this one

TimonPost:
crossterm winapi 0.6.0 is released

Lucretiel:
Will do, thanks!

Lucretiel:
Merge conflicts resolved

Lucretiel:
Need to do some fixes for the Windows changes. I have a windows machine now so I can do that tonight.

TimonPost:
Thanks for the addition and implementations! I will fix the remaining issues and it will be released this evening.

Lucretiel:
Oh, thank you! Sorry, I got busy with other stuff and I lost track of this. The fixes should all be pretty straightforward.

PR #364, opened 2020-01-13
Command trait ref

Lucretiel:
Added an implementation of `Command` for `&T: Command`. This allows commands to be queued by reference, as well as by value:

```
fn queue_many(command: impl Command, times: usize, mut dest: impl Write) -> Result<()> {
    (0..times).try_for_each(move |_| {
        // Can't do dest.queue(command) because we can't move it
        dest.queue(&command)
    })
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# master
+- Added a generic implementation `Command` for `&T: Command`. This allows
+  commands to be queued by reference, as well as by value.
+
 # Version 0.14.2
 - Fix TIOCGWINSZ for FreeBSD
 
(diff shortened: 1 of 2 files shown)


Lucretiel:
For reference, the motivating use-case here is a wrapper like this:

```
{
    // send `EnterAlternateScreen`. Then, when `writer` goes out of scope, send LeaveAlternateScreen
    let writer = push_state(&mut writer, EnterAlternateScreen, LeaveAlternateScreen);
}
```

TimonPost:
if you rebase/merge then the errors should be resolved.

Lucretiel:
Done!

TimonPost:
Great!

Issue #356, opened 2020-01-03
Cargo Warning From `queue!` macro: use of deprecated item 'std::error::Error::description'


zicklag:
**Describe the bug**

When compiling my project I get a warning from cargo wherever the `queue!` macro is used:

    warning: use of deprecated item 'std::error::Error::description': use the Display impl or to_string()

**To Reproduce**

Use the `queue!` macro in a project.

**Expected behavior**

There should be no warning

**OS**
 - Ubuntu 18.04

**Terminal/Shell**

- N/A

TimonPost:
Fixed

PR #369, opened 2020-01-16
Cleaned up command macros and StyledContent

Lucretiel:
- Substantially simplified implementation of command queue macros, including some bug fixes
    - Fixed errors being usually ignored
- Cleaned up StyledContent and ContentStyle
    - Removed unneeded trait and lifetime bounds
    - made builder pattern methods more expression oriented, and added inlines
    - added StyledContent::style_mut

This PR includes a few breaking changes, only one of which actually matters:

- Changed ContentStyle::with to take `self` instead of `&self`. This is because the former version clones `self`, which includes `Vec` clone; these are generally considered bad practice when done secretly. I'm happy to revert this change if you still prefer the old version
- changes to `execute!` and `queue!`:
    - Added explicit support for queue 0: `queue!(stdout);` is now valid.
    - you can no longer double up on a trailing comma: `queue!(stdout,,);` is no longer valid.
    - they no longer ignore errors for the first N-1 queued events.
    - execute now only flushes once, after all events are queued, and returns an error instead of panicking on a flush error.

### DIFF
--- a/src/style/styled_content.rs
+++ b/src/style/styled_content.rs
@@ -28,51 +28,70 @@ use crate::{
 /// println!("{}", styled);
 /// ```
 #[derive(Clone)]
-pub struct StyledContent<D: Display + Clone> {
+pub struct StyledContent<D: Display> {
     /// The style (colors, content attributes).
     style: ContentStyle,
     /// A content to apply the style on.
     content: D,
 }
 
-impl<'a, D: Display + 'a + Clone> StyledContent<D> {
+impl<D: Display> StyledContent<D> {
     /// Creates a new `StyledContent`.
+    #[inline]
     pub fn new(style: ContentStyle, content: D) -> StyledContent<D> {
         StyledContent { style, content }
     }
 
     /// Sets the foreground color.
-    pub fn with(mut self, foreground_color: Color) -> StyledContent<D> {
-        self.style = self.style.foreground(foreground_color);
-        self
+    #[inline]
+    pub fn with(self, foreground_color: Color) -> StyledContent<D> {
+        Self {
+            style: self.style.foreground(foreground_color),
+            ..self
+        }
     }
 
     /// Sets the background color.
-    pub fn on(mut self, background_color: Color) -> StyledContent<D> {
-        self.style = self.style.background(background_color);
-        self
+    #[inline]
+    pub fn on(self, background_color: Color) -> StyledContent<D> {
+        Self {
+            style: self.style.background(background_color),
+            ..self
+        }
     }
 
     /// Adds the attribute.
     ///
     /// You can add more attributes by calling this method multiple times.
-    pub fn attribute(mut self, attr: Attribute) -> StyledContent<D> {
-        self.style = self.style.attribute(attr);
-        self
+    #[inline]
+    pub fn attribute(self, attr: Attribute) -> StyledContent<D> {
+        Self {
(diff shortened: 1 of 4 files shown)


Lucretiel:
Working on adding tests and a changelog entry now

Lucretiel:
Also, I'd like to propose changing `ContentStyle::attributes` from a `Vec` to a `HashSet`, but that's a much more substantial API change so I wanted to ask here first

src/style/styled_content.rs
```diff
@@ -85,13 +104,16 @@ impl<D: Display + Clone> Display for StyledContent<D> {
...
+        for attr in &self.style.attributes {
             queue!(f, SetAttribute(*attr)).map_err(|_| fmt::Error)?;
             reset = true;
         }
 
-        fmt::Display::fmt(&self.content, f)?;
+        self.content.fmt(f)?;
 
+        // TODO: There are specific command sequences for "reset forground
+        // color (39m)" and "reset background color (49m)"; consider using
```
TimonPost: Actually I think we should not reset to the default for or background. Instead, we probably want to undo the attributes and colors of this `StyledContent`.

Lucretiel: I agree. I looked into doing something like this many years ago. Unfortunately, because the attributes aren't stackful (that is, you can't "pop" a background color and restore the previous one), the only way to do this correctly would be to query the style before initiating the writes, so that we can restore it afterwards.

A *mostly* correct way to do this would to require our `QueuableCommand` instances to be wrapped in a struct, which can then manually keep track of the "style stack". This would be correct so long as is only one instance of this struct, and that the end user doesn't insert their own control sequences into the stream.

Lucretiel:
> Great! I love it that we do not need the clone anymore.
> 
> Does this resolve #356 as well?

It does! I was pleasantly surprised to discover that this was a *very* easy fix.

src/style/styled_content.rs
Lucretiel: However, solving that problem requires more design work, and *any* solution we come up with will be backwards incompatible, so I figured I'd leave the comment there for now, rather than creating a fix that we'd just have to change again down the line.

Lucretiel:
Fixed (well, removed) failing Command::winapi tests.

src/style/styled_content.rs
TimonPost: There is an easier solution, like we have in the examples. We store both the enable and disable attributes in a tuple style array.

Attrs = [(BoldOff, BoldOn)]

We can query in that array for the attribute and take the opposite. This is similar to what cursive does. I do this in the terminal library as well.

Lucretiel: I'm curious to see, which example file does it?

TimonPost: https://github.com/crossterm-rs/crossterm/blob/master/examples/interactive-demo/src/test/attribute.rs#L7

Lucretiel: Oh, that's not the part I'm worried about. The part I'm worried about is tracking WHICH state to restore after you finish writing a styled content. You can't just unconditionally invert.

For example, consider this pseudocode:

```
WriteUnderlined("abc", WriteUnderlined("1234"), "xyz")
```

If the inner `WriteUnderlined` calls `UnsetUnderline` after printing 1234, then "xyz" *won't* be underlined, even though it should be
```

PR #372, opened 2020-01-23
Fix typo in README.md

thunderrabbit:
**Its**, without an apostrophe, is the possessive of the pronoun it. **It’s**, with an apostrophe, is a contraction of it is or it has.

https://grammarist.com/spelling/its-its/


TimonPost:
Thanks!

Issue #371, opened 2020-01-23
Control + J read as enter.


TimonPost:
**Describe the bug**
CTRL + J is read as enter. 

https://github.com/Canop/broot/issues/177

**Expected behavior**
I should receive CTRL + J when pressing enter

**OS**
 - e.g. MacOs, Linux

sharnoff:
I encountered this issue as well - I'd love to work on it, if possible. Where should I look?

TimonPost:
You should look at the event parsing code. This is in the event/sys/unix folder. I will put some links in here in a bit.

TimonPost:
Fixed

PR #373, opened 2020-01-25
Fixes issue #371 (As much as is possible)

sharnoff:
The key codes for newline ('\n') and Ctrl+J are both 0xA - this is given by the terminal, and there's not much we can do about that particular fact. As it turns out, terminals operating in "canonical mode" - i.e. not raw mode - will convert any carriage returns ('\r') into newline characters - *even though the user doesn't directly input newlines*.

When we enter raw mode, one of the things that gets disabled is this automatic conversion. Now, when we get newlines, there's no ambiguity; it must be Ctrl+J. Terminal apps (like vim) are then able to handle Ctrl+J separately because they make that distinction once in raw mode.

More info: [https://viewsourcecode.org/snaptoken/kilo/02.enteringRawMode.html#fix-ctrl-m](https://viewsourcecode.org/snaptoken/kilo/02.enteringRawMode.html#fix-ctrl-m); ICRNL bitflag in [termios](http://man7.org/linux/man-pages/man3/termios.3.html)

This commit (roughly speaking) changes changes event parsing of newlines to only be enabled while the terminal isn't in raw mode. Outside of raw mode, it doesn't seem possible to handle Enter and Ctrl+J separately without using some of the same features of `termios.h` that are used to enable raw mode.

### DIFF
--- a/src/event/sys/unix/parse.rs
+++ b/src/event/sys/unix/parse.rs
@@ -66,7 +66,14 @@ pub(crate) fn parse_event(buffer: &[u8], input_available: bool) -> Result<Option
                 }
             }
         }
-        b'\r' | b'\n' => Ok(Some(InternalEvent::Event(Event::Key(
+        b'\r' => Ok(Some(InternalEvent::Event(Event::Key(
+            KeyCode::Enter.into(),
+        )))),
+        // Issue #371: \n = 0xA, which is also the keycode for Ctrl+J. The only reason we get
+        // newlines as input is because the terminal converts \r into \n for us. When we
+        // enter raw mode, we disable that, so \n no longer has any meaning - it's better to
+        // use Ctrl+J. Waiting to handle it here means it gets picked up later
+        b'\n' if !crate::terminal::sys::is_raw_mode_enabled() => Ok(Some(InternalEvent::Event(Event::Key(
             KeyCode::Enter.into(),
         )))),
         b'\t' => Ok(Some(InternalEvent::Event(Event::Key(KeyCode::Tab.into())))),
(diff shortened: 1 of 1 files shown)


src/event/sys/unix/parse.rs
```diff
@@ -66,9 +66,16 @@ pub(crate) fn parse_event(buffer: &[u8], input_available: bool) -> Result<Option
...
         }
-        b'\r' | b'\n' => Ok(Some(InternalEvent::Event(Event::Key(
+        b'\r' => Ok(Some(InternalEvent::Event(Event::Key(
             KeyCode::Enter.into(),
         )))),
+        // Issue #371: \n = 0xA, which is also the keycode for Ctrl+J. The only reason we get
+        // newlines as input is because the terminal converts \r into \n for us. When we
+        // enter raw mode, we disable that, so \n no longer has any meaning - it's better to
+        // use Ctrl+J. Waiting to handle it here means it gets picked up later
+        b'\n' if !crate::terminal::sys::is_raw_mode_enabled() => Ok(Some(InternalEvent::Event(
```
TimonPost: Though the input layer isn't very useful if the raw mode is not enabled.

sharnoff: I'm not sure I get what you're saying - does `parse_event` only run while in raw mode?

sharnoff:
> I will have to run this branch. The input layer is not very useful when raw mode is not enabled therefore wouldn't it be pointless to check for this?

I don't know when this is used, but I built a version of a personal project with this change, and it fixed the issue (I'd previously noticed it there). This issue *might* also extend to Windows, depending on how the terminal is configured, but I haven't looked into that

TimonPost:
The windows terminal does not use those codes. It is UNIX only that parses those keys. Secondly, raw mode will make input bytes be written to the terminal buffer directly. And will prevent the bytes from being displayed by the user and allow the terminal to handle the bytes. This is what we do, without raw mode we are not able to fetch the input bytes.

TimonPost:
It can be realistic that CTRL + J is not possible to handle. There are limitations to some keys we can support. I will give it a try in the evening, and check with other libraries if they support this. So I will report back in some hours.

TimonPost:
image

CONTROL + J is working for me on Linux. It seems to work!

Issue #376, opened 2020-01-29
Make ContentStyle lighter with a bitmask


Canop:
The ContentStyle is cloned when you apply it. Depending on the application, there might be a lot of those clonings, and of content_styles in memory.

The struct currently contains a vector of attributes. Attribute is an enum with 24 values. Using a bitmask instead of a vector would result in a lighter structure and in faster add/remove/read operations.

The API would be changed a little (as the vector is public) but the change is probably easy to manage.

If nobody objects, I can look into proposing a PR for this.

Canop:
Replacing the `Attribute` enum with an `Attributes` bit set seems to make a lot of sense but the work is a tad intimidating after a first review:

- exposing `Attributes` and the constants in a `crossterm::style::attributes` module
- replacing the `SetAttribute` command with a `SetAttribute`s one, which functionally isn't exactly the same as it can be used to change several attributes at once (maybe "keep" the old one as an alias to `SetAttributes`)
- implementing writing the csi/sgr sequence from the Attributes
- implementing the Windows writing
- rewrite all existing style tests
- fix the documentation

I can't make a PR for all this alone. A shared idea of the goal is needed and somebody would have to do the Windows part and I'd like to have some help for the unit tests too.

TimonPost:
I remembered pancurses to have a similar implementation.  See [here](https://github.com/ihalila/pancurses/blob/master/src/attributes.rs).

They have an `Attribute**s**` struct which wraps a raw value (byte) and uses trait implementations for mapping the enum `Attribute` to a numeric value. This way we can also still keep the `Attribute` enum. Which would cause no breaking changes? As you said we can create a new command `SetAttributes` change `ContentStyle` to have the new `Attribute**s**` type and use bit or to add attributes to that type. Please check the link in order to get a better understanding.

Canop:
The pancurses implementation seems OK.
There are still breaking changes as the Vec<Attribute> is public but it can indeed be done in a less traumatic way and eases dynamic styling.

TimonPost:
Completed

PR #380, opened 2020-02-06
Attributes

Canop:
Replace the Vec<Attribute> with a bitset as proposed in #376

### DIFF
--- a/src/style/ansi.rs
+++ b/src/style/ansi.rs
@@ -3,7 +3,7 @@
 
 use crate::{
     csi,
-    style::{Attribute, Color, Colored},
+    style::{Attribute, Attributes, Color, Colored},
 };
 
 pub(crate) fn set_fg_csi_sequence(fg_color: Color) -> String {
@@ -21,7 +21,17 @@ pub(crate) fn set_bg_csi_sequence(bg_color: Color) -> String {
 }
 
 pub(crate) fn set_attr_csi_sequence(attribute: Attribute) -> String {
-    format!(csi!("{}m"), attribute as i16)
+    format!(csi!("{}m"), attribute as u64)
+}
+
+pub(crate) fn set_attrs_csi_sequence(attributes: Attributes) -> String {
+    let mut s = String::new();
+    for attr in Attribute::iterator() {
+        if attributes.has(attr) {
+            s.push_str(&format!("\x1B[{}m", attr as u64));
+        }
+    }
+    s
 }
 
 pub(crate) const RESET_CSI_SEQUENCE: &str = csi!("0m");
--- a/src/style/ansi.rs
+++ b/src/style/ansi.rs
@@ -21,14 +21,14 @@ pub(crate) fn set_bg_csi_sequence(bg_color: Color) -> String {
 }
 
 pub(crate) fn set_attr_csi_sequence(attribute: Attribute) -> String {
-    format!(csi!("{}m"), attribute as u64)
+    format!(csi!("{}m"), attribute.sgr())
 }
 
 pub(crate) fn set_attrs_csi_sequence(attributes: Attributes) -> String {
     let mut ansi = String::new();
     for attr in Attribute::iterator() {
         if attributes.has(attr) {
-            ansi.push_str(&format!(csi!("{}m"), attr as u64));
+            ansi.push_str(&format!(csi!("{}m"), attr.sgr()));
         }
     }
     ansi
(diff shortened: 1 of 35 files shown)


src/style/attributes.rs
```diff
@@ -0,0 +1,108 @@
...
+    }
+}
+
+impl Attributes {
+    #[inline(always)]
+    pub fn push(&mut self, attribute: Attribute) {
+        self.0 |= attribute.bytes();
+    }
+    #[inline(always)]
+    pub fn remove(&mut self, attribute: Attribute) {
```
TimonPost: What about `unset` in the sense of undoing the `set` operation, also to be consistent with `set` name.

src/style/attributes.rs
```diff
@@ -0,0 +1,108 @@
...
+    }
+    #[inline(always)]
+    pub fn has(&self, attribute: Attribute) -> bool {
+        self.0 & attribute.bytes() != 0
+    }
+    #[inline(always)]
+    pub fn extend(&mut self, attributes: &Attributes) {
+        self.0 = self.0 | attributes.0;
+    }
+    /// tells whether there's absolutely no attribute
```
TimonPost:
```suggestion
    /// Returns whether there is no attribute set.
```

TimonPost: In crossterm, I have plural verb + operation description.

src/style/ansi.rs
```diff
@@ -21,7 +21,17 @@ pub(crate) fn set_bg_csi_sequence(bg_color: Color) -> String {
...
 pub(crate) fn set_attr_csi_sequence(attribute: Attribute) -> String {
-    format!(csi!("{}m"), attribute as i16)
+    format!(csi!("{}m"), attribute as u64)
+}
+
+pub(crate) fn set_attrs_csi_sequence(attributes: Attributes) -> String {
+    let mut s = String::new();
+    for attr in Attribute::iterator() {
+        if attributes.has(attr) {
+            s.push_str(&format!("\x1B[{}m", attr as u64));
```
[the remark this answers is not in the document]
Canop: Unless I'm mistaken, this means another call and another string allocation. I don't think we should do it.

src/style/attributes.rs
Canop: I'm OK with that

src/style/content_style.rs
```diff
@@ -51,7 +51,7 @@ impl ContentStyle {
     /// You can add more attributes by calling this method multiple times.
     #[inline]
     pub fn attribute(mut self, attr: Attribute) -> ContentStyle {
```
TimonPost: Would it be an idea, to create a method that accepts `Attributes` as well?

src/style/attributes.rs
```diff
@@ -0,0 +1,111 @@
+use crate::style::Attribute;
+use std::ops::{BitAnd, BitOr, BitXor};
+
+/// a bitset for all possible attributes
+#[derive(Debug, Default, Clone, Copy, PartialEq, Eq)]
+pub struct Attributes(u64);
```
TimonPost: Do we really need a u64 for this? We can go way smaller right? Even something like a u8 would already be enough. Or am I mistaken?

src/style/ansi.rs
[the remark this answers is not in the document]
TimonPost: It uses `concat!(`. Which means it will be concatenated at runtime right?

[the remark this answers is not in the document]
Canop: You might be right (assuming you mean at compile-time).

src/style/attributes.rs
Canop: There are 24 values. So a u8 can't store them all.
A u32 would be enough but then we need another implementation for `Attribute.bytes()`.
A possibility would be to remove the SGR indexes from the enum and hardcode them in the unix implementation of SetAttribute and SetAttributes. Which would probably not be dirtier.

src/style/content_style.rs
Canop: probably. Calling extend.

src/style/ansi.rs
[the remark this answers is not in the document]
TimonPost: Yes that is what I meant

src/style/attributes.rs
```diff
@@ -61,26 +61,34 @@ impl BitXor for Attributes {
 }
 
 impl Attributes {
+    /// Sets the attribute.
+    /// If it's already set this does nothing.
```
TimonPost:
```suggestion
    /// If it's already set, this does nothing.
```

src/style/content_style.rs
TimonPost: Oke

TimonPost:
There are some clippy errrors. If those are resolved, this PR can be merged. Thanks again!

TimonPost:
Awesome!

PR #386, opened 2020-02-25
Added Support For Chars Colors

PeterHindes:
Is this good? I saw some functions that also clear the screen and what not, should those be implemented for chars too?


PeterHindes:
Now we can bold and other styles on a char

PeterHindes:
Done

TimonPost:
Thanks again!

PeterHindes:
I don't know if you want it, but I added an example to my forked repository just now. Merge it if you want.

PR #390, opened 2020-02-29
Direct write command ansi_codes

Riey:
This change many commands from `format! -> String -> write!` to single `write!`

### DIFF
--- a/examples/interactive-demo/src/test/cursor.rs
+++ b/examples/interactive-demo/src/test/cursor.rs
@@ -132,7 +132,7 @@ fn draw_cursor_box<W, F, T>(w: &mut W, description: &str, cursor_command: F) -> 
 where
     W: Write,
     F: Fn(u16, u16) -> T,
-    T: Command<AnsiType = String>,
+    T: Command,
 {
     execute!(
         w,
--- a/src/ansi.rs
+++ b/src/ansi.rs
@@ -1,0 +1,4 @@
+/// Wrapper type for write dynamic ansi string
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[doc(hidden)]
+pub struct Ansi<T>(pub T);
--- a/src/cursor.rs
+++ b/src/cursor.rs
@@ -46,7 +46,7 @@ pub use sys::position;
 
 #[cfg(windows)]
 use crate::Result;
-use crate::{impl_display, Command};
+use crate::{impl_display, Ansi, Command};
 use std::fmt;
 
 mod ansi;
@@ -61,18 +61,18 @@ pub(crate) mod sys;
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub struct MoveTo(pub u16, pub u16);
 
-impl fmt::Display for MoveTo {
+impl fmt::Display for Ansi<MoveTo> {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        ansi::move_to_csi_sequence(f, self.0, self.1)
+        ansi::move_to_csi_sequence(f, (self.0).0, (self.0).1)
     }
 }
 
 impl Command for MoveTo {
-    type AnsiType = Self;
+    type AnsiType = Ansi<Self>;
 
     #[inline]
     fn ansi_code(&self) -> Self::AnsiType {
-        *self
+        Ansi(*self)
     }
 
     #[cfg(windows)]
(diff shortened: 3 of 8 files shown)


Riey:
I just realized impl_display also can handle winapi it seems need patch

Riey:
However I want get feedback with this idea if you don't like it I just use my own Command for my case

TimonPost:
Hi thanks for sharing your ideas, I like it. 

So to make sure we are on the same page, we have two ANSI strings:
- static
- dynamic

- The static will stay how they are and use the `impl_display` macro
- The dynamic will implement the display manually.

Right?

Riey:
I think there is three type

1. static and have different AnsiType (like &'static str)
2. static and have Self AnsiType (It doesn't exists before but I just made it)
3. dynamic which can has dynamic content (like `PrintStyledContent<D>`)

and second's Display implementation can't handle winapi that's my mistake

Riey:
Yes and I try to using static in some command

Riey:
What you mean static?

`PrintStyledContent<D>` use `impl_display` so is this static?

TimonPost:
Soms ansi code string are static and do not change based on the user wishes. Reset color is one example of that. An example of dynamic ansi code is cursor movement. This string changes based on user input and needs to be formatted. 

Based on that we can keep the `impl_display` macro for static ansi codes and like you have already done use a formatter for dynamic ansi codes that require `format!` with user input.

Riey:
Ok I understand so I want to make write dynamic ansi string without `format!`

but ansi/winapi branch is done by impl_display macro so I think we need new type that execute ansi function in thier Display impl (like PrintStyledContent and StyledContent)

Riey:
How about make `Ansi<T>` and implement Display for `Ansi<MoveTo>`, `Ansi<MoveLeftTo>`... ?

TimonPost:
Why would we need such a type? The executing is done in the execute/queue macro.

Riey:
Because `Command` doesn't write its content immediately it return `Command::AnsiType` and AnsiType can't be Self(like this pr)

and type erased buffer type is not we want(like String before)

so we need new `Display` type for `Command::AnsiType` but not `Command` and call ansi functions directly

TimonPost:
Ah, gotcha, let's do that, we create a wrapper `Ansi<T>` type which wraps `T` that is displayable. This type will be returned from `Command::ansi_code`. I like that name as well, an alternative might be `DynamicAnsi`  though `Ansi` reads easier.

Riey:
So we have two option for make `Ansi<T>`

## 1. Generic way
```rust

pub struct Ansi<T>(pub T);

impl<T> Display for Ansi<T> where T: AnsiDisplay {
    fn fmt(&self, f: &mut Formatter) -> fmt::Result {
        self.0.ansi_fmt(f)
    }
}

pub trait AnsiDisplay {
    fn ansi_fmt(&self, f: &mut Formatter) -> fmt::Result;
}

impl AnsiDisplay for MoveTo {
    fn ansi_fmt(&self, f: &mut Formatter) -> fmt::Result {
        ansi::move_to_csi_sequence(f, self.0, self.1)
    }
}


// with this way user also can use Ansi<T> for their types

pub struct UserType;

impl AnsiDisplay for UserType { /* code */ }

// now Ansi<UserType> impl Display
// it doesn't meaningful now but we can get chance to change Command like this

pub trait Command: AnsiDisplay + Sized {
    fn ansi_code(self) -> Ansi<Self> {
        Ansi(self)
    }

    #[cfg(windows)]
    fn execute_winapi(&self) -> Result<()>;
}

```

## 2. Concrete way
```rust

pub struct Ansi<T>(pub T);

// don't need new trait

impl Display for Ansi<MoveTo> {
    fn fmt(&self, f: &mut Formatter) -> fmt::Result {
        ansi::move_to_csi_sequence(f, self.0.0, self.0.1)
    }
}


// with this way user can't use Ansi<T> because of orphan rule

pub struct UserType;

// impossible
impl Display for Ansi<UserType> { /* code */ }

// but it doesn't meaningful now

```

If you don't mind of change `Command` I prefer second and set `Ansi<T>` `#[doc(hidden)]`

TimonPost:
I prefer the second as well. Because it introduces the least breaking changes and we always will have the option - if it turns out to be better - to switch to scenario 1. Secondly, the user already can create their own implementation of `Command`.

Riey:
In fact, static ansi string like `ResetColor`  also have performance benefit
but it's very little gain so I just work for remove `format!`

```rust
impl Command for ResetColor {
    type AnsiType = &'static str;

    fn ansi_code(&self) -> Self::AnsiType {
        // copy &str and then write_str
        ansi::RESET_CSI_SEQUENCE
    }

    #[cfg(windows)]
    fn execute_winapi(&self) -> Result<()> {
        sys::windows::reset()
    }
}

impl Display for Ansi<ResetColor> {
    fn fmt(&self, f: &mut Formatter) -> fmt::Result {
        f.write_str(ansi::RESET_CSI_SEQUENCE)
    }
}

impl Command for ResetColor {
    type AnsiType = Ansi<Self>;

    fn ansi_code(&self) -> Self::AnsiType {
        // Ansi<Self> is ZST so it doesn't copy and directly write by write_str
        Ansi(Self)
    }

    #[cfg(windows)]
    fn execute_winapi(&self) -> Result<()> {
        sys::windows::reset()
    }
}
```

Riey:
It break two public apis

1. String no longer implement  `From<Colored>` it could be replace with `to_string`
2. These command's AnsiType is no longer `String`

TimonPost:
Awesome!

PR #392, opened 2020-02-29
add NONE flag to KeyModifiers

ronthecookie:
This change allows for:
```rust
        match event::read()? {
            Event::Key(k) => match k {
                KeyEvent {
                    code: KeyCode::Char(ch),
                    modifiers: KeyModifiers::NONE,
                } => println!("char: {}", ch),
                _ => {}
            },
            _ => {}
        }
```
instead of:
```rust
match k {
    KeyEvent {
        code: KeyCode::Char(ch),
        modifiers,
    } if modifiers == KeyModifiers::empty() => {}
}
```

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -337,6 +337,7 @@ bitflags! {
         const SHIFT = 0b0000_0001;
         const CONTROL = 0b0000_0010;
         const ALT = 0b0000_0100;
+        const NONE = 0b0000_0000;
     }
 }
 


TimonPost:
Thanks for paying interest in crossterm, to validate, is it true that a 0 value will be set by default, so, in this case, NONE is always set?

ronthecookie:
Yep, seems so - here's a quick [example](https://github.com/ronthecookie/crossterm-none) I made

TimonPost:
Thanks!

PR #407, opened 2020-03-31
Changed STDIN to /dev/tty

TimonPost:


japert:
I tried this branch on macOS 10.15.4.
Getting:
```
$ cargo run --features="event-stream" --example event-stream-tokio
thread 'main' panicked at 'Event source can not be initialized: IoError(Os { code: 22, kind: InvalidInput, message: "Invalid argument" })', src/event/read.rs:30:17
               note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Which is the sam IoError I got on macOS on cross term 0.17

japert:
After the merge, I am still getting the same IOError.

extrawurst:
@japert yup I can confirm it’s still throwing the error on macOs

TimonPost:
Its unfortunate I cant test on mac os. It is right that in 0.17 there was a bug with the TTY. I reverted that bug with https://github.com/crossterm-rs/crossterm/pull/405/files, however this Pr introduces the same problem again. For some reason, on mac os, we cant use the TTY for input reading. It throws an error if we are opening the file.

acidghost:
This because of tokio-rs/mio#1377.

TimonPost:
Amazing! That could be the issue. Thanks for linking.

cosmikwolf:
for anyone trying to get stdin working with crossterm, but having a hard time, I made an example repo on how to do it:
https://github.com/cosmikwolf/piped_cli_interactive

Issue #406, opened 2020-03-30
The size() is broken on MacOs.


extrawurst:
**Describe the bug**
commit https://github.com/crossterm-rs/crossterm/commit/52b9d479c16fed8208449cb0ab9c59823a858f0c broke querying the **correct** size of the terminal on MacOs aswell as the resize event. The size returned will always be 80x24 and not the *actual* size.
I am pretty sure this is related to the revert that was necessary to fix the input on Mac aswell in `0.17.1` because it also tries to read "/dev/tty".

see the gif as an illustration (note the resize event always returning 80x24 instead of the correct size values):
![demo](https://user-images.githubusercontent.com/776816/77950736-3caa2d00-72c9-11ea-9908-835276bdca49.gif)

**note:** reverting that particular commit fixes the issue

**To Reproduce**
run the example on Mac and resize the window:
```
cargo run --example event-poll-read
```

**OS**
MacOs

**Terminal/Shell**
Zsh/iTerm

*update*: clarified that by *broken* I mean the values for size are not correct.

TimonPost:
I think the problem with this issue is not TTY in specific. The reading terminal size was implemented in version 0.15.1. You said in discord that 0.16 did not have this issue, that would mean that the city is not the problem here.

I think it is a combo of falling back on stdout and using tput/libc. Here is a quick update, I have to go now, will come back at this tomorrow.

First:

```
 if let Ok(true) =
        wrap_with_result(unsafe { ioctl(STDOUT_FILENO, TIOCGWINSZ.into(), &mut size) })
    {
        Ok((size.ws_col, size.ws_row))
    } else {
        tput_size().ok_or_else(|| std::io::Error::last_os_error().into())
    }
```

**version 1 0.15:** https://github.com/crossterm-rs/crossterm/blob/e35d4d2c1cc4c919e36d242e014af75f6127ab50/src/terminal/sys/unix.rs#L35

```rust
let file = File::open("/dev/tty").unwrap();

    if let Ok(true) =
        wrap_with_result(unsafe { ioctl(file.into_raw_fd(), TIOCGWINSZ.into(), &mut size) })
    {
        Ok((size.ws_col, size.ws_row))
    } else {
        tput_size().ok_or_else(|| std::io::Error::last_os_error().into())
    }
```


**version 2:** https://github.com/crossterm-rs/crossterm/blob/e35d4d2c1cc4c919e36d242e014af75f6127ab50/src/terminal/sys/unix.rs#L35

[...]

PR #413, opened 2020-04-08
Issue 406: Keep FileDesc alive long enough for ioctl to work

m-lima:
This PR is to tackle [issue 406](https://github.com/crossterm-rs/crossterm/issues/406)
`FileDesc` was being dropped before reaching the `ioctl` call

Now `FileDesc` is in a `Option` that may live long enough.
The fallback still exists, and now relies on the presence of a file descriptor

### DIFF
--- a/src/terminal/sys/unix.rs
+++ b/src/terminal/sys/unix.rs
@@ -32,13 +32,8 @@ pub(crate) fn size() -> Result<(u16, u16)> {
         ws_ypixel: 0,
     };
 
-    let file = if let Ok(file) = File::open("/dev/tty") {
-        Some(FileDesc::new(file.into_raw_fd(), true))
-    } else {
-        None
-    };
-
-    let fd = if let Some(file) = file {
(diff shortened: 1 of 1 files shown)


extrawurst:
@m-lima Why did u close again?

m-lima:
@extrawurst I needed to do some testing for the `crossterm::event::Event::Resize` event.
Was not getting the desired outcome. So I closed it to avoid people reviewing a possibly wrong PR.

Turns out that when I created a "cleaner" version than the one I posted in the issue comment, I had a typo on the file name and was falling back to `STDOUT_FILENO`. Once I fixed that, I started observing the original problem again.  
It stopped working because my previous `if let Some(file)` was moving the file out of the `Option` (and subsequently dropping it). I mistakenly re-added the original problem I was trying to solve.

**Conclusion:**  
I simplified the file descriptor creation with a `map` and now use a `if let Ok(file)` on a reference, so that it doesn't drop.

PR #415, opened 2020-04-10
Clarify the documentation sources

PurpleBooth:
I was missed these when I was skimming through the repository the first time, I've added these links to make it a tiny bit more obvious.


TimonPost:
Awesome!

PR #416, opened 2020-04-10
Added a note about the examples repo being deprecated

PurpleBooth:
This adds the deprecated in front of the examples directory so it's clear that you're supposed to look in `examples/`


TimonPost:
What do you mean with them being deprecated? They are examples demonstrating this library.

TimonPost:
Nice, thanks for the tip

PR #418, opened 2020-04-17
Add Kitty to supported terminals

kesslern:
Kitty doesn't support ZWJ characters, but I've been playing with this library for awhile now with Kitty and it works great.


TimonPost:
Nice! What do you mean by not supporting?

PR #420, opened 2020-04-20
Add macros for `Colorize` and `Styler` impls, add an impl for `String`

sharnoff:
Per the commit message:
> This shifts the bulk of the repeated code into style/macros.rs, and
switches to using it for all of the implementations of the two traits.
>
> It also adds new implementations of `Colorize` and `Styler` for
`String`.

I found myself wanting an implementation of `Colorize` for `String`, so I figured I'd collapse the existing macros to make that implementation easier. There are two sets of new macros to look at - one for each trait. I'll just describe the one for `Colorize`, but it's nearly identical for `Styler`.

There are three new macros: `def_color_generic`, `impl_colorize`, and `impl_colorize_callback`. The first of these just collects the common elements from `def_str_color` and `def_char_color` as the only difference was the return type of the produced method:

```rust
// First, the two previous macros:

macro_rules! def_str_color {
    ($side:ident: $name:ident => $color:path) => {
        fn $name(self) -> StyledContent<&'static str> {
            StyledContent::new(
                ContentStyle {
                    $side: Some($color),
                    ..Default::default()
                },
                self,
            )
        }
    };
}

macro_rules! def_char_color {
    ($side:ident: $name:ident => $color:path) => {
        fn $name(self) -> StyledContent<char> {
            StyledContent::new(
                ContentStyle {
                    $side: Some($color),
                    ..Default::default()
                },
                self,
            )
        }
    };
}

// And now the new one:

macro_rules! def_color_generic {
    ($color_ty:ty, $side:ident: $name:ident => $color:path) => {
        fn $name(self) -> StyledContent<$color_ty> {
            StyledContent::new(
                ContentStyle {
                    $side: Some($color),
                    ..Default::default()
                },
                self,
            )
        }
    };
}
```

There was a similar process for collecting the actual `impl` blocks themselves. The blocks for the three basic types (`char`, `&'static str`, and `String`) are all created via `impl_colorize`. The more complicated implementation for `StyledContent<D>` is done by the `impl_colorize_callback` macro, which takes one of `def_color_generic` or `def_color` as input to produce the value.

I'm not certain what the best names are for these macros; the current ones might be counter-intuitive, but I actually don't know. I just picked something.

I would also be happy to add some comments describing the macros if that's something that would be valuable.

### DIFF
--- a/src/style/macros.rs
+++ b/src/style/macros.rs
@@ -1,14 +1,45 @@
-macro_rules! def_attr {
-    ((), $name:ident => $attr:path) => {
-        fn $name(self) -> StyledContent<D> {
-            self.attribute($attr)
-        }
-    };
-}
+//! Defines the macros for repetitive styling implementations
 
-macro_rules! def_attr_generic {
-    ($attr_ty:ty, $name:ident => $attr:path) => {
-        fn $name(self) -> StyledContent<$attr_ty> {
+// There's a single core set of macros structure here that's essentially repeated twice; once for
+// implementing `Styler` and once for `Colorize`. We'll go through `Styler` as the example, knowing
+// that `Colorize` works in precisely the same manner.
+//
+// There are four macros in each group. For `Styler`, they are:
+//  * def_attr_base,
+//  * def_attr_generic,
+//  * impl_styler_callback, and
+//  * impl_styler
+//
+// Fundamentally, any implementation works in a similar fashion; many methods with near-identical
+// bodies are grouped together. There are additionally two types of implementors: so-called "base"
+// implementors (`char`, `String`, etc.) and a single "generic" implementor - 'StyledContent<D>'.
+//
+// We can visualize the macro expansion with a sort of pipeline:
+//
+//                                    /--------> def_attr_base
+//   [impl_styler ->] impl_styler_callback
+//                                    \--------> def_attr_generic
+//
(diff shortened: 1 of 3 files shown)


TimonPost:
I really love those changes, yesterday I worked on some tui and discovered that I could not use String. And now you've created the PR. I am a bit busy with other things now and glimpsed at the changes. It looked good so far. However, will do an in-depth review later one. Thanks!

TimonPost:
Awesome, I like to merge this and release it with some other changes. Commenting the macro's would be very helpful. So lets add that and then merge it. Thanks!

sharnoff:
Cool - I'll write some up :+1:

src/style/macros.rs
```diff
@@ -1,14 +1,45 @@
...
-    ($attr_ty:ty, $name:ident => $attr:path) => {
-        fn $name(self) -> StyledContent<$attr_ty> {
+// There's a single core set of macros structure here that's essentially repeated twice; once for
+// implementing `Styler` and once for `Colorize`. We'll go through `Styler` as the example, knowing
+// that `Colorize` works in precisely the same manner.
+//
+// There are four macros in each group. For `Styler`, they are:
+//  * def_attr_base,
+//  * def_attr_generic,
+//  * impl_styler_callback, and
```
TimonPost, line 10: just leave and away
```suggestion
//  * impl_styler_callback, 
```

sharnoff:
Great, thanks! :smiley:

PR #425, opened 2020-05-02
Add raw mode guard

Koxiaet:
This PR adds `crossterm::terminal::RawModeGuard`, a type which enters raw mode when created and exits raw mode when destroyed.

The motivation for this is in applications that could panic - currently, if you run `enter_raw_mode()` and then panic the terminal gets all messed up if you try to use it again because it's still in raw mode. Having this type would prevent that, as its destructor would still be run even on panic.

### DIFF
--- a/src/style/macros.rs
+++ b/src/style/macros.rs
@@ -7,7 +7,7 @@
 // There are four macros in each group. For `Styler`, they are:
 //  * def_attr_base,
 //  * def_attr_generic,
-//  * impl_styler_callback, 
+//  * impl_styler_callback,
 //  * impl_styler
 //
 // Fundamentally, any implementation works in a similar fashion; many methods with near-identical
--- a/src/terminal.rs
+++ b/src/terminal.rs
@@ -107,6 +107,40 @@ pub fn disable_raw_mode() -> Result<()> {
     sys::disable_raw_mode()
 }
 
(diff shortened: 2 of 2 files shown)


extrawurst:
I am using `scopeguard` for this: https://crates.io/crates/scopeguard

TimonPost:
Hi, thanks for contributing. It is certainly a nice feature. Though I have one concern. 

Crossterm had this functionality in the past, we remove it. It went under rewrite were we reduced the number of abstractions and higher-level functionality like this. It made crossterm more complex and people were confused that the raw modes were disabled without there knowledge on a drop. Mainly because of the unused variable like this `let _raw = RawModeGuard::new();`  Back then crossterm did this by default. 

I prefer to keep this functionality in the hands of the user and let crossterm only do the most essential low-level actions (like enabling/disabling raw mode). Scopeguard or your way would be both available options to do this.

Any thoughts related to this matter?

Koxiaet:
@TimonPost Yes, I see now why you would want this. I think that we should mention  Scopeguard in the documentation of `enable_raw_mode` and `disable_raw_mode`, so they don't have this panicking issue.

TimonPost:
Great, maybe we can add this to the example folder? I can't do any harm, and will help some people to fix this issue quickly.

PR #426, opened 2020-05-02
parse key modifiers

andy128k:


TimonPost:
Thanks for your contribution! I've been busy lately from there the late response. I tested your branch and it seems to work well.

Issue #347, opened 2019-12-14
Add a function to set terminal emulator's title


Jarqwiz:
**Is your feature request related to a problem? Please describe.**
It would be nice to have a function to set the terminal emulator's title in a cross-platform fashion.

**Describe the solution you'd like**
It could be something like `crossterm::terminal::set_title(title: &str)`.
This function would change the window's title from something like `cargo run` or `./my_app` to something like `MyApp` indicated by the `title` parameter.

**Additional context**
I haven't found any crate that would implement this kind of behavior.
I have found something like this from [hunter](https://github.com/rabite0/hunter/blob/17c25221241c61acf6a77967471439775829abe8/src/term.rs#L57), but it seems to work only on Linux.

TimonPost:
Good idea! I would like to support this within crossterm.

Jarqwiz:
I'm pretty sure that the ANSI sequence is the best approach on Unix-like platforms. On Windows, [`SetConsoleTitle`](https://docs.rs/winapi/0.3.8/winapi/um/wincon/fn.SetConsoleTitleW.html) could be used.

matheuslessarodrigues:
Do you know if there's a way to restore title when the application ends?
I see that, on windows, we can store the original title with `GetConsoleTitleW`, and then set it later.
However, I don't know how to achieve that with ANSI sequences.

Asking, because I'm willing to make a PR. I'd like this feature in my application :)

Jarqwiz:
Hello there! It's nice to see someone willing to make this PR.

So with a quick search, I couldn't find any way to get the title only set the title. On the other hand, I don't think it is a problem, because AFAIK shell will automatically set the title to whatever the user has defined for it (often the current working directory). So the application does not have to care about resetting the title, because it's automatically reset when the application exits. You might find something useful from here https://www.tldp.org/HOWTO/pdf/Xterm-Title.pdf.

matheuslessarodrigues:
You're right! It does seem that the window's title is reset automatically (tested on powershell, cmd and git's bash all on windows). I guess I didn't notice it because I mainly use busybox's shell and it does not do that.

Anyways, we can focus on just implementing the `terminal::set_title(title: &str)` function?
If I'm not wrong, I should also make a PR to `crossterm-winapi` adding the `SetConsoleTitleW` there, right?

I may add that, at least on latest Windows 10, using `"\x1b]0;my window title\x07"` works.

TimonPost:
Nice!

[...]

PR #429, opened 2020-05-13
Implementing `SetTitle` command

matheuslessarodrigues:
Following our discussion in #347, this is my PR implementing it.

`src/terminal/sys/windows.rs` now has `set_window_title(title: &str) -> Result<()>`
It can fail with `ErrorKind::SettingTerminalTitleFailure` (do you think there should be extra info inside this variant? We'd have to call [GetLastError](https://docs.microsoft.com/en-us/windows/win32/api/errhandlingapi/nf-errhandlingapi-getlasterror?redirectedfrom=MSDN)).

What I'm not sure about is the `terminal::ansi::set_title_ansi_sequence(title: &str) -> String`. That is because it seems that setting the terminal title is not a CSI sequence (it starts with `\x1B]` instead of `\x1B[`) so I just did left the sequence there inside the `format!` instead of wrapping it inside a macro or something.

Besides that, you can find the command struct `SetTitle` inside `src/terminal.rs`. It has a straightforward implementation. From what I understand, that is enough to only use the ansi version on platforms that support it, correct?

That's it!

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -18,6 +18,7 @@ pub enum ErrorKind {
     Utf8Error(std::string::FromUtf8Error),
     ParseIntError(std::num::ParseIntError),
     ResizingTerminalFailure(String),
+    SettingTerminalTitleFailure,
     #[doc(hidden)]
     __Nonexhaustive,
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -62,6 +62,7 @@
 //!     [`ScrollDown`](terminal/struct.ScrollDown.html)
 //!   - Miscellaneous - [`Clear`](terminal/struct.Clear.html),
 //!     [`SetSize`](terminal/struct.SetSize.html)
+//!     [`SetTitle`](terminal/struct.SetTitle.html)
 //!   - Alternate screen - [`EnterAlternateScreen`](terminal/struct.EnterAlternateScreen.html),
 //!     [`LeaveAlternateScreen`](terminal/struct.LeaveAlternateScreen.html)
 //!
--- a/src/style/macros.rs
+++ b/src/style/macros.rs
@@ -7,7 +7,7 @@
 // There are four macros in each group. For `Styler`, they are:
 //  * def_attr_base,
 //  * def_attr_generic,
-//  * impl_styler_callback, 
+//  * impl_styler_callback,
 //  * impl_styler
 //
 // Fundamentally, any implementation works in a similar fashion; many methods with near-identical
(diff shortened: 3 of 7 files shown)


matheuslessarodrigues:
I should note that I get this when I run `cargo test` and I didnt investigate why :(

```
error[E0432]: unresolved import `crossterm::event::EventStream`
  --> examples\event-stream-async-std.rs:15:61
   |
15 |     event::{DisableMouseCapture, EnableMouseCapture, Event, EventStream, KeyCode},
   |                                                             ^^^^^^^^^^^ no `EventStream` in `event`

error[E0432]: unresolved import `crossterm::event::EventStream`
  --> examples\event-stream-tokio.rs:15:61
   |
15 |     event::{DisableMouseCapture, EnableMouseCapture, Event, EventStream, KeyCode},
   |                                                             ^^^^^^^^^^^ no `EventStream` in `event`

warning: unused import: `StreamExt`
  --> examples\event-stream-async-std.rs:10:42
   |
10 | use futures::{future::FutureExt, select, StreamExt};
   |                                          ^^^^^^^^^
   |
   = note: `#[warn(unused_imports)]` on by default

warning: unused import: `StreamExt`
  --> examples\event-stream-tokio.rs:10:42
   |
10 | use futures::{future::FutureExt, select, StreamExt};
   |                                          ^^^^^^^^^
   |
   = note: `#[warn(unused_imports)]` on by default

error: aborting due to previous error

For more information about this error, try `rustc --explain E0432`.
error: aborting due to previous error

For more information about this error, try `rustc --explain E0432`.
error: could not compile `crossterm`.
warning: build failed, waiting for other jobs to finish...
error: could not compile `crossterm`.
warning: build failed, waiting for other jobs to finish...
error: build failed
```

matheuslessarodrigues:
forgot to say that I made the modifications you asked!

TimonPost:
cargo test should be executed with all features enabled. Event stream is only available with event-stream feature.

image

You'll have to add `#[cfg(windows)]` to the windows function.

matheuslessarodrigues:
Oh, that fixed the compile errors! Although there are still some tests failing for me. Should I open an issue? Or is it because I'm on windows? Thanks for the help!

TimonPost:
As long that the CI passes it is okay

matheuslessarodrigues:
do you think it's good for a merge?

TimonPost:
Yea, I think the CI might be broken. Thanks for the PR! I will give it a final test before releasing it.

PR #431, opened 2020-05-17
Generate alt-arrow keys on unix

osa1:
Not sure how standard these keys are, but this patch works fine with st, konsole, and xfce4-terminal.

Context: I've been using these keys in [tiny][1] for a long time now, and I'm currently considering switching to crossterm.

[1]: https://github.com/osa1/tiny/blob/master/term_input/src/lib.rs#L68-L71

### DIFF
--- a/src/event/sys/unix/parse.rs
+++ b/src/event/sys/unix/parse.rs
@@ -142,7 +142,7 @@ pub(crate) fn parse_csi(buffer: &[u8]) -> Result<Option<InternalEvent>> {
                 if last_byte < 64 || last_byte > 126 {
                     None
                 } else {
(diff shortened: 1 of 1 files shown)


TimonPost:
I've fixed the error, please merge your branch with master.

TimonPost:
As I am reviewing PR #426 I see that this feature is already implemented there. I tested it locally and the ALT + ARROW keys are working. I think that this PR is not needed anymore. 

Though thanks for the contribution anyways!

TimonPost:
Going to close this PR, sorry :(. However, the other PR introduced better support for modifiers in general. It is just released in 0.17.5. Thanks for the effort!

Issue #432, opened 2020-05-18
deprecated `net2` in outdated `mio` dependency


extrawurst:
the `mio` dependency needs to be upgraded to `0.7` to remove `net2` from the dependency tree: https://github.com/tokio-rs/mio/issues/1322

right now `cargo audit` returns:
```
Crate:  net2
Title:  `net2` crate has been deprecated; use `socket2` instead
Date:   2020-05-01
URL:    https://rustsec.org/advisories/RUSTSEC-2020-0016
Dependency tree:
net2 0.2.34
├── miow 0.2.1
│   └── mio 0.6.22
│       ├── tokio 0.2.21
│       │   └── crossterm 0.17.3
│       ├── signal-hook 0.1.15
│       │   └── crossterm 0.17.3
│       ├── mio-uds 0.6.8
│       │   ├── tokio 0.2.21
│       │   └── async-std 1.5.0
│       │       └── crossterm 0.17.3
│       ├── mio-named-pipes 0.1.6
│       │   └── tokio 0.2.21
│       ├── crossterm 0.17.3
│       └── async-std 1.5.0
└── mio 0.6.22
```

extrawurst:
confirmed fixed in `0.17.6`

PR #435, opened 2020-05-18
WIP mio update

TimonPost:
Attempt to update to the latest mio version.

- Replace our waker with the waker from [mio](https://docs.rs/mio/0.7.0/mio/struct.Waker.html).
- Updated to 0.7.

I have one concern about this update which is mentioned in the mio [change log](https://github.com/tokio-rs/mio/blob/master/CHANGELOG.md):

> PollOpt was removed and all registrations use edge-triggers. See the upgrade guide on how to process event using edge-triggers.

And @zrzka once commented this in the source code:

```
        // PollOpt::level vs PollOpt::edge mio documentation:
        //
        // > With edge-triggered events, operations must be performed on the Evented type until
        // > WouldBlock is returned.
        //
        // TL;DR - DO NOT use PollOpt::edge.
        //
        // Because of the `try_read` nature (loop with returns) we can't use `PollOpt::edge`. All
        // `Evented` handles MUST be registered with the `PollOpt::level`.
        //
        // If you have to use `PollOpt::edge` and there's no way how to do it with the `PollOpt::level`,
        // be aware that the whole `TtyInternalEventSource` have to be rewritten
        // (read everything from each `Evented`, process without returns, store all InternalEvent events
        // into a buffer and then return first InternalEvent, etc.). Even these changes wont be
        // enough, because `Poll::poll` wont fire again until additional `Evented` event happens and
        // we can still have a buffer filled with InternalEvent events.
```

@zrzka I tested with mio 0.7 but don't seem to have any problems with this in this branch. Do you still think that this issue is relevant, or do you have any new insights on this matter?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+# Version 0.17.5
+- Improved support of keymodifier for linux, arrow keys, function keys, home keys etc.
+- Add `SetTitle` command to change the terminal title.
+- Mio 0.7 update
+
 # Version 0.17.4
 - Add macros for `Colorize` and `Styler` impls, add an impl for `String` 
 - Add shift modifier to uppercase char events on unix 
--- a/examples/event-stream-tokio.rs
+++ b/examples/event-stream-tokio.rs
@@ -43,7 +43,7 @@ async fn print_events() {
                             println!("Cursor position: {:?}\r", position());
                         }
 
-                        if event == Event::Key(KeyCode::Up.into()) {
+                        if event == Event::Key(KeyCode::Esc.into()) {
                             break;
                         }
                     }
--- a/src/event/source/unix.rs
+++ b/src/event/source/unix.rs
@@ -139,7 +139,6 @@ impl EventSource for UnixInternalEventSource {
                     }
                     #[cfg(feature = "event-stream")]
                     WAKE_TOKEN => {
-                        println!("WAKE TOKEN");
                         return Err(std::io::Error::new(
                             std::io::ErrorKind::Interrupted,
                             "Poll operation was woken up by `Waker::wake`",
--- a/src/event/source/unix.rs
+++ b/src/event/source/unix.rs
@@ -94,9 +94,6 @@ impl EventSource for UnixInternalEventSource {
                             match self.tty_fd.read(&mut self.tty_buffer, TTY_BUFFER_SIZE) {
                                 Ok(read_count) => {
                                     if read_count > 0 {
-                                        self.poll
-                                            .poll(&mut additional_input_events, Some(Duration::from_secs(0)))?;
-
                                         self.parser.advance(
                                             &self.tty_buffer[..read_count],
                                             false,
--- a/src/event/source/unix.rs
+++ b/src/event/source/unix.rs
@@ -1,8 +1,8 @@
 use mio::{unix::SourceFd, Events, Interest, Poll, Token};
 use signal_hook::iterator::Signals;
-use std::{collections::VecDeque, time::Duration, io};
+use std::{collections::VecDeque, io, time::Duration};
 
-use crate::{Result, ErrorKind};
+use crate::{ErrorKind, Result};
 
 #[cfg(feature = "event-stream")]
 use super::super::sys::Waker;
@@ -99,12 +99,16 @@ impl EventSource for UnixInternalEventSource {
                                             read_count == TTY_BUFFER_SIZE,
                                         );
                                     }
-                                },
+                                }
                                 Err(ErrorKind::IoError(e)) => {
                                     // No more data to read at the moment. We will receive another event
-                                    if e.kind() == io::ErrorKind::WouldBlock { break }
+                                    if e.kind() == io::ErrorKind::WouldBlock {
+                                        break;
+                                    }
                                     // once more data is available to read.
-                                    else if e.kind() == io::ErrorKind::Interrupted {continue}
+                                    else if e.kind() == io::ErrorKind::Interrupted {
+                                        continue;
+                                    }
                                 }
                                 Err(e) => return Err(e),
                             };
--- a/src/event/source/unix.rs
+++ b/src/event/source/unix.rs
@@ -1,4 +1,4 @@
-use mio::{unix::SourceFd, Events, Poll, Token, Interest};
+use mio::{unix::SourceFd, Events, Interest, Poll, Token};
 use signal_hook::iterator::Signals;
 use std::{collections::VecDeque, time::Duration};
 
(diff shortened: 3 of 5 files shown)


TimonPost:
https://github.com/crossterm-rs/crossterm/issues/432

TimonPost:
More on the motivation for this change in mio can be found here:  https://tokio.rs/blog/2019-12-mio-v0.7-alpha.1/


image

extrawurst:
@TimonPost is this good to go?

TimonPost:
I am not sure. There is a slight problem here. Mio made all polling operations edge-triggered. @zrzka wrote the code for this and commented that incase we want to use edge triggered everything needs to be rewritten. I've tested the code with the new mio code but couldn't find any issues at first hand. So now I am left with the question, is there an issue at all or don't I see this issue. I am afraid the second one is true. Thats keeping me from merging this one.

zrzka:
@TimonPost I missed the first notification, sorry, going to read the code and explain what I mean with the comment.

zrzka:
@TimonPost okay, here's the in depth explanation of them comment ...

**Level**

Let's say that there're 10 bytes available and you read just 5 (small buffer). Next time you call `poll` ...

https://github.com/crossterm-rs/crossterm/blob/5511f85d50ad59fd351fabb7e63b6101e742124e/src/event/source/unix.rs#L100

... it returns immediately until you consume everything what's available ...

https://github.com/crossterm-rs/crossterm/blob/5511f85d50ad59fd351fabb7e63b6101e742124e/src/event/source/unix.rs#L110

... in other words, you're not forced to consume everything what's available.

**Edge**

Let's say that there're 10 bytes and you read just 5 (small buffer). Next time you call `poll`, it doesn't return immediately and it waits for additional data even if there're still 5 bytes available. With edge, you're responsible for reading everything what's available until it returns would block.

The following code ...

https://github.com/crossterm-rs/crossterm/blob/5511f85d50ad59fd351fabb7e63b6101e742124e/src/event/source/unix.rs#L110-L128

... should be inside a loop and you have to call read until you get `EWOULDBLOCK`.

**Possible issues**

Current parser implementation requires `additional_input_available` argument and this is used to check if the sequence is valid - no additional input & incomplete sequence -> invalid. Additional input & incomplete sequence -> okay, maybe valid, just wait. You wont get it with `poll` & edge in case you read `TTY_BUFFER_SIZE` bytes and there's something left to read. And it's **mainly** here because of the Escape handling:

https://github.com/crossterm-rs/crossterm/blob/5511f85d50ad59fd351fabb7e63b6101e742124e/src/event/sys/unix/parse.rs#L36-L41

I don't know how often this can happen. Probably when someone will use `cat` with a big file, etc.

Pseudo code demonstrating how the level -> edge can be "fixed" / the issue here:

```
while (true) {
    match self.tty_fd.read(&mut self.tty_buffer, TTY_BUFFER_SIZE) {
        Ok(count) => {
            // here we don't know if the next read will block (no additional input) or not (additional input)
            // possible Esc problem issue
            self.parser.advance(&self.tty_buffer[..read_count], false); // or true? who knows
        }
        Err(EWOULDBLOCK) or Err(EAGAIN) => // break the while loop,
        Err(e) => return e,
    };
}
  
if let Some(event) = self.parser.next() {
    return Ok(Some(event));
}
```

... or ...

```
var bytes: BytesMut = BytesMut::with_capacity(TTY_BUFFER_SIZE);
while (true) {
    match self.tty_fd.read(&mut self.tty_buffer, TTY_BUFFER_SIZE) {
        Ok(count) => {
            // here we don't know if the next read will block (no additional input) or not (additional input), but we don't care, we have growing BytesMut
            bytes.extend_from_slice(&self.tty_buffer[..count]);          
        }
        Err(EWOULDBLOCK) or Err(EAGAIN) => // break the while loop,
        Err(e) => return e,
    };
}

// we can always say false, because we read everything what was available
self.parser.advance(&bytes[..], false);

if let Some(event) = self.parser.next() {
    return Ok(Some(event));
}
```

Frankly speaking, I don't like either way, because I'm not really sure about the Esc key.

All these comments are based on mio-0.6 knowledge. I don't know how mio-0.7 is implemented and if my comments are valid for this version or not.

While updating to mio-0.7 can just work for simple stuff, you have to test:

* `cat file.txt` with size bigger than `TTY_BUFFER_SIZE`
* if the Esc key is recognized or if it was swallowed as a part of Esc sequence
* if you read everything before next `poll` call
* probably something else too.

zrzka:
@TimonPost one more thing - imagine there will be this loop around read and you cat a big file. First event will be emitted after the cat ends & all bytes were read/parsed.

It was designed for level, not edge. Which means that the logic should be updated.

Current implementation has the same flaw, but there's no loop and the buffer is relatively small which makes it acceptable. But in cass of loop rhere can be a delay. Not for keystrokes, but flood of mouse events, cat file, ...

TimonPost:
Thanks for the hints and suggestions how I can test this! I'm gonna take a look at it.

zrzka:
Better validate my thoughts, it's more than half a year and I might be wrong.

TimonPost:
I looked into a possible solution and your tips. And validated if the barrel works. I see that The mio migration docs also indicate: "Users of level triggers basically need to put a loop around all I/O operations.". For now I can't deduce any problems or issues. 

I don't understand one thing though, what do you mean "use a cat file to test". Is this a file with ANSI codes that you write to the /dev/tty and then see if everything is read?

TimonPost:
I looked into a possible solution and your tips. And validated if the barrel works. I see that The mio migration docs also indicate: "Users of level triggers basically need to put a loop around all I/O operations.". For now I can't deduce any problems or issues. 

I don't understand one thing though, what do you mean "use a cat file to test". Is this a file with ANSI codes that you write to the /dev/tty and then see if everything is read? If not can you elaborate what you mean with this? 

@zrzka

src/event/source/unix.rs
```diff
@@ -107,20 +90,27 @@ impl EventSource for UnixInternalEventSource {
...
+                        loop {
+                            match self.tty_fd.read(&mut self.tty_buffer, TTY_BUFFER_SIZE) {
+                                Ok(read_count) => {
+                                    if read_count > 0 {
+                                        self.poll
+                                            .poll(&mut additional_input_events, Some(Duration::from_secs(0)))?;
+
+                                        self.parser.advance(
+                                            &self.tty_buffer[..read_count],
+                                            false,
```
zrzka, line 102: If you pass `false` here, then ...

https://github.com/crossterm-rs/crossterm/blob/5f1cbdeda4eaa4f36dcf9240a72ec0b92b45738b/src/event/sys/unix/parse.rs#L36-L41

... is of no use and the argument can be removed.

src/event/source/unix.rs
```diff
@@ -107,20 +90,27 @@ impl EventSource for UnixInternalEventSource {
...
-
-                            self.parser.advance(
-                                &self.tty_buffer[..read_count],
-                                additional_input_available,
-                            );
+                        loop {
+                            match self.tty_fd.read(&mut self.tty_buffer, TTY_BUFFER_SIZE) {
+                                Ok(read_count) => {
+                                    if read_count > 0 {
+                                        self.poll
```
zrzka, line 97: This line doesn't make a sense because you removed `additional_input_available` variable computation.

src/event/source/unix.rs
zrzka: Or keep it, don't always pass `false`, but rather something like `read_count == TTY_BUFFER_SIZE`.

src/event/source/unix.rs
```diff
@@ -149,7 +139,7 @@ impl EventSource for UnixInternalEventSource {
                     }
                     #[cfg(feature = "event-stream")]
                     WAKE_TOKEN => {
-                        let _ = self.waker.reset();
+                        println!("WAKE TOKEN");
```
zrzka, line 142: Just something to address before merge ...

examples/event-stream-tokio.rs
```diff
@@ -43,7 +43,7 @@ async fn print_events() {
                             println!("Cursor position: {:?}\r", position());
                         }
 
-                        if event == Event::Key(KeyCode::Esc.into()) {
+                        if event == Event::Key(KeyCode::Up.into()) {
```
zrzka, line 46: ? Usage is _Hit Esc to quit_ ...

src/event/source/unix.rs
zrzka: Hmm, but still, it wont solve the Esc issue completely.

examples/event-stream-tokio.rs
TimonPost: Yea, wanted to test ESC key ;)

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Version 0.17.5
 - Improved support of keymodifier for linux, arrow keys, function keys, home keys etc.
 - Add `SetTitle` command to change the terminal title.
+- Mio 0.7 update
```
extrawurst, line 4: @TimonPost this is not part of `0.17.5` right?

TimonPost: I will update the changelog before releasing

PR #449, opened 2020-06-30
Implement styler traits for `&'a str`

sharnoff:
I ran into a case where I wanted an implementation of `Colorize` for `&'a str`, not just `&'static str`, so I added it in. There's a weird case where some parentheses are required to compile but marked by the compiler as unnecessary, so I put `#[allow(unused_parens)]` alongside a little coment about them.

The change was (thankfully) relatively simple - just moving the implementation for `str` to use the more specialized macro instead.

### DIFF
--- a/src/style.rs
+++ b/src/style.rs
@@ -161,12 +161,22 @@ where
 }
 
 impl_colorize!(String);
-impl_colorize!(&'static str);
 impl_colorize!(char);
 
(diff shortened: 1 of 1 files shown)


sharnoff:
Thanks!

> does it still work with static life time strings?

I'm confident it should, but I haven't tested it - AFAIK `&'static str` can be used in place of any `&'a str`.

PR #451, opened 2020-07-01
Add ANSI color parsing and expose `Colored` enum

JOT85:
This adds `parse_ansi` functions to the `Colored` and `Color` enums. These try to parse ANSI color sequences which appear in places such as /etc/os-release (https://www.freedesktop.org/software/systemd/man/os-release.html) and other configuration files.

It's not too hard to read the string from the config and place it between `ESC [` and `m` that lacks validation and isn't super nice. The parsing methods a simple and allow for the sequences to be validated and used with crossterm (for example in a styling configuration).
Also, this means that these config files can be used cross-platform - I'm not entirely sure if a Windows user would ever put and ANSI color sequence in a config file but now they could!

There were a few things I was unsure about and would like feedback on. The first is exposing the `Colored` enum. The escape sequences parsed may either be a background or a foreground color so the return type must be able to express this. The Colored enum exactly expressed this so I thought it would be the most sensible thing to use since it already has display implementations etc.
If you're happy with this being exposed, then it might be nice to provide commands for using it?

Also, I've not yet included parsing of the 3/4 bit color values since they aren't currently expressible with `Color`.
They could be mapped to the corresponding existing variant (for example mapping `31 -> Red`) although then the result when these are used might not exactly match what the input was (for example `Red -> 5;9`). Alternatively, a new variant could be added or these could be left as unsupported.

Tests are provided using the formatter for Color and Colored enums.


src/style/ansi.rs
```diff
@@ -78,43 +78,256 @@ impl fmt::Display for Colored {
...
+    ///              `38;5;26 -> ForegroundColor(AnsiValue(26))`
+    ///              `48;2;50;60;70 -> BackgroundColor(Rgb(50, 60, 70))`
+    ///              `49 -> BackgroundColor(Reset)`
+    /// Invalid sequences map to None.
+    ///
+    /// Currently, 3/4 bit color values aren't supported so return None.
+    ///
+    /// See also: [Color::parse_ansi](enum.Color.html#method.parse_ansi)
+    pub fn parse_ansi(ansi: &str) -> Option<Self> {
+        use Colored::{BackgroundColor as BG, ForegroundColor as FG};
```
TimonPost, line 103: I would just leave this away and use them inline. I find it confusing if types are renamed makes code a bit more difficult to read imo.

src/style.rs
```diff
@@ -117,11 +117,10 @@ use crate::Result;
 use crate::{impl_display, Ansi, Command};
 use std::fmt;
 
-pub(crate) use self::enums::Colored;
 pub use self::{
     attributes::Attributes,
     content_style::ContentStyle,
-    enums::{Attribute, Color},
+    enums::{Attribute, Color, Colored},
```
TimonPost, line 123: Colored was exposed some versions ago, but we have removed it from the API to make it easier and create one way of styling. `Colored` was used to set for and background color, since we have moved to the command API it was no longer necessary to expose it. 

> If you are happy with this exposure, then maybe it's nice to give commands for its use?

I think that the commands are not really an option here. Because they are not made for returning values, only for executing. 

If I look at the functions, you return the type `Colored` and as you said, this type is needed to distinguish the background from the foreground colors. I don't have a problem with exposing it. It is a simple structure and does not introduce any additional api complexity. 

An alternative would be to create a stand-alone function to translate an ansi code to a color. Similar things have been done with `size`, `pos`, similar to this functionality they return a value and therefore cannot be used within a command. However, we woul still need to expose the `Colored` structure. 

I do prefer to expose the Colored and keep the function in there

src/style/ansi.rs
JOT85: Thanks :). I've changed it to not rename them. I've left the `use Colored::{BackgroundColor, ForegroundColor}` in, to keep the lines where they're used shorter. I think they look quite cumbersome with the fully qualified names. I'm happy to remove the use statement if you disagree though.

src/style/ansi.rs
```diff
@@ -78,43 +78,256 @@ impl fmt::Display for Colored {
...
+    /// For example: `5;0 -> Black`, `5;26 -> AnsiValue(26)`, `2;50;60;70 -> Rgb(50, 60, 70)`.
+    /// Invalid sequences map to None.
+    ///
+    /// Currently, 3/4 bit color values aren't supported so return None.
+    ///
+    /// See also: [Colored::parse_ansi](enum.Colored.html#method.parse_ansi)
+    pub fn parse_ansi(ansi: &str) -> Option<Self> {
+        Self::parse_ansi_iter(&mut ansi.split(';'))
+    }
+    /// The logic for parse_ansi, takes an iterator of the sequences terms (the numbers between the
```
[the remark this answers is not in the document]
JOT85: 85b55bbdf8a81ce70679858edd26e0ab27bed1c8

src/style.rs
JOT85: > I do prefer to expose the Colored and keep the function in there

Perfect, thank you!

> I think that the commands are not really an option here. Because they are not made for returning values, only for executing.

I was more thinking a command such as `SetColored` which takes a `Colored` value and applies it, just for convenience of using these values. That would add another method to set for and background colors though.

TimonPost: Right, I think it can't realy hurt to add it. I see it quite often that SetForeground and SetBackground are used together. Maybe this command would offer a nice shortcut.

JOT85: > What I originally meant was that first there were multiple ways to style. As long as the functionalities go trough the command API we are good.

That makes sense!

The obvious API is to add a command, `SetColored(pub Colored)` (or `SetColor(pub Colored)`) which set either the foreground or background color.

> I see it quite often that SetForeground and SetBackground are used together. Maybe this command would offer a nice shortcut.

The API above doesn't really provide a nice way to do that. How about an API such as:
```rust
pub struct Colors {
    pub foreground: Option<Color>,
    pub background: Option<Color>,
}

impl From<Colored> for Colors {
    // ...
}

// ...

/// A command which optionally sets the colors
pub struct SetColors(pub Colors)
``` 
That provides a shortcut to set both the foreground and background color but also means that we wouldn't need a `SetColored` command since you could write `SetColores(colored.into())`.

TimonPost: Ah, thanks for pointing out, I forgot the fact that we don't want to do `SetColored(pub Colored)` because it is almost identifcal to `SetFor/BackgroundColor(Color)`. 

The proposed API seems good to me. What do you think of: `pub struct Colors(Color, Color);` were the first entry is the foreground color and the second entry is the background color?

JOT85: > What do you think of: pub struct Colors(Color, Color); were the first entry is the foreground color and the second entry is the background color?

That API also sounds good.

The main reason I suggested it with the distinct `Colors` type is to allow for conveniently applying a `Colored` value. That's probably a fairly niche use case though and it's easily done with a tiny match expression.

We could, with the first proposal, also provide a `setColors` function such as:
```rust
fn setColors(foreground: Color, background: Color) -> SetColors {
    SetColors(Colors {
        foreground: Some(foreground),
        background: Some(background),
    })
}
```

That allows you to use `setColors(fg, bg)` as a command, along with `SetColors(colored.into())` as a command, with only one actual `SetColors` command.

That makes a reasonably simple API, but not as simple as just having a `SetColors` command without a distinct type - that however doesn't make using `Colored` convenient.

I think either solution is good, maybe it's best to stick to the most simple case.

TimonPost: I am not in favor of a `setColors` function, it introduces an extra way to use the crossterm API, I would rather stick with a single command. 

My question was more about something else. We can go about this in two ways:

```
pub struct Colors {
    pub foreground: Option<Color>,
    pub background: Option<Color>,
}

Colors { foreground: Green, background: Blue }
```

or 

```
pub struct Colors(pub Option<Color>, pub Option<Color>)

Colors(Gree, Blue)
```

The struct version has some overhead with the `foreground` and `background` colors and thus I tought about an alternative version.

They way doesnt matter directly, both can be used like this:

```
impl From<Colored> for Colors {
    // ...
}

// ...

/// A command which optionally sets the colors
pub struct SetColors(pub Colors)

SetColors(colored.into())
SetColors(Colored (...))
```

Maybe I do prefer your version that makes the variable name explicit.

JOT85: I've added that :)

I've also added `impl From<(Color, Color)> for Colors` which makes setting both known colors more convenient although I'm happy to remove it again if you feel like it's unnecessary.

TimonPost: Sure sounds good!

src/style.rs
```diff
@@ -183,6 +182,9 @@ pub fn available_color_count() -> u16 {
 ///
 /// See [`Color`](enum.Color.html) for more info.
 ///
+/// [`SetColors`](struct.SetColors.html) can also be used to set both the foreground and background
+/// color with 1 command.
```
TimonPost, line 186:
```suggestion
/// color in one command.
```

src/style.rs
```diff
@@ -213,6 +215,9 @@ impl Command for SetForegroundColor {
...
```
TimonPost, line 219:
```suggestion
/// color in one command.
```

src/style.rs
```diff
@@ -239,6 +244,61 @@ impl Command for SetBackgroundColor {
     }
 }
 
+/// A command that optionally sets the foreground and/or background color.
+///
+/// For example:
+/// ```
```
TimonPost, line 250:
```suggestion
/// ```no_run
```

src/style/types/colored.rs
```diff
@@ -0,0 +1,18 @@
+use crate::style::Color;
+
+#[cfg(feature = "serde")]
+use serde::{Deserialize, Serialize};
+
+/// Represents a foreground or background color.
+///
+/// This can be converted to a [Colors](struct.Colors.html) struct and then applied using the
+/// [SetColors](struct.SetColors.html) command, for example
+/// `stdout.execute(SetColors(colored.into()))`.
```
TimonPost, line 10:
```suggestion
```

src/style/types/colored.rs
```diff
@@ -0,0 +1,18 @@
+use crate::style::Color;
+
+#[cfg(feature = "serde")]
+use serde::{Deserialize, Serialize};
+
+/// Represents a foreground or background color.
+///
+/// This can be converted to a [Colors](struct.Colors.html) struct and then applied using the
```
TimonPost, line 8:
```suggestion
/// This can be converted to a [Colors](struct.Colors.html) by calling `into()`. It can be applied by using the
```

src/style/types/colors.rs
```diff
@@ -0,0 +1,243 @@
...
+///
+/// It can be applied using the `SetColors` command.
+///
+/// It can also be created from a [Colored](enum.Colored.html) value or a tuple of
+/// `(Color, Color)` in the order `(foreground, background)`.
+///
+/// The [then](#method.then) method can be used to combine `Colors` values.
+///
+/// For example:
+/// ```
```
TimonPost, line 13:
```suggestion
/// ```no_run
```

JOT85:
> One last round then we can merge it. Nice work! You should be able to accept my proposed changes.

Thank you very much :)
I've made the suggested changes and added the `Colors::new` method and removed `From<(Color, Color)>`

TimonPost:
Never mind, I will do it my self. Will release a new version today. Great work!

JOT85:
> Never mind, I will do it my self. Will release a new version today. Great work!

Sorry about that - I wasn't available at that time! Thanks though :)

Issue #445, opened 2020-06-15
Queuing and WinApi


sigmaSd:
Consider this example:
```rust
queue!(stdout(), SetForegroundColor(Blue));
queue!(stdout(), Print("Hello World");
queue!(stdout(), ResetColor());
```
The expected behaviour is we get a blue coloured hello world, instead we get an uncoloured one(white)
This is because with winapi SetForgroundColor and ResetColor (and some other commands) are eagerly executed as you can see here https://github.com/crossterm-rs/crossterm/blob/master/src/macros.rs#L36, while Print() will only execute later after we flush the buffer as attested here https://docs.rs/crossterm/0.17.5/src/crossterm/style.rs.html#365 

If I recall correctly I remember that you stated before that winapi is not compatible with queuing system and will execute eagerly instead, so in that case a simple fix would be to change `print!` to `write! + flush` here https://docs.rs/crossterm/0.17.5/src/crossterm/style.rs.html#365

sigmaSd:
Also side-note: considering https://github.com/crossterm-rs/crossterm/blob/master/src/macros.rs#L36 maybe `execute_winapi` should also take `$writer` as arg like `write_ansi_code` so we don't have to hard-code the output to `stdout` like it is the case currently

sigmaSd:
Also another idea, maybe its a good practice to add a [compile_error!](https://doc.rust-lang.org/std/macro.compile_error.html) when queue is used with winapi, obviously it will need some refactoring since `execute!` uses `queue!` internally but its doesn't look too difficult.

This way the interaction between queuing and winapi becomes explicit, and this spares the library documenting it

TimonPost:
> If I recall correctly I remember that you stated before that winapi is not compatible with queuing system

The main reason why this happened is because WINAPI calls are system calls and are not written to a 'writer'. This results in the fact that you cannot buffer them like we can with ANSI codes. 

> so in that case a simple fix would be to change print! to write! + flush here

This is not the functionality that is desirable, is it? Print' should be the same as 'print' and should not be executed immediately.

> maybe execute_winapi should also take $writer as arg like write_ansi_code so we don't have to hard-code the output to stdout like it is the case currently

I'm not sure I'm following you here. Namely, the WINAPI code has no idea of a writer and works completely without a 'writer'. Could you explain a little more?

TimonPost:
[...]

PR #468, opened 2020-07-18
fix_winapi

sigmaSd:
Fix for https://github.com/crossterm-rs/crossterm/issues/445

What this pr does:

1- `execute_winapi` now accepts a `&mut dyn std::io::Write` which is only used with `Print` command

2- `Print` command flushes the writer in case of winapi regardless of queuing

3 - Implementing Display trait used `queue!` internally (in macros.rs and the manual implementation in styled_content.rs), which give a `$writer` that is `std::fmt::Formatter` incompatible with `std::io::Write`. The solution implemented here is to use a new internal macro instead of `queue` `handle_fmt_command` which does almost the same thing except in the case of `execute_winapi` it ignores the given `$writer` and just use `std::io::stdout` as a placeholder to satisfy type requirement (its guaranteed not to be used since Print is the only command that uses the writer, and Print is not called from `handle_fmt_command`) 

Note: <del>No API change</del> some changes might be required

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -61,7 +61,7 @@ impl std::io::Write for FakeWriter {
     }
 }
 
-// Offer the same functionality as queue! macro, but is used only internally and with std::fmt::Write as $writer
+// Offer the same functionality as `queue!` macro, but is used only internally and with `std::fmt::Write` as `$writer`
 // The difference is in case of winapi we ignore the $writer and use a fake one
 #[doc(hidden)]
 #[macro_export]
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -33,7 +33,35 @@ macro_rules! handle_command {
             if command.is_ansi_code_supported() {
                 write_ansi_code!($writer, command.ansi_code())
             } else {
-                command.execute_winapi().map_err($crate::ErrorKind::from)
+                command
+                    .execute_winapi(&mut $writer)
+                    .map_err($crate::ErrorKind::from)
+            }
+        }
+        #[cfg(unix)]
+        {
+            write_ansi_code!($writer, $command.ansi_code())
+        }
+    }};
+}
+
+#[doc(hidden)]
+#[macro_export]
+macro_rules! handle_fmt_command {
+    ($writer:expr, $command:expr) => {{
+        use $crate::{write_ansi_code, Command};
+
+        #[cfg(windows)]
+        {
+            let command = $command;
+            if command.is_ansi_code_supported() {
+                write_ansi_code!($writer, command.ansi_code())
+            } else {
+                command
+                    // $writer is not used in execute_winapi when called from handle_fmt_command macro
+                    // we give it stdout() as place holder to satisfy the type requirement
+                    .execute_winapi(&mut std::io::stdout())
+                    .map_err($crate::ErrorKind::from)
             }
         }
         #[cfg(unix)]
@@ -160,7 +188,7 @@ macro_rules! impl_display {
     (for $($t:ty),+) => {
         $(impl ::std::fmt::Display for $t {
             fn fmt(&self, f: &mut ::std::fmt::Formatter<'_>) -> ::std::result::Result<(), ::std::fmt::Error> {
-                $crate::queue!(f, self).map_err(|_| ::std::fmt::Error)
+                $crate::handle_fmt_command!(f, self).map_err(|_| ::std::fmt::Error)
             }
         })*
     }
@@ -312,7 +340,7 @@ mod tests {
                 self.value
             }
 
-            fn execute_winapi(&self) -> CrosstermResult<()> {
+            fn execute_winapi(&self, _writer: &mut dyn std::io::Write) -> CrosstermResult<()> {
                 self.stream.borrow_mut().push(self.value);
                 Ok(())
             }
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -49,7 +49,7 @@ macro_rules! handle_command {
 // It's used in execute_winapi when called from handle_fmt_command! macro
 // In that case execute_winapi doesn't care about the $writer and we only use it to fulfill type requirement
 // The unimplemented! macro asserts that $writer is indeed not used
-struct FakeWriter;
+pub(crate) struct FakeWriter;
 
 impl std::io::Write for FakeWriter {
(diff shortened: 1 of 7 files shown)


src/macros.rs
```diff
@@ -33,7 +33,35 @@ macro_rules! handle_command {
...
+        #[cfg(windows)]
+        {
+            let command = $command;
+            if command.is_ansi_code_supported() {
+                write_ansi_code!($writer, command.ansi_code())
+            } else {
+                command
+                    // $writer is not used in execute_winapi when called from handle_fmt_command macro
+                    // we give it stdout() as place holder to satisfy the type requirement
+                    .execute_winapi(&mut std::io::stdout())
```
TimonPost, line 63: Can we use something like this:

```
struct FakeStdout;

impl Write for FakeStdout {
    fn write(&mut self, buf: &[u8]) -> Result<usize> {
        unimplemented!("This is a fake stdout and is not used anywhere")
    }

    fn flush(&mut self) -> Result<()> {
        unimplemented!("This is a fake stdout and is not used anywhere")
    }
}
```

The problem if we are using `&mut std::io::stdout()` is that it will request a write unlock for stdout which is unnecessary. We can also use a fake `Write` implementation.

src/macros.rs
```diff
@@ -33,7 +33,35 @@ macro_rules! handle_command {
...
+        #[cfg(unix)]
+        {
+            write_ansi_code!($writer, $command.ansi_code())
+        }
+    }};
+}
+
+#[doc(hidden)]
+#[macro_export]
+macro_rules! handle_fmt_command {
```
TimonPost, line 50: I would like to see a short description why this macro exists. Something in the lines of: "Offers the same functionality as queue but is used for fmt::Write"

src/macros.rs
sigmaSd: I like this so much! (especially the guarantee we get from unimplemented!).
I just changed the name to FakeWriter, I think its more correct.

src/macros.rs
sigmaSd: Comment added.

sigmaSd:
When you believe this pr is ready to be merged, I'll squash the commits.

src/macros.rs
```diff
@@ -33,7 +33,51 @@ macro_rules! handle_command {
...
+        }
+        #[cfg(unix)]
+        {
+            write_ansi_code!($writer, $command.ansi_code())
+        }
+    }};
+}
+
+// This is a fake writer
+// It's used in execute_winapi when called from handle_fmt_command! macro
```
TimonPost, line 49:
```suggestion
// It's used in `execute_winapi`when called from `handle_fmt_command!` macro.
```

src/macros.rs
```diff
@@ -33,7 +33,51 @@ macro_rules! handle_command {
...
+// In that case execute_winapi doesn't care about the $writer and we only use it to fulfill type requirement
```
TimonPost, line 50:
```suggestion
// In that case, `execute_winapi` doesn't care about the `$writer` and we only use it to fulfill type requirement.
```

src/macros.rs
```diff
@@ -33,7 +33,51 @@ macro_rules! handle_command {
...
+// The unimplemented! macro asserts that $writer is indeed not used
```
TimonPost, line 51:
```suggestion
// The `unimplemented!` macro asserts that `$writer` is indeed not used.
```

src/macros.rs
```diff
@@ -33,7 +33,51 @@ macro_rules! handle_command {
...
+    fn write(&mut self, _buf: &[u8]) -> std::io::Result<usize> {
+        unimplemented!("This is a fake writer and is not used anywhere")
+    }
+
+    fn flush(&mut self) -> std::io::Result<()> {
+        unimplemented!("This is a fake writer and is not used anywhere")
+    }
+}
+
+// Offer the same functionality as queue! macro, but is used only internally and with std::fmt::Write as $writer
```
TimonPost, line 64:
```suggestion
// Offer the same functionality as `queue!` macro, but is used only internally and with `std::fmt::Write` as `$writer`
```

src/macros.rs
```diff
@@ -33,7 +33,51 @@ macro_rules! handle_command {
...
+// The difference is in case of winapi we ignore the $writer and use a fake one
```
TimonPost, line 65:
```suggestion
// The difference is in case of winapi we ignore the `$writer` and use a fake one.
```

TimonPost:
Also, the CI errors are because of some other error that's not related to your PR. I need to fix that.

sigmaSd:
cargo test on windows fails on https://github.com/sigmaSd/crossterm/blob/impl_Iterator_for_Attributes/src/macros.rs#L386 cannot borrow writer as mutable , I need to figure this out before merging

sigmaSd:
Just to explain the issue: 
1- this pr tries not to break the API -> queue!(writer) and queue!(&mut writer) should work but its not currently the case

https://github.com/sigmaSd/crossterm/blob/fix_winapi/src/macros.rs#L37 $writer here could be `std::io::Write` or `&mut std::io::Write`, while its hard-coded here https://github.com/sigmaSd/crossterm/blob/fix_winapi/src/command.rs#L30 to &mut 

I think it should be possible to accept both since this is implemented on std `impl<'_, W: Write + ?Sized> Write for &'_ mut W`

sigmaSd:
The solution that is currently used is to hard-code to &mut Write and always take a &mut reference of the argument internally https://github.com/sigmaSd/crossterm/blob/fix_winapi/src/macros.rs#L37 which works most of the time but some adjustement might be needed, for example https://github.com/sigmaSd/crossterm/blob/impl_Iterator_for_Attributes/src/macros.rs#L392 fails, it can be fixed by adding mut `test_harness(&["cmd1", "cmd2"], |mut writer, stream|`

sigmaSd:
I prefer if I find a can find a real fix for this because currently
```rust
fn a(w: &mut W)
queue!(w, ...)
```
wont work unless I add mut like 
```rust
 fn a(mut w: &mut W)
```

Obviously one other solution is to break the api and accept only &mut W, this will make the above work

sigmaSd:
closed in favour of https://github.com/crossterm-rs/crossterm/pull/469

PR #469, opened 2020-07-18
Fix winapi

sigmaSd:
Fix for https://github.com/crossterm-rs/crossterm/issues/445

What this pr does:

1- `execute_winapi` now accepts a function `FnMut() -> Result<()>`, which is 
```rust
|| {
  write!($writer, "{}", command.ansi_code())?;
 // winapi doesn't support queuing
 $writer.flush()?;
 Ok(())
}
```
this function is only called with Print command in winapi case

2-  - Add handle_fmt_command! macro to be used internally by Display trait implementation (instead of queue!) in order to avoid $writer type conflict (std::io::Write vs std::fmt::Write)

Note: No API breakage

### DIFF
--- a/src/command.rs
+++ b/src/command.rs
@@ -27,7 +27,7 @@ pub trait Command {
     ///
     /// This method does not need to be accessed manually, as it is used by the crossterm's [Command Api](../#command-api)
     #[cfg(windows)]
-    fn execute_winapi(&self) -> Result<()>;
+    fn execute_winapi(&self, writer: impl FnMut() -> Result<()>) -> Result<()>;
 
     /// Returns whether the ansi code representation of this command is supported by windows.
     ///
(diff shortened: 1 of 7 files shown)


sigmaSd:
The problem with that approach is that functions don't capture the environment so we will end up with something like this 
`    fn execute_winapi<W: std::io::Write>(&self, writer: fn(W) -> Result<()>, command: Command) -> Result<()>;`
I'm not sure it will work, also I think the cost is really negligible.

sigmaSd:
Btw I think closures are also on the stack so unnecessary allocations shouldn't happen.

sigmaSd:
According to https://stackoverflow.com/questions/27831944/how-do-i-store-a-closure-in-a-struct-in-rust, you can allocate the closure on the heap by calling Box on it (so its on the stack).
Also out of the 3 methods (Generic parameters, Trait references, Boxed closures) Generic parameters is the most efficient, which is what is used here.

TimonPost:
Ah, thanks for the insights, your right. Let us keep it like this then.

PR #474, opened 2020-08-27
Switching to futures-lite, reduces build time.

halzy:
While using `tui` I noticed that my build time increased quite a bit when switching from the `termion` backend to `crossterm`. Updating `crossterm` to use [`futures-lite`](https://crates.io/crates/futures-lite) dropped the build time on my box by **15 seconds**.

Note: The build speed improvement only happens when using the `event-stream` feature.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,7 +27,7 @@ all-features = true
 #
 [features]
 default = []
-event-stream = ["futures-util"]
+event-stream = ["futures-lite"]
 
 #
 # Shared dependencies
@@ -36,7 +36,7 @@ event-stream = ["futures-util"]
 bitflags = "1.2"
 lazy_static = "1.4"
 parking_lot = "0.10"
-futures-util = { version = "0.3", optional = true, default-features = false }
+futures-lite = { version = "1.0", optional = true, default-features = false }
 serde = { version = "1.0", features = ["derive"], optional = true }
 
 #
--- a/examples/event-stream-async-std.rs
+++ b/examples/event-stream-async-std.rs
@@ -7,8 +7,8 @@ use std::{
     time::Duration,
 };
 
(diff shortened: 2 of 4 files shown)


halzy:
These clippy errors appear unrelated to the PR.

archseer:
I'd like to see this merged too but I wonder if it breaks compatibility anywhere. `futures-util` is really heavy because it pulls in `pin_project` which relies on a bunch of proc_macro/syn etc that really increase the compile time.

archseer:
I wonder if we could get away with just relying on `futures-core` which would be an even smaller subset of `futures-lite`. It seems to just rely on `Stream` and task.

archseer:
Seems to work for me and is even faster:

```
❯ c b --features event-stream
   Compiling libc v0.2.76
   Compiling cfg-if v0.1.10
   Compiling log v0.4.11
   Compiling bitflags v1.2.1
   Compiling arc-swap v0.4.7
   Compiling scopeguard v1.1.0
   Compiling smallvec v1.4.2
   Compiling lazy_static v1.4.0
   Compiling futures-core v0.3.5
   Compiling lock_api v0.3.4
   Compiling signal-hook-registry v1.2.1
   Compiling parking_lot_core v0.7.2
   Compiling mio v0.7.0
   Compiling parking_lot v0.10.2
   Compiling signal-hook v0.1.16
   Compiling crossterm v0.17.7 (/Users/speed/src/crossterm)
warning: associated function is never used: `reset`
  --> src/event/sys/unix/waker.rs:36:19
   |
36 |     pub(crate) fn reset(&self) -> Result<()> {
   |                   ^^^^^
   |
   = note: `#[warn(dead_code)]` on by default

warning: 1 warning emitted

    Finished dev [unoptimized + debuginfo] target(s) in 4.67s
```

TimonPost:
Can be closed because of #480

Issue #457, opened 2020-07-07
I think crossterm::cursor::MoveRight(0) should not move the cursor


sunshinewithmoonlight:
crossterm::cursor::MoveRight(0) moves the cursor one position to the right instead of zero.  
I want to know if 0 really works when I move the cursor right by 1 to 0？

TimonPost:
Agree, if this is the case it should not move the cursor position.

MichaelMcDonnell:
Similarly, doing either MoveLeft(0) or MoveLeft(1) also moves to the left by 1 which is confusing.

TimonPost:
Implemented by #476

PR #476, opened 2020-09-06
Fix move by zero

MichaelMcDonnell:
Moving the cursor by zero currently moves it by one. This only happens in terminals with ANSI support.

I have published a small test example called [mz_crossterm_test](https://github.com/MichaelMcDonnell/mz_crossterm_test) and tested the behavior with it.

This fixes issue #457.


TimonPost (approved):
Awesome, thanks!

PR #477, opened 2020-09-06
Fix clippy warnings

MichaelMcDonnell:
The current clippy warnings are causing the GitHub Actions build to fail. The rule has been renamed.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -12,6 +12,7 @@ pub type Result<T> = std::result::Result<T, ErrorKind>;
 
 /// Wrapper for all errors that can occur in `crossterm`.
 #[derive(Debug)]
+#[non_exhaustive]
 pub enum ErrorKind {
     IoError(io::Error),
     FmtError(fmt::Error),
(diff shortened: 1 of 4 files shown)


MichaelMcDonnell:
Looks like there are still two warnings left.

TimonPost:
Thanks! 

Its about the Attribute enum and Error enum

https://github.com/crossterm-rs/crossterm/blob/ddc25d4225dd8e02d485000bd173a8978a2ed022/src/error.rs#L23

https://github.com/crossterm-rs/crossterm/blob/ddc25d4225dd8e02d485000bd173a8978a2ed022/src/style/types/attribute.rs#L143

If you see the opportunity to remove them feel free to do so.

MichaelMcDonnell:
I've updated the pull request. I fixed the non_exhaustive attribute warnings but then I got another warning about matches. I fixed that one too. There are no warning left now.

TimonPost:
Awesome and thanks!

Issue #482, opened 2020-09-09
Reset Foreground/Background color


truchi:
Should be easy:
- add `Color::ResetForeground` (39) & `Color::ResetBackground` (49)
- reset colors properly in `StyledContent::fmt`
- remove `Color::Reset`, `ResetColor` & `RESET_CSI_SEQUENCE`?

Do you want me to do it?

Edit: Oh wait, it's actually #2 only!

truchi:
Heck I'm doing it.

TimonPost:
Merged, new issue is to be created for attributes individual resetting

PR #483, opened 2020-09-09
StyledContent::fmt: reseting foreground/background on their own

truchi:
Fix for #482 

Please read the NOTE and TODO in the code. There is still a bug when printing a styled content with attributes but no colors.


TimonPost (approved):

TimonPost:
Thanks! It improves the situation as it is now

Issue #359, opened 2020-01-08
AZERTY keyboard suppoirt (Windows)


3kyro:
**Describe the bug**
In AZERTY keyboards (Belgian french in this case) some common ascii characters ( { ,} , [, ] , #, \ ) are only accessible with the right Alt button pressed. In windows, no event is generated for these characters.

**To Reproduce**
1. Run the event-poll-read example on a windows/AZERTY machine and try to get a KeyEvent for any of the above characters

**OS**
 - Windows 10

**Terminal/Shell**
 - CMD/Powershell

In crossterm/src/event/sys/windows/parse.rs there is no differentiation of right or left key pressed

`let alt = state.has_state(LEFT_ALT_PRESSED | RIGHT_ALT_PRESSED);`

while later in the same file, in `parse_key_event_record` function the key code is replaced by the virtual_key_code which no longer corresponds to the right character.

``` Rust
                if modifiers.contains(KeyModifiers::ALT) {
                    // If the ALT key is held down, pressing the A key produces ALT+A, which the system does not treat as a character at all, but rather as a system command.
                    // The pressed command is stored in `virtual_key_code`.
                    let command = key_event.virtual_key_code as u8 as char;

                    if command.is_alphabetic() {
                        character = command;
                    } else {
                        return None;
                    }
```

Here is also an output from the event-poll-read example with some debug information.

```
[src\event\sys\windows\parse.rs:80] character_raw = 93
[src\event\sys\windows\parse.rs:88] command = 'º'
Event::Key(KeyEvent { code: Char('º'), modifiers: CONTROL | ALT })
```

TimonPost:
Thanks for filing the issue, your correct, crossterm in general does not differentiate between left/right ctrl, shift, control. I will have to look into those key codes.

sharnoff:
This seems to be somewhat related to #373 - given the limitations of other terminal applications (and the documentation for xterm, etc.) it seems like terminal emulators won't actually provide different key codes for left/right control or shift+ctrl. More info: [https://vi.stackexchange.com/a/4291](https://vi.stackexchange.com/a/4291).

sigmaSd:
Hi, removing the else clause here fixes the issue
```rust
                if modifiers.contains(KeyModifiers::ALT) {
                    // If the ALT key is held down, pressing the A key produces ALT+A, which the system does not treat as a character at all, but rather as a system command.
                    // The pressed command is stored in `virtual_key_code`.
[...]

PR #488, opened 2020-09-17
Fix azerty keyboard handling for windows

sigmaSd:
Fix https://github.com/crossterm-rs/crossterm/issues/359

This pr adresses 2 issues:

1- Ignore invalid keys (for example pressing numpad lock, adds a space instead of doing nothing)
2- Accept (nums + alt) key combination by removing the early return, On windows at least on my testing the whole `if modifiers.contains(alt_key)` block is not needed, I added a comment for now till things becomes clear

I should note that on  my testing on azerty windows, I noticed that the left alt key is different from the right one:

- The right alt key is the important one , it returns `KeyModifiers::CONTROL |KeyModifiers::ALT`
- The left alt key (I don't know what this is used for, it doesn't do anything as far as I know on windows), it returns `KeyModifers::ALT`

edit: turns out its windows way of handling [AltGr](https://en.wikipedia.org/wiki/AltGr_key)

### DIFF
--- a/src/event/sys/windows/parse.rs
+++ b/src/event/sys/windows/parse.rs
@@ -79,19 +79,16 @@ fn parse_key_event_record(key_event: &KeyEventRecord) -> Option<KeyEvent> {
             let character_raw = key_event.u_char;
 
             if character_raw < 255 {
+                // Invalid character
+                if character_raw == 0 {
+                    return None;
+                }
+
                 let mut character = character_raw as u8 as char;
 
-                if modifiers.contains(KeyModifiers::ALT) {
-                    // If the ALT key is held down, pressing the A key produces ALT+A, which the system does not treat as a character at all, but rather as a system command.
-                    // The pressed command is stored in `virtual_key_code`.
-                    let command = key_event.virtual_key_code as u8 as char;
-
-                    if command.is_alphabetic() {
-                        character = command;
-                    } else {
-                        return None;
-                    }
-                } else if modifiers.contains(KeyModifiers::CONTROL) {
+                if modifiers.contains(KeyModifiers::CONTROL)
+                    && !modifiers.contains(KeyModifiers::ALT)
+                {
                     // we need to do some parsing
                     character = match character_raw as u8 {
                         c @ b'\x01'..=b'\x1A' => (c as u8 - 0x1 + b'a') as char,


src/event/sys/windows/parse.rs
```diff
@@ -79,17 +79,21 @@ fn parse_key_event_record(key_event: &KeyEventRecord) -> Option<KeyEvent> {
             let character_raw = key_event.u_char;
 
             if character_raw < 255 {
```
TimonPost, line 81:
```suggestion
            if character_raw < 255 && character_raw != 0 {
```

src/event/sys/windows/parse.rs
```diff
@@ -79,17 +79,21 @@ fn parse_key_event_record(key_event: &KeyEventRecord) -> Option<KeyEvent> {
...
+                // Invalid character
+                if character_raw == 0 {
```
TimonPost, line 83: Remove this

src/event/sys/windows/parse.rs
```diff
@@ -79,17 +79,21 @@ fn parse_key_event_record(key_event: &KeyEventRecord) -> Option<KeyEvent> {
...
+                }
+
                 let mut character = character_raw as u8 as char;
 
                 if modifiers.contains(KeyModifiers::ALT) {
                     // If the ALT key is held down, pressing the A key produces ALT+A, which the system does not treat as a character at all, but rather as a system command.
                     // The pressed command is stored in `virtual_key_code`.
                     let command = key_event.virtual_key_code as u8 as char;
 
+                    // TODO: figure out if this is needed
```
TimonPost, line 94: I tested it out, if I press the numbers + ALT the ascii number for all characters pressed along ALT is placed in the `character_raw`, this does not happen with Numpad numbers. 

The reason why there is an alphabetic check is that windows interpret ALT as a system key and will place the pressed character in `key_event.virtual_key_code`. This will only be done for alphabetic characters. However, as discovered above, we can just as well use `character` for this. 

Here we overwrite the 'character' with the system keycode which is only an alphabetic character. 
https://github.com/crossterm-rs/crossterm/pull/488/files#diff-ff9f93aac33f0c41012d6df50b082828R92

I propose to remove this line which will give us both numeric and alphabetic support. 

One major change is that the `virtual key code` is capitalized by default while `character` isn't. Perhaps we can call `to_upper`.

TimonPost: I just realized this entire if statement line `89`-`97` are useless. We can remove this.

sigmaSd: I agree.

src/event/sys/windows/parse.rs
sigmaSd: This is not equivalent. with this change if character_raw ==0 it will still return Some(Keycode)

sigmaSd:
I modified this condition 
```rust if modifiers.contains(KeyModifiers::CONTROL)
                    && !modifiers.contains(KeyModifiers::ALT
```
to keep the old behavior (without it AltGr key events wont be recognized)

sigmaSd:
The current behavior of crossterm on windows is to not allow (ctrl + alt) key  combination. It might be on purpose but if its not it might be possible to fix it by removing https://github.com/crossterm-rs/crossterm/blob/master/src/event/sys/windows/parse.rs#L99 after this patch, but I can't tell whats the implication of that since it might be an intended behavior.

TimonPost (approved):
Seems to work perfect 👍

PR #489, opened 2020-09-17
Expose supports_ansi function

sigmaSd:
This is a useful function, I can only argue about my own use case though:

Calling , from rust code, binaries that have a `--color` flag, its useful to know if the terminal supports ansi or no, to disable the flag if it doesn't (or even handling it by parsing  ansi to winapi) , since most of the time its ansi coloring (cargo for example)

### DIFF
--- a/src/ansi_support.rs
+++ b/src/ansi_support.rs
@@ -50,6 +50,7 @@ lazy_static! {
     };
 }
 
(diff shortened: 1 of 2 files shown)


TimonPost:
Thanks, not objections

PR #496, opened 2020-10-09
Improve formatting for color parsing docs

sharnoff:
Happened to come across this today - just another little change for you. The lists of examples here didn't appear the same way when rendered, and I figured it's better to wrap things like `None` with backticks for inline code.


TimonPost (approved):
Epic!

PR #497, opened 2020-10-10
Fix panic enabling ANSI support when stdout is redirected

### LINKED ISSUE #444: Windows: "Getting mode failed" when stdout is redirected
**Describe the bug**
When stdout is redirected, or a program is invoked in $( ), crossterm prints "Getting mode failed" on the first execute!, and does not work properly (the screen is not cleared, a text is not shown or laid out, etc.).

**To Reproduce**
On Windows, the following program prints "Getting mode failed":
```rust
use crossterm::{cursor::Show, execute};
use std::io::{stderr, Write};

fn main() {
    execute!(stderr(), Show).unwrap();
    // stdout or stderr, and which command to execute does not matter
}
```

when invoked with (on cmd, PowerShell):
```sh
cargo run > foo
```

also with (on PowerShell):
```sh
$(cargo run)
```

but NOT with (on cmd, PowerShell):

```sh
cargo run
```

I tested it on Ubuntu and macOS with gnome-terminal and hyper, but the problem is specific to Windows.

**Expected behavior**
It works properly when stdout is redirected.

**OS**
- Windows

**Terminal/Console**
- ConHost, Windows Terminal
- cmd, Powershell

dsherret:
Fixes #444

Uses "CONIN$" instead as pointed out here https://stackoverflow.com/questions/33476316/win32-getconsolemode-error-code-6

OS error was that the handle was invalid.

### DIFF
--- a/src/ansi_support.rs
+++ b/src/ansi_support.rs
@@ -1,4 +1,4 @@
-use crossterm_winapi::ConsoleMode;
+use crossterm_winapi::{ConsoleMode, Handle};
 use winapi::um::wincon::ENABLE_VIRTUAL_TERMINAL_PROCESSING;
 
 use lazy_static::lazy_static;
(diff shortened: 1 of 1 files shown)


TimonPost:
Amazing man!

Issue #500, opened 2020-10-23
/dev/tty does not work on macOS with kqueue


acidghost:
**Describe the bug**
As described in tokio-rs/mio#1377, `kqueue` does not play well with `/dev/tty`. This is a problem for crossterm (as in #407), being unable to work properly.

**To Reproduce**
See reproducer in tokio-rs/mio#1377.

**Expected behavior**
Do not fail to poll on `/dev/tty`.

**OS**
MacOs.

**Terminal/Console**
 Any.

**Possible solutions**
The general solution is to use `select` instead of `kqueue` or `epoll` (`poll` should not work either). Changing Mio to support a `select`-based `Poll` requires major refactoring that is probably never going to happen. Given that `mio::Poll` is used in crossterm only (AFAIK) to poll `SIGWINCH` and `/dev/tty`, I suggest dropping the dependency on Mio and implement a `select`-based polling system with `libc` or `nix`.

Two major problems:
- how to poll from fd and signal on non-linux (w/o `signalfd`);
- how does it play with the waker system?

I'd be open to make a PR. Any suggestion?

acidghost:
To handle signals we could use the `signal_hook` crate with pipes and `select` on `/dev/tty` and the read end of the pipe.

I've never used Rust asynchronous programming features (i.e. `async/await`) so I'm a bit lost for the `event-stream` feature.

acidghost:
Here is some relevant code that uses this approach:

https://github.com/wez/wezterm/blob/f12df0be9b10491bdfdead3c4fc57918b208068d/termwiz/src/terminal/unix.rs#L390-L479

https://github.com/wez/wezterm/blob/f12df0be9b10491bdfdead3c4fc57918b208068d/filedescriptor/src/unix.rs#L421-L476

TimonPost:
- As for async and input system, crossterm has to functions: poll/read. If those work the (async) input system works for the most part. The thing that brings in some inconvenience is the `Waker` which is able to trigger the MIO listener and wake it up.
- Crossterm already uses the signal crate with the mio extension for [terminal size](https://github.com/crossterm-rs/crossterm/blob/master/src/event/source/unix.rs#L129).

There are some options here. First, given that this usecase of `/dev/tty` only doesnt work on macos we could limit this feature and make macos use STDOUT by default. Which might bring some sad faces along. 

[...]

PR #501, opened 2020-10-24
Prefer /dev/tty over stdin in terminal functions

acidghost:
This is part of #500.

### DIFF
--- a/src/terminal/sys/unix.rs
+++ b/src/terminal/sys/unix.rs
@@ -1,16 +1,16 @@
 //! UNIX related logic for terminal manipulation.
 use std::{io, mem, process, sync::Mutex};
 
-use crate::event::sys::unix::file_descriptor::FileDesc;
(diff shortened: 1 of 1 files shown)


TimonPost (approved):
One quick thing in case you were not aware of this, the `tty_fd` does not yet use prefer /dev/tty over STDIN. 

https://github.com/crossterm-rs/crossterm/blob/5c12c9cd0003be224a4148b6f9d92e55876ea90f/src/event/sys/unix/file_descriptor.rs#L67

But this is a good thing for now, until the MacOs issue is resolved we should not default it to /dev/tty. However, with this pull request its just a matter of changing the dev/tty at a later stage. 

Is that good with you? if so ill merge it.

acidghost:
Select on macOS supports both stdin and `/dev/tty`, the `isatty` should be enough to check which of the two to use.

Anyhow, you can merge it as is and I’ll do some testing when `select` implementation is in.

PR #502, opened 2020-10-27
Add support for any event tracking

Koxiaet:
This adds support for any-event tracking, where all mouse movements are reported even if no buttons are held down. I used Xterm's [ctlseqs](https://invisible-island.net/xterm/ctlseqs/ctlseqs.html#h2-Mouse-Tracking) as reference, and I have tested this on URXVT, xterm, st and alacritty in both normal mode and SGR mode, and it seems to work fine.

### DIFF
--- a/src/event/sys/unix/parse.rs
+++ b/src/event/sys/unix/parse.rs
@@ -1,5 +1,5 @@
 use crate::{
-    event::{Event, KeyCode, KeyEvent, KeyModifiers, MouseButton, MouseEvent},
+    event::{Event, KeyCode, KeyEvent, KeyModifiers, MouseButton, MouseEvent, MouseEventKind},
     ErrorKind, Result,
 };
 
@@ -141,8 +141,8 @@ pub(crate) fn parse_csi(buffer: &[u8]) -> Result<Option<InternalEvent>> {
             code: KeyCode::BackTab,
             modifiers: KeyModifiers::SHIFT,
         })),
-        b'M' => return parse_csi_x10_mouse(buffer),
-        b'<' => return parse_csi_xterm_mouse(buffer),
+        b'M' => return parse_csi_normal_mouse(buffer),
+        b'<' => return parse_csi_sgr_mouse(buffer),
         b'0'..=b'9' => {
             // Numbered escape code.
             if buffer.len() == 3 {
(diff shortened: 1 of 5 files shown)


TimonPost:
Thats epic! I will do the review later.

src/event/sys/unix/parse.rs
```diff
@@ -397,58 +342,77 @@ pub(crate) fn parse_csi_xterm_mouse(buffer: &[u8]) -> Result<Option<InternalEven
...
+    let cb = next_parsed::<u8>(&mut split)?;
+    let (kind, modifiers) = parse_cb(cb)?;
 
     // See http://www.xfree86.org/current/ctlseqs.html#Mouse%20Tracking
     // The upper left character position on the terminal is denoted as 1,1.
     // Subtract 1 to keep it synced with cursor
     let cx = next_parsed::<u16>(&mut split)? - 1;
     let cy = next_parsed::<u16>(&mut split)? - 1;
 
+    let kind = if buffer.last() == Some(&b'm') {
```
TimonPost, line 354: Can you add a comment on what this statement is about?

src/event/sys/unix/parse.rs
```diff
@@ -397,58 +342,77 @@ pub(crate) fn parse_csi_xterm_mouse(buffer: &[u8]) -> Result<Option<InternalEven
...
+        match kind {
+            MouseEventKind::Down(button) => MouseEventKind::Up(button),
```
TimonPost, line 356: especially this one

TimonPost (commented):
Nice changes, I like the change with MouseEventKind and the coordinates being decoupled. It is a big change in the API so I will have to do a major increment. I tested it on windows and it seems to work perfectly. I will have to give it a test on Linux soon if it work perfect. I think we are good to go here. Have you tested on macos already?

TimonPost:
Thanks. I tested it on Linux and it works. Let's merge this one :+1: . I tried to fire up a mac terminal over SSH with this project https://github.com/fastai/fastmac. But SSH doesn't seem to record my mouse events on the remote machine which makes it not possible to test this behavior. Can we assume it to work? Or should we better ask around some people to test this branch on a real mac system?

Koxiaet:
We might as well ask on the discord - I'll do that now.

octotep:
I’ll copy and paste my comment from discord in case people don’t check there as frequently - hope it’s not considered spammy.

I ran a test with this branch on a Mac using iTerm2 and witnessed: 

- Moving the mouse without any buttons pressed produced MouseEvents with the kind Moved
- Button presses were detected
- Drags were detected fine, with the correct button in the kind
- Button releases were detected
- ScrollUp and ScrollDown were detected

Looks great on Mac, well done :)

TimonPost:
Cool, I merge this PR

PR #503, opened 2020-10-27
Use a thread for event stream

### LINKED ISSUE #448: `EventStream` spawns a new thread on every event
**Describe the bug**
Using `EventStream` seems to spawn a new thread on every event.

**To Reproduce**
The behaviour can be reproduced with the `event-stream-tokio.rs` example. Notice the new PID in `htop` in the gif below which changes on every keypress.

<details><summary>How to reproduce [gif]</summary>
<p>

Peek 2020-06-29 20-47

N.B. In the gif above, `core_threads` is set to one for clarity (i.e. `#[tokio::main(core_threads = 1)]`).
</p>
</details>

**Expected behavior**
A new thread should not be created on every keystroke. 

**OS**
Ubuntu 20.04 LTS (Linux 5.4.0-37-generic x86_64)

**Terminal/Console**
Alacritty + tmux.

---

After investigating this, I thought it may be related to `tput` because of the following comment in the source code

> This can take a really long time, because terminal::size can launch new process (tput) and then it parses its output.

However, this doesn't seem to be the case -- the branch where tput gets called (when `/dev/tty` is not available) is not taken and at any rate removing the signal handling logic for `signal_hook::SIGWINCH` doesn't change the behaviour.

quininer:
Close #448

### DIFF
--- a/src/event/stream.rs
+++ b/src/event/stream.rs
@@ -14,8 +14,6 @@ use futures_core::{
     task::{Context, Poll},
 };
 
-use lazy_static::lazy_static;
-
 use crate::Result;
 
 use super::{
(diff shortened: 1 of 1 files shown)


TimonPost:
Thanks for this improvement! I like the Task-based approach. I tested the branch and everything seems to behave as it is supposed to do. Lets merge this.

PR #506, opened 2020-10-30
Added some control code to KeyEvent mappings

fabi321:


TimonPost:
I tried this branch but it gives me the wrong results. When pressing `Backspace` only or `Backspace with CTRL` I get only Backspace. When I change the code to be:

```
 b'\x7F' => Ok(Some(InternalEvent::Event(Event::Key(KeyEvent::new(
            KeyCode::Backspace,
            KeyModifiers::NONE,
        ))))),
        // Backspace control code
        b'\x08' => Ok(Some(InternalEvent::Event(Event::Key(KeyEvent::new(
            KeyCode::Backspace,
            KeyModifiers::CONTROL,
        ))))),
```

I do get the correct results. x08 seems to be the CTRL + Backspace while /x7F seems to be the normal backspace.

fabi321:
That is indeed what you get on the gnome console or Konsole. But in ASCII, 0x08 or ^H is defined as backspace. i.e. if you go to a shell and press ^H or ctrl + backspace, you delete only the last char. if you want to delete the last word on a shell, you need to press ^W. 0x7F is delete according to ASCII.

fabi321:
[Gnome console bug](https://bugzilla.gnome.org/show_bug.cgi?id=733246) and [Konsole bug](https://bugs.kde.org/show_bug.cgi?id=391075) so it seems like all major Unix terminals have this behavior

PR #511, opened 2020-11-11
Fix InternalEventReader::poll timeout computation

carl-erwin:
**Describe the bug**

The poll function does not respect the passed Duration.
Sometime the polling time is doubled in **InternalEventReader::poll** when
**event_source.try_read** returns **None** and **poll_timeout.elapsed()** returns **false**

**poll_timeout.leftover()** must be used insted of **timeout**

**To Reproduce**

1. run the event poll example
  cargo run --example  event-poll-read
  
2. let the program run without doing anything

**Expected behavior**
The '.' characters should be displayed every seconds

**OS**
 - Linux Ubuntu 20.10

**Terminal/Console**
 - Konsole

### DIFF
--- a/src/event/read.rs
+++ b/src/event/read.rs
@@ -65,7 +65,7 @@ impl InternalEventReader {
         let poll_timeout = PollTimeout::new(timeout);
 
         loop {
-            let maybe_event = match event_source.try_read(timeout) {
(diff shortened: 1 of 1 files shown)


TimonPost (approved):
Mmm, thanks for finding it!

Issue #515, opened 2020-11-22
Supporting multiple modifier key recognition (like Ctrl+Shift)


curlpipe:
**Is your feature request related to a problem? Please describe.**
This is a great library, probably one of the most complete libraries I have found. The problem I'm finding is that the users of my program (that uses Crossterm) wish to be able to use keyboard shortcuts such as Ctrl+Shift and Alt+Shift etc... When I try implementing these modifiers, it just registers as usual Ctrl and Alt keys.

**Describe the solution you'd like**
I'd like to be able to bind multiple modifier keys on top of each other. E.g. `Ctrl + Shift + S` or `Ctrl + Alt + R` etc...

**Describe alternatives you've considered in any**
Not really sure how to get around this, I could maybe write some janky code to work around this using another library or using libc for now but I don't want to go into that territory.

Many thanks :)

vlisivka:
Example key codes sent by gnome-terminal:

Up: \e[A
Shift Up: \e[1;2A
Alt Up: \e[1;3A
Shift+Alt Up: \e[1;4A
Control Up: \e[1;5A
Control+Shift Left: \e[1;6D

pianohacker:
Can this be closed with the addition of the [keyboard enhancement feature](https://docs.rs/crossterm/latest/crossterm/event/struct.PushKeyboardEnhancementFlags.html) to 0.25?

kyoheiu:
Is this implemented? In my app I tried to catch `Ctrl + Shift + c` (`Ctrl + C`) but couldn't.
FYI, changing the key from `c` to e.g. 'z' does not work either.
If I miss something, please let me know.

LevitatingBusinessMan:
This can be done using the [KeyboardEnhancementFlags](https://docs.rs/crossterm/latest/crossterm/event/struct.KeyboardEnhancementFlags.html) if you have a terminal that supports it.

As of two weeks, it is supported by Alacritty: https://github.com/alacritty/alacritty/commit/cb03806e2ab85674c45e87e1bb24dfe2fd1a918c

PR #518, opened 2020-12-02
Add support for combining some modifier keys

Kartoffelsaft:
Partially helps with #515 but doesn't completely solve the issue, as it only really helps with the alt key and only on Unix systems. Getting ctrl+shift might be difficult though, as from my testing combining those keys outputs the same values to the buffer as just ctrl.


TimonPost:
Is anyone able to validate this? I do not have a Linux distro to test it on quickly. @Kartoffelsaft maybe if you post a screenshot that it works we can get this merged. Also what is your definition of 'partial'?

TimonPost:
I validated that it made CTRL + ALT key working

PR #522, opened 2020-12-19
Simplify ansi_support

Koxiaet:
This greatly simplifies the `ansi_support` module.

- Instead of having a function to toggle vt processing it can only enable it, since that's all we do in Crossterm.
- Checking for a specific term is completely redundant since all that matters is that `$TERM` isn't `dumb`.
- It isn't necessary to attempt to enable ANSI support in terminal.rs since it's done automatically whenever a command is called.

### DIFF
--- a/src/ansi_support.rs
+++ b/src/ansi_support.rs
@@ -5,29 +5,23 @@ use lazy_static::lazy_static;
 
 use crate::Result;
 
-/// Toggle virtual terminal processing.
+/// Enable virtual terminal processing.
 ///
-/// This method attempts to toggle virtual terminal processing for this
-/// console. If there was a problem toggling it, then an error returned.
-/// On success, the caller may assume that toggling it was successful.
+/// This method attempts to enable virtual terminal processing for this
+/// console. If there was a problem enabling it, then an error returned.
+/// On success, the caller may assume that enabling it was successful.
 ///
 /// When virtual terminal processing is enabled, characters emitted to the
 /// console are parsed for VT100 and similar control character sequences
 /// that control color and other similar operations.
-pub(crate) fn set_virtual_terminal_processing(yes: bool) -> Result<()> {
+fn enable_vt_processing() -> Result<()> {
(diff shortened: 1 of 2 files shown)


TimonPost:
Where is the TERM environment variable coming from?

Koxiaet:
@TimonPost I don't know really, the code just does the exact same thing as it used to. It was [your commit](https://github.com/crossterm-rs/crossterm-utils/commit/ad74f6b5242e6ed998bef6240b738785ea449ffe#diff-e16ffaffd5cd62e75b8757032ddc16f48122eea91df28aceaf12d683a85d0a2bR90) that introduced `TERM`-checking.

src/ansi_support.rs
```diff
@@ -5,76 +5,35 @@ use lazy_static::lazy_static;
...
+    if old_mode & mask == 0 {
+        console_mode.set_mode(old_mode | mask)?;
     }
 
     Ok(())
 }
 
 lazy_static! {
     static ref SUPPORTS_ANSI_ESCAPE_CODES: bool = {
-        // Some terminals on windows like GitBash can't use WinaApi calls directly
```
TimonPost, line 38: I would like to keep this comment.

TimonPost (commented):
Ah I see why I did that. The comment above it sated:
```
// Some terminals on windows like GitBash can't use WinaApi calls directly
// so when we try to enable the ANSI-flag for windows this won't work.
// Because of that we should check first if the TERM-variable is set
// and see if the current terminal is a terminal who does support ANSI.
```

I like the change it simplifies the code. I think in the past I used to use the boolean parameter to disable this terminal flag when the program closes.

src/ansi_support.rs
Koxiaet: OK, it's been restored. I wasn't really thinking when I removed it :sweat_smile:.

TimonPost (approved):
Great, thanks!

PR #523, opened 2020-12-20
Replace AnsiType with write_ansi

Koxiaet:
The main change is changing the `Command` trait from:
```rust
trait Command {
    type AnsiType: Display;

    fn ansi_code(&self) -> Self::AnsiType;

    #[cfg(windows)]
    fn execute_winapi(&self, writer: impl FnMut() -> Result<()>) -> Result<()>;
    #[cfg(windows)]
    fn is_ansi_code_supported(&self) -> bool { ... }
}
```
to:
```rust
trait Command {
    fn write_ansi(&self, f: &mut impl fmt::Write) -> fmt::Result;

    #[cfg(windows)]
    fn execute_winapi(&self, writer: impl FnMut() -> Result<()>) -> Result<()>;
    #[cfg(windows)]
    fn is_ansi_code_supported(&self) -> bool { ... }
}
```

This avoids cloning in `Print`-like commands, and avoids allocating `String`s in other commands, as well as simplifying the `Command` trait.

I also had to rework command execution to do the actual logic in the `QueueableCommand` and `ExecutableCommand` traits since it's easier to write code in functions than macros. The macros are still there, they just call the functions in `command.rs`.

### DIFF
--- a/src/command.rs
+++ b/src/command.rs
@@ -1,6 +1,5 @@
-use std::{fmt::Display, io::Write};
-
-use crate::{execute, queue};
+use std::fmt;
+use std::io::{self, Write};
 
 use super::error::Result;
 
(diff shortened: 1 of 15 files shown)


src/command.rs
```diff
@@ -179,8 +176,56 @@ where
...
         Ok(self)
     }
 }
+
+/// Writes the ANSI representation of a command to the given writer.
+fn write_command_ansi<C: Command>(
+    io: &mut (impl io::Write + ?Sized),
+    command: C,
+) -> io::Result<()> {
+    struct Adapter<T> {
```
TimonPost, line 191: Im not sure why we need this adaptor? Can we not write directly to 'io'?

TimonPost (commented):
- I like the return type to writer change. One negative thing about this is that a user is not able to use the static ANSI string references anymore. 
- I like the change that the macro calls the static code.

src/command.rs
Koxiaet: The commands write to a `fmt::Write` type, whereas this is an `io::Write` type. The [standard library](https://github.com/rust-lang/rust/blob/master/library/std/src/io/mod.rs#L1499) has a similar adaptor. The alternative is to have an adaptor that wraps a `Command` and implements `Display`, however that has to create and go through a `Formatter`, while this writes it directly.

TimonPost: Okey perfect

PR #524, opened 2020-12-20
Make imports & docs more consistent

Koxiaet:
This PR makes the import ordering more consistent, and makes all the comments use `WinAPI` instead of `winapi`/`Winapi`/`WinApi`.


TimonPost (commented):
Nice changes

TimonPost:
I tried to resolve conflicts on other branches but see it resulted in an error. Are you able to fix that? Then we can merge this branch and do a release.

PR #528, opened 2020-12-29
Support taking any Display in SetTitle

Koxiaet:
In terms of performance it should be the same - there are no extra allocations or anything.


TimonPost (approved):
Nice! Its a breaking change so we will have to go to 0.20

PR #529, opened 2020-12-29
Inline ansi modules

Koxiaet:
Having separate modules that store the ANSI codes makes the codebase harder to read, as there is a layer of indirection. If they are inline, it reduces the number of lines of code, improves the usefullness of the `[src]` button in the documentation, and reduces the amount of jumping around required when reading or editing the code.


src/cursor.rs
```diff
@@ -64,7 +63,7 @@ pub struct MoveTo(pub u16, pub u16);
 
 impl Command for MoveTo {
     fn write_ansi(&self, f: &mut impl fmt::Write) -> fmt::Result {
-        ansi::move_to_csi_sequence(f, self.0, self.1)
+        write!(f, csi!("{};{}H"), self.0 + 1, self.1 + 1)
```
TimonPost, line 66: When debugging I discoverd those arguments should be switched around. 

```rs
write!(f, csi!("{};{}H"), self.1 + 1, self.0 + 1)
```

TimonPost (changes requested):
Nice changes it does really make it easier to read. During debugging everything still seems to work as supposed except for moving the cursor position.

Koxiaet:
Ah, that's subtle. It should be fixed now.

TimonPost:
Thanks!

PR #530, opened 2020-12-29
Remove lazy_static dependency

Koxiaet:
This replaces the `lazy_static` dependency with either atomics or `parking_lot`'s const mutexes.


TimonPost (commented):
Great Change I wasn't aware of that constant mutex, that's a nice feature. Have you tried running the demo-application? There are many fundamental changes here. So I would like to test it with that.  I might give it a shot later.

PR #532, opened 2021-01-01
Add movetorow command

sigmaSd:


src/cursor.rs
```diff
@@ -132,6 +132,25 @@ impl Command for MoveToColumn {
...
+///
+/// # Notes
+///
+/// Commands must be executed/queued for execution otherwise they do nothing.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct MoveToRow(pub u16);
+
+impl Command for MoveToRow {
+    fn write_ansi(&self, f: &mut impl fmt::Write) -> fmt::Result {
+        ansi::move_to_row_csi_sequence(f, self.0)
```
TimonPost, line 145: In light of #529 I would prefer that this PR also moves the ANSI code into this function such that we will not have to open an other PR in the future.

TimonPost (approved):

TimonPost:
I merged #529 this PR has to be updated such that the ANSI code is inlined in the 'write_ansi' function.

src/cursor.rs
sigmaSd: done

TimonPost (approved):

PR #541, opened 2021-02-20
KeyEvent case normalizing equality

### LINKED ISSUE #517: KeyEvent's KeyCode's case could be normalized to be less redundant
It is a little frustrating when comparing something like 
```
KeyEvent { KeyCode::Char('a'), KeyModifiers::SHIFT } == KeyEvent { KeyCode::Char('A'), KeyModifiers::SHIFT }
```
and getting false, when chances are these are the same. This also leads to issues where if a lowercase keycode is combined with the shift key modifier (or an uppercase without shift), it will always be unequal to the output of something like read().

When deriving Eq/PartialEq it could convert to the same casing for both the left and right hand side before comparison.

The casing for KeyCode could be normalized at construction. This was my first idea, but it is possible some people might be pulling the character out of the keycode with the assumption it will preserve casing, and this would be a breaking change.

**Additional context**
Came across this when writing tests for combining key modifiers (`KeyModifiers::SHIFT | KeyModifiers::ALT` type thing) and found it unusual that I had to specify casing both as a modifier and in the char.

necrosovereign:
Manually implements `Eq`, `PartialEq` and `Hash` for `KeyEvent`

Adds:
* Manual implementation of `PartialEq` for `KeyEvent` with the property that lowercase ASCII characters are equal to uppercase in the presence of SHIFT modifiers and modifiers with SHIFT are equal to modifiers without SHIFT in the presence of an uppercase character. For example:
```rust
KeyEvent { code: Char( 'd',), modifiers: SHIFT } == KeyEvent { code: Char( 'D',), modifiers: SHIFT }
KeyEvent { code: Char( 'D',), modifiers: SHIFT } == KeyEvent { code: Char( 'D',), modifiers: NONE }
```
* Manual implementation of `Hash` for `KeyEvent` to preserve the property:
```
k1 == k2 -> hash(k1) == hash(k2)
```
* Private method `KeyEvent::normalize_case` that assists in implementation of `PartialEq` and `Hash`.
* Tests for the intended behavior of `Eq` and `Hash` on `KeyEvent`.

Closes #517.

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -466,3 +466,41 @@ pub(crate) enum InternalEvent {
     #[cfg(unix)]
     CursorPosition(u16, u16),
 }
+
+#[cfg(test)]
+mod tests {
+    use std::collections::hash_map::DefaultHasher;
+    use std::hash::{Hash, Hasher};
+
+    use super::{KeyCode, KeyEvent, KeyModifiers};
+
+    #[test]
+    fn test_equality() {
+        let lowercase_d_with_shift = KeyEvent::new(KeyCode::Char('d'), KeyModifiers::SHIFT);
+        let uppercase_d_with_shift = KeyEvent::new(KeyCode::Char('D'), KeyModifiers::SHIFT);
+        let uppercase_d = KeyEvent::new(KeyCode::Char('D'), KeyModifiers::NONE);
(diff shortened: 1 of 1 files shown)


TimonPost (approved):
Thanks, good improvement!

PR #548, opened 2021-03-26
Added Cursor Shape Support

cowboy8625:
So as far as I can tell this works on linux and windows.  I have no way to test on Mac.   I also didn't update Readme and other things. I can if you want me too.

### DIFF
--- a/src/cursor.rs
+++ b/src/cursor.rs
@@ -360,6 +360,42 @@ impl Command for DisableBlinking {
     }
 }
 
+/// All supported cursor shapes
+///
+/// # Note
+///
+/// - Used with SetCursorShape
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum CursorShape {
+    UnderScore,
+    Line,
+    Block,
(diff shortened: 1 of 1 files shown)


cowboy8625:
its failing for things I did not submit.

src/cursor.rs
```diff
@@ -360,6 +360,42 @@ impl Command for DisableBlinking {
...
+        match self.0 {
+            UnderScore => f.write_str("\x1b[3 q"),
+            Line => f.write_str("\x1b[5 q"),
+            Block => f.write_str("\x1b[2 q"),
+        }
+    }
+
+    #[cfg(windows)]
+    fn execute_winapi(&self, _writer: impl FnMut() -> Result<()>) -> Result<()> {
+        Ok(())
```
TimonPost, line 395: I think this behavior is possible to replicate with windows API. Tho it is not mandatory to me.

TimonPost (approved):
Thanks for the work, I will update the readme when releasing.

PR #551, opened 2021-03-27
refactor: fix clippy warnings

SirWindfield:
All of the current PRs fail CI due to clippy failing on the main branch.
I fixed the errors that got returned. I had to change some internal
method signatures as some methods returned `Result<Option<T>>` but never
returned an error value, making the `Result` wrapping  unnecessary.


TimonPost (approved):
Nice, thanks!

PR #552, opened 2021-03-27
refactor: remove unneeded imports

SirWindfield:
Some warnings show up in the CI logs (click on the `Files changed` tab of an open PR, for example #551, and scrolling down a little bit).

### DIFF
--- a/src/event/stream.rs
+++ b/src/event/stream.rs
@@ -113,7 +113,9 @@ impl Stream for EventStream {
             Ok(false) => {
                 if !self
                     .stream_wake_task_executed
-                    .compare_and_swap(false, true, Ordering::SeqCst)
+                    .compare_exchange(false, true, Ordering::SeqCst, Ordering::SeqCst)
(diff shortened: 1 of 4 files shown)


src/event/stream.rs
```diff
@@ -113,7 +113,9 @@ impl Stream for EventStream {
             Ok(false) => {
                 if !self
```
SirWindfield, line 114: I am not entirely sure about the semantics of this code snippet. `compare_and_swap` (per docs): `The return value is always the previous value. If it is equal to current, then the value was updated`. 

So this checks that the previous value was false, am I right? In that case, maybe changing the below code to

```rust
if let Ok(value) = self.stream_wake_task_executed.compare_exchange(false, true, Ordering::SeqCst, Ordering::SeqCst) {
	if !value {
		//...
	}
}
```

Might show the intention of the code a little bit better.

TimonPost: Perfect to me.

TimonPost (approved):
Nice

PR #553, opened 2021-03-27
Use io::Error for public error type

Plecra:
This makes interoperation with the standard library much easier for users, which is especially useful given crossterm's low level APIs.

A custom error type is only useful if the context it provides outweighs its ergonomic burden, and in this case only the `ResizingTerminalFailure` and `SettingTerminalTitleFailure` variants were being used.

Do we have examples of these variants being used in dependent code? I can add them to the new representation if so.


TimonPost:
I agree, the error types we have currently do not really add that much value. Thanks for the pull request. Give it a look in a bit.

TimonPost (approved):

Issue #545, opened 2021-03-11
Bracketed Paste Events


bhgomes:
**Is your feature request related to a problem? Please describe.**
I want to be able to process input that contains line feeds. Allowing the terminal to enter [bracketed paste mode](https://cirw.in/blog/bracketed-paste) allows us to buffer an incoming paste and collect non-`Enter` line feeds. This is already a standard feature of many terminal emulators. 

**Describe the solution you'd like**
- Add the relevant enable/disable pair: `enable_bracketed_paste` (`EnableBracketedPaste`) and `disable_bracketed_paste` (`DisableBracketedPaste`), either to global scope or to the terminal depending on the unix/windows implementation details.
- Here are a few options for formatting the paste:
  - [1] Add `Event::Paste` variant that holds the entire pasted string
  - [2] Add `KeyCode::BracketedPasteStart` and `KeyCode::BracketedPasteEnd` variants and allow the user to access escape codes so that they can implement paste filtering themselves (not very user-friendly) or another part of this library will do the paste filtering.
- Because both of these options change the enum interfaces, we can probably put this behind an optional feature.

**Describe alternatives you've considered in any**
An alternative is to use something like [`rustyline`](https://github.com/kkawakam/rustyline) to handle shell-like input, and only use `crossterm` to handle other kinds of interaction. The problem with something like [`rustyline`] is that it [only allows stdin/stdout](https://github.com/kkawakam/rustyline/issues/377) as far as I am aware and won't be able to implement something like the following "half-shell":

```
-----------------------------
|             |             |
|  [shell     |             |
|  output]    |   [other    |
|             |   widget]   |
|-------------|             |
| > [prompt]  |             |
-----------------------------
```

where the `[other widget]` can read arrow key and other non-text inputs (i.e. how is [`rustyline`] supposed to know that the input should wrap only at half the terminal width?). Also, multi-panel input components like [`tui-rs`](https://github.com/fdehau/tui-rs) widgets. See [this issue](https://github.com/kkawakam/rustyline/issues/494) for other compatibility issues with [`rustyline`](https://github.com/kkawakam/rustyline).

**Additional context**
- The [`termwiz`](https://crates.io/crates/termwiz) crate has a similar kind of [`InputEvent`](https://docs.rs/termwiz/0.11.0/termwiz/input/enum.InputEvent.html) as solution [1].
[...]

PR #557, opened 2021-04-01
[draft] Add Bracketed Paste Events

sigmaSd:
https://github.com/crossterm-rs/crossterm/issues/545

Couple of points:

- I can flatten  `PasteEvents(Vec<Event>)` to become `PasteEvent(String)`
- I'm not sure how to restore `Copy` bound to `Event`
- I think for winapi I should just stub it with Ok(())
- Is flushing events like this https://github.com/crossterm-rs/crossterm/compare/master...sigmaSd:add_brakceted_paste_mode?expand=1#diff-835dd5a0632e3b0d7b10539464274511152432fac2dce69a999604c76c86f0f9R128 Ok?

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -295,7 +295,7 @@ impl Command for DisableMouseCapture {
 
 /// Represents an event.
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
-#[derive(Debug, PartialOrd, PartialEq, Eq, Clone, Copy, Hash)]
+#[derive(Debug, PartialOrd, PartialEq, Eq, Clone, Hash)]
 pub enum Event {
     /// A single key event with additional pressed modifiers.
     Key(KeyEvent),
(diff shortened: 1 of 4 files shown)


sigmaSd:
To not flush the events I need to detect BracketdPasteStart/End here https://github.com/crossterm-rs/crossterm/blob/master/src/event/source/unix.rs#L76 but I think they're being filtered? I would appreciate some pointers there.

The idea if i detect them at that point I can 
- Add to `UnixInternalEventSource` struct, a `pate_flag` and `paste_buffer`
- If the event at line 76 is `BracketedPasteStart` set `paste_flag` to `true`
- if `paste_flag` is `true`, push the received event at line 76 to `paste_buffer` and don't early return
- If the event at line 76 is `BracketedPasteEnd` set `paste_flag` to `false`, `return Event(PasteEvent(self.paste_buffer.drain())`

sigmaSd:
closed in favor of https://github.com/crossterm-rs/crossterm/pull/693

PR #558, opened 2021-04-16
Update to signal-hook 0.3.8, signal-hook-mio 0.2.1

erickt:
Note, the documentation for creating an iterator from `&signals` suggests
it returns an infinite stream. I instead switched to `pending()`, since
presumably we don't want to block forever waiting for signals.


TimonPost (approved):
Great!

PR #566, opened 2021-05-06
Fix building with the event-stream feature

sigmaSd:


TimonPost (approved):
Nice, was just about to make a release ready for 0.20

PR #570, opened 2021-05-16
Remove writer parameter from execute_winapi

Kestrer:
The only command that used this parameter was `Print`, which we can force to call `write_command_ansi` with by always returning `true` from `is_ansi_code_supported`. This does mean that `Print` will no longer call `.flush()` on the writer when queued, but any bugs potentially caused by that can be mitigated by flushing before calling any WinAPI commands.


TimonPost:
Good change, yea flushing is perfectly fine, queuing with winapi didnt work anyways.

PR #571, opened 2021-05-16
Unify `Colorize` and `Styler` as `Stylize`

Kestrer:
`Colorize` and `Styler` were implemented on basically the exact same types, so I thought it would be easier if they used a single trait. I called it `Stylize` because the name `Styler` implies that the types implementing the trait are the ones doing the styling, when in fact it's the other way around. Naming traits after verbs is also fairly common (e.g. `Write`, `Clone`, `Borrow` etc).

A few other notes:
* I redesigned the trait itself to make it easier to implement and avoid the complex macro callback thing we had before. Instead of requiring implementations to implement all the methods themselves, they must simply provide access to a `ContentStyle`, and the trait can generate default implementations in terms of that.
* I moved `StyledContent::{with, on, attribute}` to the `Styled` trait to allow code like `"Hello!".on(Color::Black)`. Of course `"Hello!".on_black()` is better in that case but this method is nice to have. I'm not really sure about the naming here - `with`/`on` could also be called `foreground`/`background` like `ContentStyle` did, that might be more explicit.
* I implemented `Stylize` on `ContentStyle` to allow it to have all the convenience methods too. `ContentStyle::{foreground, background, attribute}` are no longer necessary as `Stylize::{with, on, attribute}` exist.
* I changed the implementation of `Display` on `StyledContent` to forward to `PrintStyledContent` to bring it in line with `Attribute` and to avoid printing useless ANSI escape sequences on older Windows.


TimonPost:
I like the changes you've made, thanks for the contribution. They will be released in 0.20 in some days along with the other changes

PR #579, opened 2021-06-30
Add types for ‘dimensions’

arzg:
This PR extracts types for both the size of the terminal window, as well as the cursor’s position. Previously the coordinate order for these was specified only in documentation, which has led to a number of bugs in my own code. If, as a user of crossterm, I have to spell out that I am accessing the cursor’s `column` or the window’s `width`, I am much less likely to make such a mistake.

I understand this change is subjective and might make code too wordy for the tastes of some, so feel free to close if you disagree.

I don’t have access to a Windows device, so I probably haven’t fixed all the errors in the Windows portion of this PR.

### DIFF
--- a/examples/event-read.rs
+++ b/examples/event-read.rs
@@ -31,7 +31,7 @@ fn print_events() -> Result<()> {
             println!("Cursor position: {:?}\r", position());
         }
 
-        if let Event::Resize { .. } = event {
+        if let Event::Resize(_) = event {
             let (original_size, new_size) = flush_resize_events(event);
             println!("Resize from: {:?}, to: {:?}", original_size, new_size);
         }
--- a/examples/event-read.rs
+++ b/examples/event-read.rs
@@ -9,7 +9,7 @@ use crossterm::{
     cursor::position,
     event::{read, DisableMouseCapture, EnableMouseCapture, Event, KeyCode},
     execute,
-    terminal::{disable_raw_mode, enable_raw_mode},
+    terminal::{disable_raw_mode, enable_raw_mode, TerminalSize},
     Result,
 };
 use std::time::Duration;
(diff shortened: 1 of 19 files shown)


TimonPost (commented):
Thanks for trying to contribute. While I think having this struct can create clarity for the API, I am not sure if I like the extra boilerplate it introduces. Many APIs I have seen all work with simple x, y float variables instead of asking to create an instance of a type. I am not so sure if I like adding that boilerplate to the API. Can you provide some arguments that could make me like it more?

arzg:
It definitely does add a lot of extra typing, which I’m also not a fan of. I guess once you know the order, you’re unlikely to make the same mistake again. Then again, I have 😆

As a middle ground, what would you think of `MoveTo` being

```rust
struct MoveTo {
    column: u16,
    row: u16,
}
```

thereby reducing typing while remaining unambiguous?

The terminal_size crate [uses an interesting approach](https://docs.rs/terminal_size/0.1.17/terminal_size/fn.terminal_size.html): rather than differentiating width and height using differently-named fields, they instead introduce separate types for width and height:

```rust
pub fn terminal_size() -> Option<(Width, Height)>;

pub struct Width(pub u16);
pub struct Height(pub u16);
```

I’m not sure whether this is a better or worse approach, but I think it’s worth considering.

TimonPost:
Then I would consider type defs instead of structs ...

arzg:
Although typedefs wouldn’t make mistakes any less likely, I think they are still an improvement over the current situation by lifting something mentioned only in the documentation to the code level. Because they’re typedefs, making a mistake isn’t an error, though.

TimonPost:
I think its a bit overhead imo. Even with those type devs or structs. It does make the API more specific but I think a comment is sufficient to let the user know how to pass in the arguments. Or just the parameter names.

arzg:
Fair enough. Thanks for taking the time to review this PR! :)

PR #581, opened 2021-07-04
Docs: Add some links related to raw mode

mgunyho:
Hi! I just started trying out crossterm and I think having these links make the docs a bit easier to navigate.

I didn't use `code` style in the links to the enable/disable functions, because I saw that it was not used elsewhere. Or is that incorrect?

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -15,7 +15,7 @@
 //! * use the [`read`](fn.read.html) & [`poll`](fn.poll.html) functions on any, but same, thread
 //! * or the [`EventStream`](struct.EventStream.html).
 //!
-//! **Make sure to enable raw mode in order for keyboard events to work properly**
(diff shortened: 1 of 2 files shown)


TimonPost:
What is code style in particullary? Do you mean that they become MD-styled?

mgunyho:
Sorry for the late reply. Yes, I mean that I did not use the backticks `` ` `` around the link. Now it looks like this:

crostterm-link-no-backtick

Typically, at least in the rust official documentation, links to functions/methods are written like `` [`method_name`] ``, (see for example [here](https://doc.rust-lang.org/std/option/enum.Option.html#method.unwrap)) so that it would look like this:

crostterm-link-backtick

I think this is a nice way of highlighting that this is a link to a function. But this style was not used elsewhere in the documentation, so maybe it's better to not use it here for consistency.

TimonPost:
Its good like this in the future the library can be updated to use the other syntax.

Issue #573, opened 2021-05-26
Incorrect color names in docs


tobx:
**Describe the bug**

In the docs at [Platform-specific Notes](https://docs.rs/crossterm/0.19.0/crossterm/style/enum.Color.html#platform-specific-notes) it says `DarkWhite` for the dark version of `White` and `Grey` for the light version of `Black`.

**Expected behavior**

The dark version of `White` should be `Grey` while the light version of `Black` should be `DarkGrey`. `DarkWhite` does not exist. At least this is how it is coded [here](https://github.com/crossterm-rs/crossterm/blob/b31782249e3ef1c2e5d8f2f668763fddb2e0a060/src/style/types/color.rs#L126).

Just for interest, AFAIK the ANSI colors are defined as "normal" (e.g. "red") and "bright" (e.g. "bright red"). In crossterm, why are the colors named "dark" (e.g. DarkRed) for normal ANSI colors and "normal" (e.g. Red) for bright ANSI colors?

JamRemco:
in version 2.0 Color::White maps to the same as Color::Grey. 
In other words there is no color (bright) white.

TimonPost:
I think

PR #585, opened 2021-07-23
Fix Greys in colors table

mgunyho:
I came across the same as issue #573. This PR makes the docs consistent with the code (I checked that `DarkGrey` is darker than `Grey` which is darker than `White`).

However, it should be noted that at least in the table [here](https://en.wikipedia.org/wiki/ANSI_escape_code#3-bit_and_4-bit), "white" is what crossterm calls "grey", "bright black" or "gray" is what crossterm calls "dark grey" and "bright white" is what crossterm calls "white". With a quick search I couldn't find another good reference for these names.

### DIFF
--- a/src/style/types/color.rs
+++ b/src/style/types/color.rs
@@ -16,14 +16,14 @@ use crate::style::parse_next_u8;
 ///
 /// | Light | Dark |
 /// | :--| :--   |
-/// | `Grey` | `Black` |
+/// | `DarkGrey` | `Black` |
 /// | `Red` | `DarkRed` |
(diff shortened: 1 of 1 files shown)


JamRemco:
in fact all colors are name incorrectly. 
eg: Red should be Red bright/Intensive. and Darkred just Red.
      Blue should be Blue bright/Intensive. and Darkblue just Blue.

Although not the end of the world, its a bit confusing.

Regards,

Remco

TimonPost (approved):
👍

Issue #164, opened 2019-06-22
Make Windows clear the whole buffer and not only the visible part.


TimonPost:
# Task Description
If you call `terminal.clear(ClearType::All)` the whole buffer, the visible, and the invisible part should be cleared. This happens correctly on Linux, however, the winapi implementation only clears the visible buffer. The same as the command 'cls' works. The winai is used on windows versions `<`than 8.

# Task Todo
Make `terminal.clear(ClearType::All)`, clear the whole screen buffer instead of only the visible part.

- [ ] Figure out how to get the screen buffer dimensions, not the visible part, but also the invisible part. 
- [ ] Replace the size [here](https://github.com/TimonPost/crossterm/blob/master/crossterm_terminal/src/terminal/winapi_terminal.rs#L31) so that all screen contents are removed.

_example_
![Naamloos](https://user-images.githubusercontent.com/19969910/59968720-96154c80-953e-11e9-9ba2-ac9a7071d79a.png)

TimonPost:
Can be closed since, this is not needed

Peohta:
Why it is not needed? I am trying to use crossterm on Windows 10 and `terminal.clear(ClearType::All)` behaves weird compared to `cls`.

TimonPost:
Good question. The answer is not archived in my brain anymore. Let's reopen It until I know why ;)

scambier:
Bump, I have this issue on Windows 10 while using [tui](https://docs.rs/tui/0.13.0/tui/index.html)

KnoxZZ:
Only way to make the `terminal.clear(ClearType::All)` to clear the whole screenbuffer would require the usage of the legacy winapi instead of ANSI. And since modern unix terminals handle the ESC [2J inconsistently I don't think it should get handeled differently on windows. 
One way to make it more consisten across all ANSI supporting terminals would be to use ESC [H before the ESC [2J to set the cursor at the top-left of the viewport before clearing.

Termina94:
Having the same issue on linux mint with gnome-terminal and alacritty. Adding another option (`Purge` for lack of a better name) as ANSI 3J seems to fix the issue for me

#591

tobiasvl:
@Termina94 `Purge` already is the name of a `ClearType` though.

Termina94:
@tobiasvl this was merged in August 
https://github.com/crossterm-rs/crossterm/pull/591#issuecomment-898381852

TimonPost:
So this can be closed

PR #591, opened 2021-08-10
Add option to clear history - Update terminal.rs

Termina94:
A possible solution for [issue 164](https://github.com/crossterm-rs/crossterm/issues/164)


TimonPost:
It doesn't fix the issue directly since this particular issue is about winapi if I recall correctly. But still thanks for this addition.

PR #592, opened 2021-08-11
expose terminal::is_raw_mode_enabled()

Canop:
This enables functions which have to work both when the
raw mode is already set and when it isn't:

```
fn query(some_args) -> Result<SomeType, SomeError> {
    use crossterm::terminal::*;
    let switch_to_raw = !is_raw_mode_enabled()?;
    if switch_to_raw {
        enable_raw_mode()?;
    }
    let res = query_in_raw(some_args);
    if switch_to_raw {
        disable_raw_mode()?;
    }
    res
}
```

### DIFF
--- a/src/terminal.rs
+++ b/src/terminal.rs
@@ -98,6 +98,21 @@ use crate::{csi, impl_display, Result};
 
 pub(crate) mod sys;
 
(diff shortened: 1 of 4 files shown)


TimonPost:
Thanks!

PR #595, opened 2021-08-22
Add CSIu parser for unix

blaggacao:
closes #595 

This is used for [`helix-editor`](https://github.com/helix-editor/helix) keymaps.

### DIFF
--- a/src/event/sys/unix/parse.rs
+++ b/src/event/sys/unix/parse.rs
@@ -159,6 +159,7 @@ pub(crate) fn parse_csi(buffer: &[u8]) -> Result<Option<InternalEvent>> {
                     match buffer[buffer.len() - 1] {
                         b'M' => return parse_csi_rxvt_mouse(buffer),
                         b'~' => return parse_csi_special_key_code(buffer),
+                        b'u' => return parse_csi_u_encoded_key_code(buffer),
                         b'R' => return parse_csi_cursor_position(buffer),
                         _ => return parse_csi_modifier_key_code(buffer),
                     }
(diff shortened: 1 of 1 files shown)


src/event/sys/unix/parse.rs
```diff
@@ -241,6 +242,53 @@ pub(crate) fn parse_csi_modifier_key_code(buffer: &[u8]) -> Result<Option<Intern
...
+    let keycode = {
+        if let Some(c) = char::from_u32(codepoint) {
+            match c {
+                '\x1B' => KeyCode::Esc.into(),
+                '\r' => KeyCode::Enter.into(),
+                // Issue #371: \n = 0xA, which is also the keycode for Ctrl+J. The only reason we get
+                // newlines as input is because the terminal converts \r into \n for us. When we
+                // enter raw mode, we disable that, so \n no longer has any meaning - it's better to
+                // use Ctrl+J. Waiting to handle it here means it gets picked up later
+                '\n' if !crate::terminal::sys::is_raw_mode_enabled() => KeyCode::Enter.into(),
```
blaggacao, line 273: I do not fully understand if the raw-mode check is warranted:

CSIu decimal ASCII encoding should be rather unambigous
```
10 -> LF
13 -> CR
```

blaggacao (commented):
One remaining question..

src/event/sys/unix/parse.rs
sigmaSd: This pr have the context: https://github.com/crossterm-rs/crossterm/pull/373

TimonPost (approved):
Late response. Anyhow, thanks! I would prefer to see some tests for this like the other codes. I can assume it work but I don't have the machine to test it myself. Anyhow can't do much harm better some support then non at all.

PR #598, opened 2021-08-24
Don't run raw_mode test on the ci + some cleanup

sigmaSd:


sigmaSd:
I think this commit https://github.com/crossterm-rs/crossterm/pull/598/commits/57f711864e93d7ac1488bf3bcae2223a792fc7bb an API breaking change

Issue #599, opened 2021-09-05
MoveToPreviousLine(0) and MoveToNextLine(0) move cursor by 1 line


and-nissa:
**Describe the bug**
hi all,
i noticed that using `crossterm::cursor::MoveToPreviousLine` and her sister command to move 0 lines moves the cursor by 1 line.

**To Reproduce**
Cargo.toml
```
...
[dependencies]
crossterm = "0.21.0"
```

main.rs
```
use std::io::stdout;

use crossterm::{cursor, execute};

fn main() {
    let stdout = stdout();
    let mut handle = stdout.lock();

    // MoveToPreviousLine(0)
    {
        let original_position = current_line();
        execute!(handle, cursor::MoveToPreviousLine(0)).unwrap();
        let new_position = current_line();

        print!("previous lines moved: {}", original_position - new_position);
    }

    // MoveToNextLine(0)
    {
        let original_position = current_line();
        execute!(handle, cursor::MoveToNextLine(0)).unwrap();
        let new_position = current_line();

        print!("next lines moved: {}", new_position - original_position);
    }
}

// position is a tuple: (column, row)
// this returns the row (at index 1)
fn current_line() -> u16 {
    cursor::position().unwrap().1
}
```

**Expected behavior**
shouldn't these *only* reset the column position? the documentation says that these commands "move the terminal cursor {up/down} the given number of lines...", but that given number is 0.

**OS**
x86_64-unknown-linux-gnu

**Terminal/Console**
xterm

```
$ xterm -version
XTerm(330)
```

**also**
this is my first opened issue, sorry if i messed anything up! i hope this is helpful in some way <3

and-nissa:
it might also be worth mentioning that they behave as expected when the parameter is 1 (they move by 1 line)

sigmaSd:
Hello, as you  can see here https://github.com/crossterm-rs/crossterm/blob/e1260446e94e9a8f7809fef61dc1369b6f8d6e12/src/cursor.rs#L86  crossterm is not doing anything special beides writing the correct eacape sequence, so if it moves with 0 then thats what ansi does.

The value 0 is underspecified in the spec https://en.m.wikipedia.org/wiki/ANSI_escape_code some sequences have a default value but the minimum value is not mentioned.

and-nissa:
mkay~ so then it's a limitation (feature?) of ansi escape code standards.
[...]

PR #600, opened 2021-09-05
Changed MoveToNextLine(0) and MoveToPreviousLine(0) to move 0 lines

and-nissa:
this pr specifies the minimum cursor line movement in commands `cursor::MoveToNextLine` and `cursor::MoveToPreviousLine` (see #599)

are there any exceptions to the behavior outlined in #599? i don't want to make a blanket statement that they will always move a minimum of 1 line, given 0.

### DIFF
--- a/src/cursor.rs
+++ b/src/cursor.rs
@@ -77,19 +77,24 @@ impl Command for MoveTo {
 ///
 /// # Notes
 ///
-/// * Moves a minimum of 1 line.
-/// * Commands must be executed/queued for execution otherwise they do nothing.
+/// Commands must be executed/queued for execution otherwise they do nothing.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub struct MoveToNextLine(pub u16);
 
 impl Command for MoveToNextLine {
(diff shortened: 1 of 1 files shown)


TimonPost:
Thanks for the contribution. I am doubting if we want to limit the API to move one line at a time or that we should create some if check to validate if the line count is `0` and if it is so to ignore the statement.

sigmaSd:
Adding ifs everywhere could imply a tangible performance degredation for no real benefit.

Im personaly of the opinion of just documenting the current behaviour.

TimonPost:
I don't think it will be that tangible. A simple condition check is next to nothing and given that this function most likely isn't going to be called 60 times per second it seems like no concern to me.

sigmaSd:
I'm assuming that other commands will need the same change by this logic (like scroll down/up move down/up).

Well I can't really argue about performance, and tui-rs for example seems to only use MoveTo/Print extensively so It wont be affected.

TimonPost:
Good call. Created an issue for it: https://github.com/crossterm-rs/crossterm/issues/626

PR #606, opened 2021-10-01
make KeyEvent::new a const fn

Canop:
This enables a more convenient declaration of const keys,
for example

```rs
pub const CONTROL_C: KeyEvent = KeyEvent::new(KeyCode::Char('c'), KeyModifiers::CONTROL);
```

Note that other functions could probably be made const, I changed only
this one because I very frequently declare const keys.

If you want I may have a wider look.

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -395,7 +395,7 @@ pub struct KeyEvent {
 }
 
 impl KeyEvent {
-    pub fn new(code: KeyCode, modifiers: KeyModifiers) -> KeyEvent {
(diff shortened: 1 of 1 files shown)


sigmaSd:
You can find those with clippy
`cargo clippy -- -D clippy::missing_const_for_fn`

One note is reverting from const to non const is a breaking change.

Canop:
The ones which would be cool would mostly be some ::default() but for those you need to switch from a derive to a manual implementation so I won't do such a change without discussion.

TimonPost:
O wow, this is supported now!? Great.

PR #609, opened 2021-10-03
Add `tokio` & `async-std` runtime support.

Berrysoft:
- [x] Add `tokio` & `async-std` reference.
- [x] Control them with features.
- [ ] Check if the usage of polling is correct.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,6 +28,8 @@ all-features = true
 [features]
 default = []
 event-stream = ["futures-core"]
+event-stream-tokio = ["event-stream", "tokio"]
+event-stream-async-std = ["event-stream", "async-std"]
 
 #
 # Shared dependencies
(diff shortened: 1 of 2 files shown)


Berrysoft:
I've noticed that `async-stream` should simplify the implementation without adding `tokio` or `async-std` reference.

Berrysoft:
It seems that it is too naive as it will cause spins on CPU. I'm closing this pull request now, and will rethink my idea.

BoolPurist:
I agree that the underlying stream should not use to spawn a naked OS-thread outside any async runtime.

The [previous attempt to fix this](https://github.com/crossterm-rs/crossterm/pull/609) seems to go back quite a while.

I am willing to pick this up from here. 

However, I have a few questions before starting the rewrite.

1. Is there now any interest in the rewrite ?
2. How to do it in terms of the compile features ? I have the following suggestion.
 - Only supporting tokio under the flag "event-stream"
 - The flag "event-stream" is working the context of tokio. 
   One new compiler flag "asyn-std" is then using asyn-std.
 - Two new flags "tokio" and "asyn-std" in which tokio and asyn-std are respectively used. 
   In this variant, I would deprecate the flag "event-stream" and type EventStream simply 
   since the implementation of this type with naked OS thread should not be used anymore 
   going forward in my opinion.

Issue #536, opened 2021-01-05
Combination of CONTROL + <Non-Alphanumeric> i.e. [ , ], \, / etc. gives, faulty readings or no readings at all.


ahmednooor:
**Describe the bug**
Combination of CONTROL + \<Non-Alphanumeric\> i.e. [ , ], \, / etc. gives, faulty readings or no readings at all.
// OUTPUT
// Press Ctrl+[ : KeyEvent { code: Esc, modifiers: NONE }
// Press Ctrl+] : KeyEvent { code: Char('5'), modifiers: CONTROL }
// Press Ctrl+/ : KeyEvent { code: Char('7'), modifiers: SHIFT | CONTROL }
// Press Ctrl+\ : KeyEvent { code: Char('4'), modifiers: CONTROL }
Some events such as Ctrl+, and Ctrl+. are not read at all. Ctrl+[ is not read in cmd.

**To Reproduce**
1. Go to 'https://github.com/ahmednooor/crossterm-err-rep-05012021/blob/main/src/main.rs#L25-L29'

Clone repo and cargo run.

**Expected behavior**
// Press Ctrl+[ : KeyEvent { code: Char('['), modifiers: CONTROL }
// Press Ctrl+] : KeyEvent { code: Char(']'), modifiers: CONTROL }
// Press Ctrl+{ : KeyEvent { code: Char('{'), modifiers: SHIFT | CONTROL }
// Press Ctrl+} : KeyEvent { code: Char('}'), modifiers: SHIFT | CONTROL }
// Press Ctrl+/ : KeyEvent { code: Char('/'), modifiers: CONTROL }
// Press Ctrl+\ : KeyEvent { code: Char('\'), modifiers: CONTROL }
// Press Ctrl+, : KeyEvent { code: Char(','), modifiers: CONTROL }
// Press Ctrl+. : KeyEvent { code: Char('.'), modifiers: CONTROL }

**OS**
 - e.g. Windows

**Terminal/Console**
 - e.g. Windows Terminal

> P.S. Thank you for creating this. It has worked beautifully so far.

ahmednooor:
Could be related to https://github.com/crossterm-rs/crossterm/issues/504

ahmednooor:
**Some Initial Findings**

After some `println!` here and there, it seems that,

in `crossterm_winapi>src>console.rs>read_input()>ReadConsoleInputW()`, 

[...]

PR #619, opened 2021-11-21
make windows key parsing logic better

WindSoilder:
Handle #536 in windows

For more reference about the comment:
> The other implementation is to take the ASCII code produced by the key and bitwise AND it with 31, forcing bits 6 and 7 to zero. For example, pressing "control" and the letter "g" or "G" (code 107 in octal or 71 in base 10, which is 01000111 in binary), produces the code 7 (Bell, 7 in base 10, or 00000111 in binary).

https://en.wikipedia.org/wiki/Control_character

And here:
> Pressing Ctrl together with a letter or other symbol written 10xxxxx₂ (binary notation) sends the control character whose code is 00xxxxx₂, e.g. Ctrl+[ sends character number 27₁₀ = 0011011₂ because [ is 91₁₀ = 1011011₂.

https://vi.stackexchange.com/questions/3225/disable-esc-but-keep-c

### DIFF
--- a/src/event/sys/windows/parse.rs
+++ b/src/event/sys/windows/parse.rs
@@ -95,7 +95,7 @@ fn parse_key_event_record(key_event: &KeyEventRecord) -> Option<KeyEvent> {
                     // it with 31, forcing bits 6 and bits 7 to zero.
                     // So we can make a bitwise OR back to see what's the raw control character.
                     let c = character_raw as u8;
-                    if c < b'\x1F' {
+                    if c <= b'\x1F' {
                         character = (c | b'\x40') as char;
                     } else {
                         return None;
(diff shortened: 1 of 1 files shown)


superlou:
Using the [event-read.rs](https://github.com/crossterm-rs/crossterm/blob/master/examples/event-read.rs) example, I verified on Windows 10 that nothing is printed for ctrl+[ with Crossterm 0.22. Applying this PR to master provides the expected result:

```
Event: Key(KeyEvent { code: Char('['), modifiers: CONTROL })
```

archseer:
Hey @TimonPost sorry for the bump but could we get a review on this?

TimonPost:
Thanks for the PR, seems to work quite well!

PR #629, opened 2022-02-16
fix control-character parsing in windows

WindSoilder:
Sorry for introducing a new bug in #619 

The bug is: when user enters `ctrl-g`, the result will be `G with modifier CONTROL`.  And we need to convert from upper case `G` to lowercase `g`.

This pr is trying to fix it.

### DIFF
--- a/src/event/sys/windows/parse.rs
+++ b/src/event/sys/windows/parse.rs
@@ -99,7 +99,7 @@ fn parse_key_event_record(key_event: &KeyEventRecord) -> Option<KeyEvent> {
                         character = (c | b'\x40') as char;
                         // if we press something like ctrl-g, we will get `character` with value `G`.
                         // in this case, convert the `character` to lowercase `g`.
-                        if character.is_ascii_uppercase() {
+                        if character.is_ascii_uppercase() && !modifiers.contains(KeyModifiers::SHIFT) {
                             character.make_ascii_lowercase();
                         }
                     } else {
--- a/src/event/sys/windows/parse.rs
+++ b/src/event/sys/windows/parse.rs
@@ -97,6 +97,11 @@ fn parse_key_event_record(key_event: &KeyEventRecord) -> Option<KeyEvent> {
                     let c = character_raw as u8;
                     if c <= b'\x1F' {
                         character = (c | b'\x40') as char;
+                        // if we press something like ctrl-g, we will get `character` with value `G`.
+                        // in this case, convert the `character` to lowercase `g`.
+                        if character.is_ascii_uppercase() {
+                            character.make_ascii_lowercase();
+                        }
                     } else {
                         return None;
                     }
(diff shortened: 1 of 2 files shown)


archseer:
But does this fully fix the issue, what it `ctrl shift g` is pressed (`ctrl-G`)? Could some unit tests be added for the various cases?

superlou:
```
Event: Key(KeyEvent { code: Char('g'), modifiers: NONE })            <-- g
Event: Key(KeyEvent { code: Char('G'), modifiers: SHIFT })           <-- shift+g
Event: Key(KeyEvent { code: Char('g'), modifiers: CONTROL })         <-- ctrl+g
Event: Key(KeyEvent { code: Char('g'), modifiers: SHIFT | CONTROL }) <-- ctrl+shift+g
```

Since shift+G returns `Char('G')`, is ctrl+shift+G also supposed to return `Char('G')`?

WindSoilder:
> Could some unit tests be added for the various cases?

I'll try where to add this, but it may takes some time.

>Since shift+G returns Char('G'), is ctrl+shift+G also supposed to return Char('G')?

For this case, I've compared between this version and 0.22, and tried it on Unix, it seems that they all interpreted as `Char('G')`, when cap lock is on, `shift+G` returns `Char('g')`, I can't find an issue talking about this, so I supposed it's the expected behavior?

TimonPost (approved):

TimonPost:
Not sure what is desired behaviour there. Seems like that if SHIFT is pressed with other control keys the character should be just capitalized right?

superlou:
On Linux Mint, I get:

```
Event: Key(KeyEvent { code: Char('g'), modifiers: NONE })            <-- g
Event: Key(KeyEvent { code: Char('G'), modifiers: SHIFT })           <-- shift+g
Event: Key(KeyEvent { code: Char('g'), modifiers: CONTROL })         <-- ctrl+g
                                                                     <-- ctrl+shift+g
Event: Key(KeyEvent { code: Char('G'), modifiers: SHIFT | ALT })     <-- alt+shift+g
```

Unfortunately, I don't get ctrl+shift+G, though something in the desktop environment might be catching it. However, shift+alt+G capitalizes the char code. However, the alt+shift+g case makes me exepct ctrl+shift+G should behave similarly. Is there a document or wikipage somewhere defining how crossterm's intent?

TimonPost:
When pressing SHIFT an uppercase is printed in all applications, we decided in the past that crossterm should also do this. Similarly, with capslock. When capslock is enabled the user will receive SHIFT + capitalized char. However, I do get your point, more or less a preference thing.

TimonPost:
On ubuntu, it also won't record SHIFT + CONTROL keys. Tho Unix terminals are more limited in their event support. Windows has a easier API for reading input.

src/event/sys/windows/parse.rs
```diff
@@ -97,6 +97,11 @@ fn parse_key_event_record(key_event: &KeyEventRecord) -> Option<KeyEvent> {
                     let c = character_raw as u8;
                     if c <= b'\x1F' {
                         character = (c | b'\x40') as char;
+                        // if we press something like ctrl-g, we will get `character` with value `G`.
+                        // in this case, convert the `character` to lowercase `g`.
+                        if character.is_ascii_uppercase() {
```
TimonPost, line 102: This would leave capitalize char when `SHIFT + CTRL` is pressed. 

```suggestion
                        if character.is_ascii_uppercase() && !modifiers.contains(KeyModifiers::SHIFT) {
```

TimonPost (changes requested):

TimonPost:
Thanks.

Issue #640, opened 2022-03-07
Feature Request: Add the ability to ignore certain mouse events.


markus-bauer:
### Problem
This issue came out of this: https://github.com/crossterm-rs/crossterm/issues/635.

Let's say you're reading events using poll and are only interested in mouse clicks. At the moment any mouse movement will send an event and cause poll to return. This makes poll useless, since it will always immediately return as long as the mouse is being moved.

### Solution
The user should be able to define which events are being reported and which are being ignored.
For example, Curses has the function `mousemask`:
https://manpages.debian.org/bullseye/ncurses-doc/mousemask.3ncurses.en.html

As was mentioned in https://github.com/crossterm-rs/crossterm/issues/635#issuecomment-1055152689, some of this is already handled, just not exposed:
https://github.com/crossterm-rs/crossterm/blob/9bc5cd3527a2d26caa6e2ccdc499beb4764fca95/src/event.rs#L241-L254

mrjackwills:
I have come across an issue, so far in my testing it appears via Wezterm on Windows & WSL & Ubuntu, and whatever the popOS terminal is, in where mouse events are printed to the screen, in my opinion, erroneously.

I have a very simple example Rust application [here](https://www.github.com/mrjackwills/mouse_err), and the video below will show the issue in said application.

I had encountered this issue, intermittently, in my application, [oxker](https://www.github.com/mrjackwills/oxker), which uses the [tui-rs](https://github.com/fdehau/tui-rs) with a Crossterm backend, and on occasion on application exit, a stream of gibberish text, which I later learnt was the mouse events, was printed to the screen.

In my application, I only need to capture mouse events when a mouse button is clicked, and so in order to achieve this I have replaced the `execute!(io::stdout(), EnableMouseCapture)` with `io::stdout().write_all(concat!(crossterm::csi!("?1000h"),crossterm::csi!("?1015h"),crossterm::csi!("?1006h"),).as_bytes()).unwrap_or(())`, which so far in my testing seems to have completely alleviated the issue.


I build [oxker](https://www.github.com/mrjackwills/oxker) for x86, arm6, and arm64, and I run it on a low powered raspberry pi zero w, and when using the `csi!` option, as above, instead of `EnableMouseCapture`, as it was previously implemented, I have screen a huge improvement of performance, which is much needed to such a low end machine.

I would like to help implement this fine-grained control option, for those who need it.

[...]

PR #641, opened 2022-03-07
Change Enable/Diable_MouseCapture to be more granular

sigmaSd:
fix https://github.com/crossterm-rs/crossterm/issues/640

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -294,6 +294,71 @@ impl Command for DisableMouseCapture {
     }
 }
 
+/// A command that enables mouse event capturing with more granualarity.
+///
+/// Mouse events can be captured with [read](./fn.read.html)/[poll](./fn.poll.html).
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum MouseCaptureOptions {
+    /// Capture mouse button click events
+    // Normal tracking: Send mouse X & Y on button press and release
+    CaptureButton(bool),
+    /// Capture mouse button clicks + drag events
+    // Report button motion events (dragging)
+    CaptureDrag(bool),
+    // Any-event tracking: Report all motion events
+    /// Capture mouse button click + drag + movement events
+    CaptureAny(bool),
+}
+impl Command for MouseCaptureOptions {
+    fn write_ansi(&self, f: &mut impl fmt::Write) -> fmt::Result {
+        let enable = match *self {
+            Self::CaptureButton(enable) => {
+                let cmd = if enable {
+                    csi!("?1000h")
+                } else {
+                    csi!("?1000l")
+                };
+                f.write_str(cmd)?;
+                enable
+            }
+            Self::CaptureDrag(enable) => {
+                let cmd = if enable {
+                    csi!("?1002h")
+                } else {
+                    csi!("?1002l")
+                };
+                f.write_str(cmd)?;
+                enable
+            }
+            Self::CaptureAny(enable) => {
+                let cmd = if enable {
+                    csi!("?1003h")
+                } else {
+                    csi!("?1003l")
+                };
+                f.write_str(cmd)?;
+                enable
+            }
+        };
+        if enable {
+            f.write_str(concat!(
+                // RXVT mouse mode: Allows mouse coordinates of >223
+                csi!("?1015h"),
+                // SGR mouse mode: Allows mouse coordinates of >223, preferred over RXVT mode
+                csi!("?1006h"),
+            ))?;
+        } else {
+            f.write_str(concat!(csi!("?1015l"), csi!("?1006l"),))?;
+        }
+        Ok(())
+    }
+
+    #[cfg(windows)]
+    fn execute_winapi(&self) -> Result<()> {
+        sys::windows::enable_mouse_capture()
+    }
+}
+
 /// Represents an event.
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 #[derive(Debug, PartialOrd, PartialEq, Eq, Clone, Copy, Hash)]


sigmaSd:
Windows is currently hardcoded to use the winapi which doesn't seem to offer this level of control according to https://docs.microsoft.com/en-us/windows/console/setconsolemode

What if we make windows use ansi as well, maybe it understand these codes?

markus-bauer:
First of all, thank you for the quick response and making this pull request.

I did a quick test (not extensive) with several linux terminal and it works as expected.

Some observations:
1) Should the more esoteric options be exposed: I wouldn't know what mouse mode I should the user enable? Rxvt, SGR, both or neither?
3) This breaks backwards compatibility:
I think the commands `EnableMouseCapture` and `DisableMouseCapture` should stay as they are and call something like `EnableMouseCaptureOptions::default()`.
4) While testing this, it seemed as if `MouseTrackingMode::Any` doesn't just "Report all motion events" but all events (motion, drag and click). This would mean that there really are only three options that the user might want: any, drag (button+motion), button.
So the Interface could be `EnableMouseCapture` (=`EnableMouseCaptureAll`), `EnableMouseCaptureButton`,  `EnableMouseCaptureDrag`. This is obviously only true if my assumption about `Any` is true.
5) `write!(f, "{}", mode.enable()` can be replaced with `f.write_str(mode.enable())`

sigmaSd:
I agree with all these points, for 3 I had the same thoughts If someone can clear the windows situation I'll rewrite it as you suggested

sigmaSd:
Here is another take https://github.com/crossterm-rs/crossterm/compare/master...sigmaSd:mouse1?expand=1 @markus-bauer

markus-bauer:
Yes this also works.

I realized, that there are cases where this might not do what the user thinks it does. For example:

```rust
stdout.execute(MouseCaptureOptions::CaptureButton(true));
// This does not disable all events:
stdout.execute(MouseCaptureOptions::CaptureAny(false));

// Only after this are button events disabled:
stdout.execute(MouseCaptureOptions::CaptureButton(false));

```
This is just a thought:  CaptureAny(false) _could_ disable everything (Any, Drag, Button). Similarly, CaptureDrag(false) could disable Drag and Button. Otherwise, this has to be documented properly.

---
I would like to know the opinion of @TimonPost, before continuing:

1) What should be done about the windows code?
2) Would you merge this, if it can't be made cross platform?
3) Should this expose all options or should it be the three options we mentioned?
4) What should the interface be like?
    The proposals by @sigmaSd so far were:

    a) 
    ```rust
    EnableMouseCapture([MouseTrackingMode::Normal, ...])
    DisableMouseCapture([MouseTrackingMode::Normal, ...])
    ```
    b)
    ```rust
    // enable:
    MouseCaptureOptions::CaptureButton(true)
    // disable:
    MouseCaptureOptions::CaptureButton(false)

    // and similarly for other options:
    MouseCaptureOptions::CaptureAny(true)
    ...
    ```


    c) I would suggest: If there are only 3 options, then there could be 3 separate commands:

    ```rust
    stdout.execute(EnableMouseCapture)
    stdout.execute(DisableMouseCapture)

    stdout.execute(EnableMouseCaptureClick)
    stdout.execute(DisableMouseCaptureClick)
    
    stdout.execute(EnableMouseCaptureDrag)
    stdout.execute(DisableMouseCaptureDrag)
    ```

sigmaSd:
Regarding windows, recent  versions support ansi well, it could be it just works, if someone wants to test you need to use this branch https://github.com/crossterm-rs/crossterm/compare/master...sigmaSd:mouse1?expand=1

src/event.rs
```diff
@@ -232,26 +232,58 @@ where
     reader.read(filter)
 }
 
+/// Mouse capture mode
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum MouseTrackingMode {
```
TimonPost, line 237: Maybe something like a bitmask (attributes) (already have that in the deps) work here?

TimonPost:
We could hide the implementation behind a target platform feature flag. But that will make the code break if you switch to windows. Alternatively, we could also just do nothing, and add in the comments the behavior will only properly work on UNIX. Defaulting to just passing on all events seem reasonable to me, at least it doesn't break the API. Currently, key combinations work the same way, windows support way more than Unix.

TimonPost:
I would prefer this API:
```
EnableMouseCapture([MouseTrackingMode::Normal, ...])
DisableMouseCapture([MouseTrackingMode::Normal, ...])
```
or maybe with the bitmask we can do:
```
EnableMouseCapture(MouseTrackingMode::Normal | MouseTrackingMode::Other)
DisableMouseCapture(MouseTrackingMode::Normal, ...)
```

markus-bauer:
But what do you think what the actual goal of this should be?
1) Make it possible for the user to set/unset all mouse flags. 
2) Set one of three modes (Move+Drag+Click, Drag+Click, Click) and handle the rest under the hood.

TimonPost:
I would also not know when to enable Rxvt or sgr. It seems cleaner if the user can just combine move, drag, click. But I am also not sure if we can detect which one to use if the user would not pass it?

markus-bauer:
As I mentioned above, I don't thinks this is not how it works. They seem to be distinct modes.

If you enable Drag, then it sends events for Drag and Button (and Scroll). So you don't need to combine them.
If you enable Any it will send all events. If you then disable Button, for example, it will still send all events, because this mode is still enabled (at least this is how this pull request works).

markus-bauer:
Quick correction: Before, I was testing in gnome-terminal, but in xterm I get different results:

```rust
stdout.execute(MouseCaptureOptions::CaptureAny(true));
// this turns off *all* events:
stdout.execute(MouseCaptureOptions::CaptureButton(false));
```

```rust
stdout.execute(MouseCaptureOptions::CaptureAny(true));
// this overwrites Any. Now only Button events are reported:
stdout.execute(MouseCaptureOptions::CaptureButton(true));
```

If this is true, then these modes don't even combine. And this would mean, that the existing code for `EnableMouseCapture` doesn't work as it seems, because the later modes would overwrite the earlier ones. You can test it by changing the sequence in `EnableMouseCapture`.

I don't have time to do more testing today. I will look into it next week.

src/event.rs
```diff
@@ -232,26 +232,58 @@ where
...
+    ButtonEvent,
+    /// Any-event tracking: Report all motion events
+    Any,
+    /// RXVT mouse mode: Allows mouse coordinates of >223
+    RXVTMode,
+    /// SGR mouse mode: Allows mouse coordinates of >223, preferred over RXVT mode
+    SGRMode,
+}
+impl MouseTrackingMode {
+    fn enable(&self) -> &'static str {
```
markus-bauer, line 250: @sigmaSd There's a mistake in the original pull request:
enable and disable are not symmetric. For example Normal should be 1000h and 1000l, but you just copied the reversed sequence from the current EnableMouseCapture.

Can you fix this, because I would like to do some tests where I need to set and unset these modes in sequence.

src/event.rs
```diff
@@ -232,26 +232,58 @@ where
...
+            MouseTrackingMode::SGRMode => csi!("?1006h"),
+        }
+    }
+    fn disable(&self) -> &'static str {
+        match self {
+            MouseTrackingMode::Normal => csi!("?1006l"),
+            MouseTrackingMode::ButtonEvent => csi!("?1015l"),
+            MouseTrackingMode::Any => csi!("?1003l"),
+            MouseTrackingMode::RXVTMode => csi!("?1002l"),
+            MouseTrackingMode::SGRMode => csi!("?1000l"),
```
markus-bauer, line 265:
```suggestion
            MouseTrackingMode::Normal => csi!("?1000l"),
            MouseTrackingMode::ButtonEvent => csi!("?1002l"),
            MouseTrackingMode::Any => csi!("?1003l"),
            MouseTrackingMode::RXVTMode => csi!("?1015l"),
            MouseTrackingMode::SGRMode => csi!("?1006l"),
```

markus-bauer:
Here are some tests (This is a long one):

### Setup
- This uses the original pull request (including the recent fixes), because you can enable/disable the modes individually and as a sequence.

- System: Ubuntu 21.10

- Terminal Emulators:
    - `G: GNOME Terminal 3.38.1 using VTE 0.64.2 +BIDI +GNUTLS +ICU +SYSTEMD`
    - `X: XTerm(366)`
    - `K: kitty 0.24.2`
    - `R: rxvt-unicode (urxvt) v9.22`

- Terminal Emulators were newly started for every run.

- Mouse events are defined as:
    - `Move`
    - `Button` (incl. Scroll)
    - `Drag`
    - `All` = `Move` + `Button` + `Drag`.
    - `None` = no mouse events

- Program used for testing (code at the end of this post):
    1) enable a sequence of mouse modes
    2) optionally disable some mouse modes
    3) get events in a loop. The following results show which kind of events this sends.
    4) at the end: disable the mouse modes

---

### Results

##### a) Enable(Any), loop, Disable(Any):
- G: All
- X: All
- K: All
- R: All

##### b) Enable(Normal), loop, Disable(Normal):
- G: Button
- X: Button
- K: Button
- R: Button

##### c) Enable(Any, Normal), loop, Disable(Any, Normal):
- G: All
- X: Button
- K: Button
- R: Button
    
##### d) Enable(Normal, Any, ButtonEvent), loop, Disable(Normal, Any, ButtonEvent):
- G: All
- X: Drag + Button
- K: Drag + Button
- R: Drag + Button

##### e) Enable(Normal, Any), loop, Disable(Normal, Any):
- G: All
- X: All
- K: All
- R: All

##### f) Enable(Normal, Any), Disable(Normal), loop, Disable(Any)
- G: All
- X: None
- K: None
- R: All

##### g) Enable(Normal, Any), Disable(Any), loop, Disable(Normal)
- G: Button
- X: None
- K: None
- R: None

##### h) Enable(Any, Normal), Disable(Normal), loop, Disable(Any)
- G: All
- X: None
- K: None
- R: None

##### i) Enable(Any, Normal), Disable(Any), loop, Disable(Normal)
- G: Button
- X: None
- K: None
- R: Button

##### j) Enable(Normal), Disable(Any), loop, Disable(Normal)
- G: Button
- X: None
- K: None
- R: Button

##### k) Enable(Any), Disable(Normal), loop, Disable(Any)
- G: All
- X: None
- K: None
- R: All
---

**Can someone please reproduce/confirm at least some of the tests. And try other terminals.**

--- 
### Some observations:
- The bug in the original pull request showed, that enabling and disabling has to be symmetric, otherwise it will leave the terminal in a broken state.
- Combining modes can lead to unexpected and unpredictable behaviour and seems to be dependent on the implementation of the terminal emulator.
- Combining modes is not necessary, since they encompass each other:
    - Normal = Button
    - ButtonEvent = Drag + Button
    - Any = Move + Drag + Button
- Modes are distinct. You could assume the following, but it isn't true: Enable(Any) + Disable(ButtonEvent) = (Move + Drag + Button) - (Button + Drag) = Move

---

### Example program used for testing:

```rust
use crossterm::{
    cursor::MoveTo,
    event::{
        poll, read, DisableMouseCapture, EnableMouseCapture, Event, KeyCode, KeyEvent,
        MouseEventKind, MouseTrackingMode,
    },
    style::Print,
    terminal::{self, Clear, ClearType, EnterAlternateScreen, LeaveAlternateScreen},
    ExecutableCommand, Result,
};
use std::io::stdout;
use std::time::Duration;

fn main() -> Result<()> {
    let mut stdout = stdout();
    terminal::enable_raw_mode()?;
    stdout.execute(EnterAlternateScreen)?;

    // -- parameter:
    stdout.execute(EnableMouseCapture([
        MouseTrackingMode::Any,
        MouseTrackingMode::Normal,
    ]))?;

    // -- parameter:
    stdout.execute(DisableMouseCapture([MouseTrackingMode::Any]))?;

    loop {
        if poll(Duration::from_millis(500)).unwrap() {
            match read().unwrap() {
                Event::Key(KeyEvent {
                    code: KeyCode::Char('q'),
                    ..
                }) => break,
                Event::Mouse(event) => {
                    stdout.execute(MoveTo(0, 0))?;
                    match event.kind {
                        MouseEventKind::Up(..) => stdout.execute(Print("up"))?,
                        MouseEventKind::Down(..) => stdout.execute(Print("down"))?,
                        MouseEventKind::ScrollUp => stdout.execute(Print("scroll up"))?,
                        MouseEventKind::ScrollDown => stdout.execute(Print("scroll down"))?,
                        MouseEventKind::Drag(..) => stdout.execute(Print("drag"))?,
                        MouseEventKind::Moved => stdout.execute(Print("move"))?,
                    };
                    stdout.execute(Clear(ClearType::UntilNewLine))?;
                }
                _ => (),
            }
        }
    }

    // -- parameter:
    stdout.execute(DisableMouseCapture([MouseTrackingMode::Normal]))?;

    terminal::disable_raw_mode()?;
    stdout.execute(LeaveAlternateScreen)?;

    Ok(())
}
```

markus-bauer:
I probably won't contribute to this PR anymore. Most has been said and the final decisions are not mine to make.
My thoughts on the situation:

1) We should keep `EnableMouseCapture`/`DisableMouseCapture` as they are.
The main reason is, that we should not mess with backwards compatibility. These commands should be a **high level** command to just enable/disble all mouse functionality.
Although the unix code can perhaps me simplified (we might not need to set all modes).
2) There should be a new **low level** api that lets the user set any mouse mode they want. It will be on the user to not break things.
This should be available on all platforms: If someone on windows wants to use xterm codes, why shouldn't we let them? There's always the option to fall back to `EnableMouseCapture`
3) Based on this, I think the interface for the low level api I like the most is this:
```rust
// multiple
EnableMouseModes([MouseMode::Any, MouseMode::Rxvt])
DisableMouseModes([MouseMode::Any, MouseMode::Rxvt])

// optional: single mode
EnableMouseMode(MouseMode::Any)
DisableMouseMode(MouseMode::Any)
```
If my tests are correct, then it might be relevant in what order the Modes are set:
`Enable(Any, Normal)` might not give the same results as `Enable(Normal, Any)`. So flags might not work. But I'm not sure about this.
4) Additionally to 1), we could introduce new high level Commands. If `EnableMouseCapture` boils down to setting `MouseMode::Any` (+ SGR/RXVT), then there could be Commands for the other two modes: `EnableMouseCaptureButton` = `MouseMode::Normal` (+...) and  `EnableMouseCaptureDrag` = `MouseMode::ButtonEvent` (+...)

---

Summary:

1) low level api:
```rust
EnableMouseModes([MouseMode::Any, MouseMode::Rxvt])
DisableMouseModes(MouseMode::Any, ...])
```
2) new high level Commands:
`EnableMouseCaptureButton`/`DisableMouseCaptureButton` and `EnableMouseCaptureDrag`/`DisableMouseCaptureDrag` 
3) perhaps: simplify unix code of `EnableMouseCapture`

TimonPost:
Sorry for being idle. One question I would have if users often swish between modes, enabling certain, disabling others. Now I am not a CLI app developer, but I imagine one would like enable input in some way, then leave it like that for the duration of the program. Reading the previous comments it seems as if the behaviour of those flags is very unpredictable for various platforms. If this is true, I guess we could just expose everything and let the users configure it as it desires. This also aligns with crossterm its view, don't keep to much state and internal logic, and let users decide how to implement behaviour with it. Yes, this might make the API for this particular case a bit more complicated, but I think we could just add a comment with a recommendation like 'enable all mode'. If users start to experiment with enable/disable features its up to them to make it work. And we will just comment that it might have unexpected results.

TimonPost:
Thus, agreeing with the 1:

EnableMouseModes([MouseMode::Any, MouseMode::Rxvt])
DisableMouseModes(MouseMode::Any, ...])

This to mee seems to make most sense, as playing with combinations is upto the users. Thus, this PR is on the right track working towards this.

markus-bauer:
I agree with everything you said. The way to go is version 1 (what you said).
In fact, this is what this PR already is.

About switching modes: In my program, I would enable the modes I want at the start, and disable the same ones at the end. I don't think switching modes would be useful or even recommended. 

The weird thing is:
It seems that combining modes doesn't always do what you expect and the behavior might be up to the terminal.
And, combining modes might not be necessary.
Then the user might ask: Then why should I pass several modes to EnableMouseModes?
You certainly have to document this feature properly.

---

But what do you think about the rest? Did I explain the situation properly?

1) If my tests were correct, then there might not be such a thing as enabling all modes at once (like the current version of `EnableMouseCapture` thinks it is doing). Take this example from before:
```
Enable(Normal, Any, ButtonEvent):
    G: All
    X: Drag + Button
    K: Drag + Button
    R: Drag + Button
```
And a quick test I did just now gave this:
```
Enable(ButtonEvent, Normal):
    G: Button + Drag
    R=X=K: Button

Enable(Any, ButtonEvent, Normal):
    G: All
    R=X=K: Button
```
So they don't combine.
This means `EnableMouseCapture` just works, because you set Any as the last mode.

2) And if this is true, then `EnableMouseCapture` will turn into the equivalent of
```
EnableMouseModes([MouseMode::Any, MouseMode::Rxvt, MouseMode::SGR])
```

So the other modes (which enable subsets of Any) would be analogous:
```
EnableMouseCaputureButton = EnableMouseModes([MouseMode::Normal, MouseMode::Rxvt, MouseMode::SGR])
EnableMouseCaputureDrag = EnableMouseModes([MouseMode::ButtonEvent, MouseMode::Rxvt, MouseMode::SGR])
```

TimonPost:
Interesting finding. And nice writeup btw! I dont run a Linux PC so I cant easily validate your results. How would you suggest going on with this? Seems to me there is one good solution right now: Just allow all combinations and write documentation on how to achieve the three most sense-making modes (all, drag, button) while stressing that combining things leads to unexpected behavior across various platforms.

markus-bauer:
About combining: 

https://invisible-island.net/xterm/ctlseqs/ctlseqs.html#h2-Mouse-Tracking:

<blockquote> 
There are two sets of mutually exclusive modes:

   o   mouse protocol

   o   protocol encoding

</blockquote>

And from https://terminalguide.namepad.de/mouse/:

<blockquote> 

Configuration is for 2 independent aspects:
 
 - What events are sent and their button encoding.
 - Which report encoding is used

...

Events

These modes are mutually exclusive. The last activated mode wins.

</blockquote>

That last page explains the situation pretty well.

What I understand from reading this is:
There is no combining. There is one mode you set. If you set another one, this will not somehow combine with the previous one. Additionally (and independently) you can set the encoding. I would also assume that only one is set in the terminal at a time.
So, anything beyond [Any, Rxvt] (for example) is probably not needed in practice.

markus-bauer:
But I want to be clear:
I know next to nothing about any of this. Before this PR, I didn't even know what xterm mouse modes are.
All I personally want is a feature to turn off events for mouse movement. I don't know if the mouse modes are the right solution.
I just stumbled on the fact, that different terminals reacted differently to this PR. Then I did the tests. I explained my test, gave the code, and reported the results.

The final version should not be solely based on something I did on my machine, based on no actual knowledge.

sigmaSd:
@markus-bauer  thanks for all the work you have done!

My personal take based on your findings and https://invisible-island.net/xterm/ctlseqs/ctlseqs.html#h2-Mouse-Tracking is to just make the events separate, and enable sgr and rxvt since that's whats already done in the original command.

Regarding your last comment, no one has the full picture and if we wait for someone like that to showup it might take forever. I think a better approach is to start with something and people will use it and report issue and improve it with time. Also this is a new command so its backward compatible so it won't be that much of a problem.

This pr will use ansi on windows if available, I think it should work but I haven't tested it.

TimonPost:
I was trying out your branch but found some issues on windows in powershell. But none of the mouse options seem to work.

sigmaSd:
I don't have windows currently to test, hopefully someone else picks this up if  its still interesting.

Issue #643, opened 2022-03-13
Ctrl + Space Key Event not read on Windows.


thinkingreed:
Ctrl + Space Key Event not read on Windows.

■ Verification environment
・ WindowsTerminal on Windows 10・11
・ PowerShell on Windows 10・11

TimonPost:
Is likely fixed by: https://github.com/crossterm-rs/crossterm/pull/651. If not reopen.

Issue #561, opened 2021-04-22
Some unicode chars not working on windows terminal


Kl4rry:
**Describe the bug**
Some unicode characters like emoji create no events when inserted into the terminal.

**To Reproduce**
1. Wait on input using event::read
2. Insert emoji or other unicode character
3. No event is created

**Expected behavior**
Emojis and other unicode chars should create key events where the KeyCode enum contains an char.

**OS**
Windows 10

**Terminal/Console**
Windows terminal

PR #651, opened 2022-04-10
Windows input improvements

matkaas:
Hi

I was playing around with tui and crossterm and ran into issue #561, so I thought I'd dig in and fix it. While I was there, I then also decided to fix some other issues around missing or inaccurate events for certain key combinations (#536, #643).

The unicode issue turns out to be fairly straightforward; we need to parse surrogate pairs (for Windows Terminal) and handle alt codes (for Conhost).

The missing/inaccurate keys issue turns out to be much tougher. See commit f52347c for details. I don't think it's possible to solve correctly in general -- the available data and APIs are simply insufficient. The solution I propose makes a best-effort attempt at providing an accurate event for all key combinations.
This works as one would want in Windows Terminal, but is unreliable in Conhost if users change their keyboard layout. I could not find any reliable way to detect the active keyboard layout under a Conhost terminal. I decided to settle for a partial solution (works when not changing keyboard layouts) rather than hacking together an unreliable solution.
Regarding a hacky, unreliable solution: It _is_ actually possible to find a window handle for a Conhost terminal that can be queried for the active keyboard layout, but no API directly offers such a handle. I found this out by looking through process/window/thread handles in Spy++ for a Conhost terminal session. Enumerating and digging through attached console processes and their window handles seemed very brittle to me, so I didn't pursue it further. We could dig more in that direction if you think it's worthwhile to have that for Conhost terminals, even if it could break for any future change to the OS/Conhost infrastructure.

### DIFF
--- a/src/event/source/windows.rs
+++ b/src/event/source/windows.rs
@@ -16,6 +16,7 @@ use super::super::{
 pub(crate) struct WindowsEventSource {
     console: Console,
     poll: WinApiPoll,
+    surrogate_buffer: Option<u16>,
 }
 
 impl WindowsEventSource {
@@ -28,6 +29,8 @@ impl WindowsEventSource {
             poll: WinApiPoll::new(),
             #[cfg(feature = "event-stream")]
             poll: WinApiPoll::new()?,
+
+            surrogate_buffer: None,
         })
     }
 }
@@ -41,7 +44,9 @@ impl EventSource for WindowsEventSource {
                 let number = self.console.number_of_console_input_events()?;
                 if event_ready && number != 0 {
                     let event = match self.console.read_single_input_event()? {
-                        InputRecord::KeyEvent(record) => handle_key_event(record),
+                        InputRecord::KeyEvent(record) => {
+                            handle_key_event(record, &mut self.surrogate_buffer)
+                        }
                         InputRecord::MouseEvent(record) => handle_mouse_event(record),
                         InputRecord::WindowBufferSizeEvent(record) => {
                             Some(Event::Resize(record.size.x as u16, record.size.y as u16))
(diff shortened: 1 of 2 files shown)


matkaas:
Whoops, commit 90d11bb breaks the alt code parsing introduced by 0fcce0d. I'll get that fixed.

matkaas:
> Whoops, commit [90d11bb](https://github.com/crossterm-rs/crossterm/commit/90d11bbc6c7b14428ec6b371c8ec7843dd9aed61) breaks the alt code parsing introduced by [0fcce0d](https://github.com/crossterm-rs/crossterm/commit/0fcce0dcb431dc1e903b04e3cefa7260c28668b4). I'll get that fixed.

Made a fixup for 90d11bb; it's fixed in the new commit f52347c.

TimonPost:
What exactly is a surrogate?

matkaas:
> What exactly is a surrogate and what does it try to fix?

In the context of Unicode, surrogates are code points in the range 0xD800 - 0xDFFF which are used in UTF-16 to encode other code points that don't otherwise fit in a single 2-byte code unit (i.e. every code point larger than 65535).
The Wikipedia page for UTF-16 explains in more details: https://en.wikipedia.org/wiki/UTF-16

They're relevant because we retrieve UTF-16 encoded input from the Win32 Console API when using ReadConsoleInputW in crossterm_winapi::console::Console::read_input. When the terminal wants to send us a character from outside the Basic Multilingual Plane (BMP), it encodes the character as a surrogate pair and passes it to us in two consecutive INPUT_RECORDs, each containing one UTF-16 code unit (the `u_char` member). We have to decode the two code units together to retrieve the character (unicode scalar value).

TimonPost (approved):
Awesome! Gave it a try and a test today. Looks like a good solution! Thanks for the contribution.

PR #653, opened 2022-04-17
Use internal doc-links and add a new link

sornas:
- The link to futures:stream::Stream can now be followed when offline.
- The std link now links to the current rust toolchain in use. (e.g.
  https://doc.rust-lang.org/nightly/.. if cargo +nightly doc. Not that
  this really matters for nightly specifically, in this case.)

Additionally, the links are now "upgraded" along with versions. The
updated futures-link pointed to a yanked version, for example.


TimonPost (approved):

TimonPost (approved):
Thanks!

Issue #656, opened 2022-04-25
Escape sequences not working for Git-Bash + Windows Terminal / Alacritty / WezTerm


dnlmlr:
**Describe the bug**
There is some weird behavior with git-bash running in terminal emulators other than the included mingw terminal. Somehow the escape codes are not properly accepted by the terminals and printed to the screen instead. This does not happen for the mingw terminal included with the git-bash installation.

I was able to track this down to the `ENABLE_VIRTUAL_TERMINAL_PROCESSING` console mode flag not being set. The reason why it is not set can be found [in ansi_support.rs](https://github.com/crossterm-rs/crossterm/blob/0b4a06a97fd946dbe11d6eaa9b5db0b319bfb555/src/ansi_support.rs#L41) where it is only enforced when the `TERM` envvar is unset or equal to `dumb` due to the OR short-circuiting.

The git-bash mingw terminal apparently works without `ENABLE_VIRTUAL_TERMINAL_PROCESSING`, but every other terminal emulator I've tried needs this to work even when runing the git-bash shell. On the other hand on the mingw terminal emulator it simply fails by returning `0` for the `GetConsoleMode` & `SetConsoleMode`, so nothing breaks when attempting to set it there.

**Suggested Fix**
Switching around the sides of the OR in the previously mentioned [ansi_support.rs](https://github.com/crossterm-rs/crossterm/blob/0b4a06a97fd946dbe11d6eaa9b5db0b319bfb555/src/ansi_support.rs#L41) should fix the problem without introducing incompatibilities as far as I can tell.

```rust
let supported = std::env::var("TERM").map_or(false, |term| term != "dumb")
            || enable_vt_processing().is_ok();
```
to
```rust
let supported = enable_vt_processing().is_ok() 
            || std::env::var("TERM").map_or(false, |term| term != "dumb");
```
This should first try to set `ENABLE_VIRTUAL_TERMINAL_PROCESSING` (fixing this issue). If that fails, like with the git-bash included mingw terminal, it falls back to checking for the dumb `TERM` envvar instead of the other way around.

**To Reproduce**
1. Create an empty project and add `crossterm`
2. Add the following code to main `execute!(stdout(), Clear(ClearType::CurrentLine),).unwrap();` and build
3. Run the resulting executable (**not** `cargo run`, that works) using git-bash in Windows Terminal, Alacritty or WezTerm

**Expected behavior**
The given code clears the current line without anything else, so nothing should be printed.

**Actual behavior**
[...]

PR #657, opened 2022-04-25
Fix escape codes for git-bash on 3rd party terms

dnlmlr:
- Fix ansi escape codes not being handled properly for git-bash when
  running terminal emulators other than the included mingw terminal
- ENABLE_VIRTUAL_TERMINAL_PROCESSING is necessary for terminal emulators
  other than the included
- Switching around the OR always tries to enable it, but if it doesn't
  work still reports ansi support depending on the `TERM` envvar
- This should fix issue #656


TimonPost (approved):

martinlindhe:
Can workflow be approved so this PR can be merged please?

PR #661, opened 2022-05-04
remove stray println in color serialization

nicklan:
See description.

Further question:
The `rgb_(a,b,c)` format is kinda weird, and non-standard. would you be open to PR to support ser/deser to the more standard hex format (`#RGB`)? I would suggest having deser support both, and serialization switch to the hex format (perhaps feature flagged).

### DIFF
--- a/src/style/types/color.rs
+++ b/src/style/types/color.rs
@@ -240,7 +240,6 @@ impl serde::ser::Serialize for Color {
         };
 
         if str == "" {
-            println!("color: {:?}", self);
             match *self {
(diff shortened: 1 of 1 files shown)


TimonPost:
The use case was that it should be a human-readable string and support both ANSI and RGB colors. Only serializing the hex string still does not tell if it is ansi/rgb color. ani_###, rgb_### could be an alternative but that is not much different from what already exists. If you have an idea on how to improve this feel free to open a PR!

PR #681, opened 2022-06-12
Add support for cursor/arrow keys in application mode

awahlig:
A simple change to add support for Home/End and Arrow Keys in application mode (besides the normal mode).

For example, the Arrow Up key can be sent as:

| Normal mode | Application mode |
|-|-|
| ESC [ A | ESC O A |

I discovered the issue when I couldn’t move the cursor in Helix editor using the Blink Shell app on iOS (which seems to always use the application mode).


TimonPost:
Ah, never mind, found this document: https://documentation.help/PuTTY/config-appcursor.html

Issue #683, opened 2022-06-29
Consider reverting `cursor::MoveToColumn(0)` non-zero handling or changing it?


dsherret:
**Is your feature request related to a problem? Please describe.**

I was using `MoveToColumn(0)` because I assumed it was 0-indexed, but in https://github.com/crossterm-rs/crossterm/pull/627 it was changed to be silently ignored so the output of my code started acting funny when I upgraded crossterm.

It seems better to just let `0` be the same as `1` for `MoveToColumn`? I think some other people might also get confused why it's not working as well and blame crossterm on being buggy.

**Describe the solution you'd like**

`MoveToColumn(0)` is equivalent to `1` or perhaps panics? Alternatively, maybe this should only accept a NonZeroU16, though that api might not be so nice. `0` equivalent to `1` seems most ideal.

I can understand how this current functionality is desired for what's described in #626, but not so much for this.

Johann150:
I think it would make more sense if they were zero indexed because `MoveTo` and `cursor::position` are zero indexed, and this is documented. The phrasing of the documentation lead me to believe that the statement was true for the whole API, which is confusing:
> Top left cell is represented as `0,0`.
-- https://docs.rs/crossterm/0.23.2/crossterm/cursor/struct.MoveTo.html#notes

TimonPost:
I see there is correct confusing how the API operates and in being consistent. The aim is to be 0-based but ANSI codes are often 1-based. `MoveTo` internally increments the passed parameters to conform that. 

![image](https://user-images.githubusercontent.com/19969910/176745085-7888c444-231c-412f-be32-0cf7576e0e73.png)

I think this is indeed something to be reverted. We should then just increment the parameter by one and accept 0 parametes everywere, in all cursor functions.

TimonPost:
Previous time I decided that it was confusing having Move Left by 0 makes the terminal move left by 1. But if you see it like: Move left to the zero'th cell then it makes more sense.

TimonPost:
So the question is more about do we want all cursor functions to be consistent

Johann150:
For me it does not make sense to adjust relative moves in this way. Something like "go two steps left" always has the same meaning, no matter if the coordinate system starts at `0,0` or `1,1`. I would be very confused if I say "go two steps left" and the cursor goes *three* steps left.

Either way, documenting it properly would be good.

TimonPost:
[...]

Issue #632, opened 2022-02-17
MoveToColumn(0) no longer works after version 0.23


chenxiaolong:
**Describe the bug**

After version 0.23, it seems that `MoveToColumn(0)` no longer has any effect. I did a git bisect between 0.22 and 0.23 and it seems the problem was introduced in commit 9a50fd2ce2701bb15f58722495024937143ad34d. I suspect it's due to this check:

https://github.com/crossterm-rs/crossterm/blob/9a50fd2ce2701bb15f58722495024937143ad34d/src/cursor.rs#L137-L139

**To Reproduce**

Small reproducer:

```
use std::io::{stderr, Write};

use crossterm::{
    cursor::MoveToColumn,
    queue,
    style::Print,
    Result,
};

// cargo run --example stderr
fn main() -> Result<()> {
    let stderr = &mut stderr();

    queue!(
        stderr,
        Print("hello"),
        MoveToColumn(0),
        Print("world"),
    )?;

    stderr.flush()?;

    Ok(())
}
```

On Windows and Linux, the output is:

```
helloworld
```

**Expected behavior**

Prior to version 0.23, the output was:

```
world
```

(`hello` being overwritten due to moving the cursor to the leftmost column)

**OS**

Windows 10 21H2; Fedora 35

**Terminal/Console**

Windows Terminal on Windows; Alacritty on Linux

chenxiaolong:
Thanks, I saw that the documentation for `MoveTo()` used 0-indexed coordinates. I didn't realize `MoveToColumn()` would be 1-indexed.

sidju:
It would be less unexpected if crossterm consistently used either 0- or 1-indexed cursor coordinates for all absolute movement, so either MoveTo or MoveToColumn/Row should follow the praxis of the other.

PR #684, opened 2022-06-30
Assure all mouse commands are 0 based and document this.

TimonPost:
There is correct confusion about crossterm claiming to be 0-based at one place but being 1-based on other places. This is inconsistent and is addressed in this PR.  This PR assures all functions are 0-based and reverts the changes made in that just ignored 0-arguments.

Fixes: https://github.com/crossterm-rs/crossterm/issues/632
Fixes: https://github.com/crossterm-rs/crossterm/issues/599
Fixes: https://github.com/crossterm-rs/crossterm/issues/683

### DIFF
--- a/src/cursor.rs
+++ b/src/cursor.rs
@@ -55,7 +55,7 @@ pub(crate) mod sys;
 /// A command that moves the terminal cursor to the given position (column, row).
 ///
 /// # Notes
-///
+/// * This command is 0 based, meaning 0 is 1.
 /// * Top left cell is represented as `0,0`.
 /// * Commands must be executed/queued for execution otherwise they do nothing.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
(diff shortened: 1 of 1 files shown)


src/cursor.rs
```diff
@@ -149,16 +143,14 @@ impl Command for MoveToColumn {
 /// A command that moves the terminal cursor to the given row on the current column.
 ///
 /// # Notes
-///
-/// Commands must be executed/queued for execution otherwise they do nothing.
+/// * This command is 0 based, meaning 0 is 1.
```
Johann150, line 146: Similar to above:
```suggestion
/// * This command is 0 based, meaning 0 is the topmost row.
```

TimonPost:
@Johann150 thanks for review. Made some updates as we discussed. Can you find some time to give it an other look?

Johann150 (approved):
Looks good to me. Tried using this branch in the project where I was confused by the crossterm behaviour earlier and it now works as I would have expected. 👍

PR #687, opened 2022-07-05
Create the SetStyle tuple struct

AhoyISki:
This is a pretty simple PR, it just creates a command for setting a style, given a `ContentStyle`. This style will be applied to all subsequent text, like `Set(Foreground|Background|Underline)Color` or `SetAttribute(s)`.

This command is mostly for convenience, letting the user execute the aforementioned commands within just a single one.

PS: I know this is a breaking change, but shouldn't `ResetColor`really be called `ResetStyle`, since it also resets attributes?

### DIFF
--- a/src/style.rs
+++ b/src/style.rs
@@ -340,6 +340,34 @@ impl Command for SetAttributes {
     }
 }
 
+/// A command that sets a style (colors and attributes).
+///
+/// # Notes
+///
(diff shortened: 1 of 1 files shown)


AhoyISki:
Do you plan to merge this?

TimonPost:
Yes, thanks for the reminder.

TimonPost (approved):

TimonPost:
Nice!

PR #688, opened 2022-07-22
Increase support for kitty enhanced keyboard protocol

pianohacker:
Add parsing of repeat/release key events, and allow sending the CSIs to enable/disable the enhanced keyboard protocol.

See https://sw.kovidgoyal.net/kitty/keyboard-protocol/ for more information.

**Note**: this changes the form of the `KeyEvent` struct.

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -294,6 +294,77 @@ impl Command for DisableMouseCapture {
     }
 }
 
+bitflags! {
+    /// Represents special flags that tell compatible terminals to add extra information to keyboard events.
+    ///
+    /// See https://sw.kovidgoyal.net/kitty/keyboard-protocol/#progressive-enhancement for more information.
+    #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+    pub struct KeyboardEnhancementFlags: u8 {
+        /// Represent Escape and modified keys using CSI-u sequences, so they can be unambiguously
+        /// read.
+        const DISAMBIGUATE_ESCAPE_CODES = 0b0000_0001;
+        /// Add extra events with [`KeyEvent.kind`] set to [`KeyEventKind::Repeat`] or
+        /// [`KeyEventKind::Release`] when keys are autorepeated or released.
+        const REPORT_EVENT_TYPES = 0b0000_0010;
+        /// Send [alternate keycodes](https://sw.kovidgoyal.net/kitty/keyboard-protocol/#key-codes)
+        /// in addition to the base keycode.
+        ///
+        /// *Note*: these are not yet supported by crossterm.
+        const REPORT_ALTERNATE_KEYS = 0b0000_0100;
+        /// Represent all keyboard events as CSI-u sequences. This is required to get repeat/release
+        /// events for plain-text keys.
+        const REPORT_ALL_KEYS_AS_ESCAPE_CODES = 0b0000_1000;
+        /// Send the Unicode codepoint as well as the keycode.
+        ///
+        /// *Note*: this is not yet supported by crossterm.
+        const REPORT_ASSOCIATED_TEXT = 0b0001_0000;
+    }
+}
+
+/// A command that enables extra kinds of keyboard events.
+///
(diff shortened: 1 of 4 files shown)


TimonPost:
Interesting. Would be nice to see the modifiers like Meta implemented as well https://sw.kovidgoyal.net/kitty/keyboard-protocol/#modifiers in a followup. Lots of people commented on that and this protocol seems to allow fetching it

src/event.rs
```diff
@@ -294,6 +294,77 @@ impl Command for DisableMouseCapture {
...
+    }
+
+    #[cfg(windows)]
+    fn execute_winapi(&self) -> Result<()> {
+        unimplemented!("keyboard progressive enhancement not implemented on Windows")
+    }
+
+    #[cfg(windows)]
+    fn is_ansi_code_supported(&self) -> bool {
+        false
```
TimonPost, line 364: if this code is not supported on windows I would prefer `execute_winapi` to return an error rather then panicing. Could you transition that to an `io::Error(ErrorKind::Other)`? Same for the other added command.

src/event.rs
```diff
@@ -294,6 +294,77 @@ impl Command for DisableMouseCapture {
...
+        ///
+        /// *Note*: these are not yet supported by crossterm.
+        const REPORT_ALTERNATE_KEYS = 0b0000_0100;
+        /// Represent all keyboard events as CSI-u sequences. This is required to get repeat/release
+        /// events for plain-text keys.
+        const REPORT_ALL_KEYS_AS_ESCAPE_CODES = 0b0000_1000;
+        /// Send the Unicode codepoint as well as the keycode.
+        ///
+        /// *Note*: this is not yet supported by crossterm.
+        const REPORT_ASSOCIATED_TEXT = 0b0001_0000;
```
TimonPost, line 320: Is there a way to make this private maybe? Or should we perhaps remove the unsupported flags. This just to prevent not working API functionality

src/event.rs
```diff
@@ -294,6 +294,77 @@ impl Command for DisableMouseCapture {
...
+        unimplemented!("keyboard progressive enhancement not implemented on Windows")
+    }
+
+    #[cfg(windows)]
+    fn is_ansi_code_supported(&self) -> bool {
+        false
+    }
+}
+
+/// A command that disables extra kinds of keyboard events.
```
TimonPost, line 346: While you did refer to the documentaiton, I think its goodto also refer to this command is only supported by The [foot terminal](https://codeberg.org/dnkl/foot/issues/319), The [WezTerm terminal](https://wezfurlong.org/wezterm/config/lua/config/enable_kitty_keyboard.html), The [notcurses library](https://github.com/dankamongmen/notcurses/issues/2131), The [neovim text editor](https://github.com/neovim/neovim/pull/18181), The [kakoune text editor](https://github.com/mawww/kakoune/issues/4103), And the [dte text editor](https://gitlab.com/craigbarnes/dte/-/issues/138). Just to be clear for users not to try enabling this flag if they use any other flag.

pianohacker: Good idea.

src/event.rs
pianohacker: Commented out for now.

src/event.rs
pianohacker: Good idea. I went with https://doc.rust-lang.org/stable/std/io/enum.ErrorKind.html#variant.Unsupported .

pianohacker:
Agreed; I didn't want to add too much at once, but the extra modifiers would be great.

I'd also like to add support for https://sw.kovidgoyal.net/kitty/keyboard-protocol/#functional-key-definitions , but that's a weighty PR.

pianohacker:
I ended up adding support for the functional keys, as otherwise they end up as unicode gibberish: https://github.com/pianohacker/crossterm/compare/pianohacker_add_csi_u_types...pianohacker:crossterm:pianohacker_add_functional_keys

I'll create a PR for that after this one.

TimonPost (approved):
Looks good! Thank you so much

PR #689, opened 2022-07-22
Emit focus events

groves:
I'd like to receive [focus change events](https://terminalguide.namepad.de/mode/p1004/) from crossterm to let me save on loss of focus in [Helix](https://helix-editor.com/).

This adds focus emitting on Unix, though I still need to add documentation and tests.

I wanted to open a PR now as I'm not clear on the best way to support this on Windows. 

**Option 1:** Windows Terminal added these Unix-style focus events in [1.14.186](https://github.com/microsoft/terminal/releases/tag/v1.14.1861.0). That came out a couple weeks ago.  I *think* we'd be able to get those codes since we're enabling virtual terminal processing in Windows, but I don't know enough about the Windows terminal to be sure. I don't see us parsing ansi codes anywhere else in the Windows code, so that makes me worry it's not the case. It would also restrict this to really recent Windows terminal versions.

**Option 2:** Listen for the [FOCUS_EVENT_RECORD](https://docs.microsoft.com/en-us/windows/console/focus-event-record-str) in crossterm-winapi. That would be hacky as that focus record is supposed to be internal. It looks like that's what they're using in Windows Terminal to emit the Unix-style codes, so I'm guessing we could use it.

I'm open to any other options, of course. Thoughts?

P.S. This is the first Rust code I've written, and it was super easy to drop it in and modify the example to try it. Thanks for putting together such a well structured project!

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -34,6 +34,8 @@
 //!     loop {
 //!         // `read()` blocks until an `Event` is available
 //!         match read()? {
+//!             Event::FocusGained => println!("FocusGained"),
+//!             Event::FocusLost => println!("FocusLost"),
 //!             Event::Key(event) => println!("{:?}", event),
 //!             Event::Mouse(event) => println!("{:?}", event),
 //!             Event::Resize(width, height) => println!("New size {}x{}", width, height),
@@ -57,6 +59,8 @@
 //!             // It's guaranteed that the `read()` won't block when the `poll()`
 //!             // function returns `true`
 //!             match read()? {
+//!                 Event::FocusGained => println!("FocusGained"),
+//!                 Event::FocusLost => println!("FocusLost"),
 //!                 Event::Key(event) => println!("{:?}", event),
 //!                 Event::Mouse(event) => println!("{:?}", event),
 //!                 Event::Resize(width, height) => println!("New size {}x{}", width, height),
@@ -74,6 +78,7 @@
 
 use std::fmt;
 use std::hash::{Hash, Hasher};
+#[cfg(windows)]
 use std::io;
 use std::time::Duration;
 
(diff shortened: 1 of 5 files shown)


TimonPost:
This document specifies what is supported by windows: https://docs.microsoft.com/en-us/windows/console/console-virtual-terminal-sequences

TimonPost:
It doesn't seem from the specs they support this yet. So then option two seems a good way to do this. Its quite simple to add custom logic for windows only systems by implementing the execute_windows function.

src/event.rs
```diff
@@ -307,6 +326,18 @@ pub enum Event {
     Resize(u16, u16),
 }
 
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[derive(Debug, PartialOrd, PartialEq, Eq, Clone, Copy, Hash)]
```
TimonPost, line 330: Do we expect this struct to have more fields in the future? Otherwise, I would prefer flattening this and just keep the enum.

TimonPost (commented):
Nice!

src/event.rs
groves: I don't think it'll get more fields. I changed this to FocusLost and FocusGained members in the top-level Event enum. I think that's what you meant? Let me know if I misunderstood.

groves:
OK, option 2 implemented! Seems to work on Windows and I believe this is ready to merge. I couldn't think of a straightforward way to test an actual terminal gaining and losing focus. I only wrote a unit test as a result.

@TimonPost I've rebased it on https://github.com/crossterm-rs/crossterm/pull/690 to get it to compile on Windows. If this PR looks good, you can merge this one and close 690. If this still needs work, you may want to merge 690 to fix Windows compilation in the meantime.

src/style.rs
```diff
@@ -366,6 +366,16 @@ impl Command for SetStyle {
 
         Ok(())
     }
+
+    #[cfg(windows)]
```
TimonPost, line 370: If you rebase, i'll approve and merge. Thanks!

groves: Done!

TimonPost (approved):

PR #690, opened 2022-07-24
Add execute_winapi on SetStyle to fix Windows compilation

groves:
Compilation on Windows is failing for me currently with this error:

```
error[E0046]: not all trait items implemented, missing: `execute_winapi`
   --> src\style.rs:351:1
```

I believe https://github.com/crossterm-rs/crossterm/pull/687 introduced the problem.

Reading `execute_fmt`, I believe it'll call `is_ansi_code_supported` on each of the subcommands, so this fix should be good.

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -74,6 +74,7 @@
 
 use std::fmt;
 use std::hash::{Hash, Hasher};
+use std::io;
 use std::time::Duration;
 
 use bitflags::bitflags;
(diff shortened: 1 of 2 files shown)


groves:
https://github.com/crossterm-rs/crossterm/pull/688 broke Windows compilation, too. I've added a fix for it to this branch.

TimonPost:
Oh thanks for patching! Must have slipped through in a previous PR.

PR #691, opened 2022-07-25
Add support for functional key codes from kitty keyboard protocol

pianohacker:
Followup to #688 .

Open question: is the nested structure for the new keycodes the best way? I didn't want to inflate the `KeyCode` enum, but the new structure results in somewhat verbose names.

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -806,8 +806,8 @@ pub enum KeyCode {
     /// Caps Lock key.
     ///
     /// **Note:** this and all following keys can only be read if
-    /// [[`KeyboardEnhancementFlags::DISAMBIGUATE_ESCAPE_CODES`]] has been enabled with
-    /// [[`PushKeyboardEnhancementFlags`].
+    /// [`KeyboardEnhancementFlags::DISAMBIGUATE_ESCAPE_CODES`] has been enabled with
+    /// [`PushKeyboardEnhancementFlags`].
     CapsLock,
     /// Scroll Lock key.
     ScrollLock,
@@ -825,7 +825,7 @@ pub enum KeyCode {
     Media(MediaKeyCode),
     /// A modifier key.
     ///
-    /// The [[`KeyboardEnhancementFlags::REPORT_ALL_KEYS_AS_ESCAPE_CODES`]] flag is required to
+    /// The [`KeyboardEnhancementFlags::REPORT_ALL_KEYS_AS_ESCAPE_CODES`] flag is required to
     /// read these keys.
     Modifier(ModifierKeyCode),
 }
--- a/src/event.rs
+++ b/src/event.rs
@@ -375,9 +375,9 @@ impl Command for PushKeyboardEnhancementFlags {
 
     #[cfg(windows)]
     fn execute_winapi(&self) -> Result<()> {
-        Err(io::Error::new(
-            io::ErrorKind::Unsupported,
-            "Keyboard progressive enhancement not implemented on Windows.",
+        Err(std::io::Error::new(
+            std::io::ErrorKind::Unsupported,
+            "Keyboard progressive enhancement not implemented for the legacy Windows API.",
         ))
     }
 
@@ -402,9 +402,9 @@ impl Command for PopKeyboardEnhancementFlags {
 
     #[cfg(windows)]
     fn execute_winapi(&self) -> Result<()> {
-        Err(io::Error::new(
-            io::ErrorKind::Unsupported,
-            "Keyboard progressive enhancement not implemented on Windows.",
+        Err(std::io::Error::new(
+            std::io::ErrorKind::Unsupported,
+            "Keyboard progressive enhancement not implemented for the legacy Windows API.",
         ))
     }
 
--- a/src/event.rs
+++ b/src/event.rs
@@ -553,6 +553,14 @@ pub enum KeyEventKind {
     Release,
 }
 
+bitflags! {
+    /// Represents extra state about the key event.
+    #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+    pub struct KeyEventState: u8 {
+        const KEYPAD = 0b0000_0001;
+    }
+}
+
 /// Represents a key event.
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 #[derive(Debug, PartialOrd, Clone, Copy)]
@@ -563,6 +571,8 @@ pub struct KeyEvent {
     pub modifiers: KeyModifiers,
     /// Kind of event.
     pub kind: KeyEventKind,
+    /// Keyboard state.
+    pub state: KeyEventState,
 }
 
 impl KeyEvent {
@@ -571,6 +581,7 @@ impl KeyEvent {
             code,
             modifiers,
             kind: KeyEventKind::Press,
+            state: KeyEventState::empty(),
         }
     }
 
@@ -583,6 +594,21 @@ impl KeyEvent {
             code,
             modifiers,
             kind,
+            state: KeyEventState::empty(),
+        }
+    }
+
+    pub const fn new_with_kind_and_state(
+        code: KeyCode,
+        modifiers: KeyModifiers,
+        kind: KeyEventKind,
+        state: KeyEventState,
+    ) -> KeyEvent {
+        KeyEvent {
+            code,
+            modifiers,
+            kind,
+            state,
         }
     }
 
@@ -610,6 +636,7 @@ impl From<KeyCode> for KeyEvent {
             code,
             modifiers: KeyModifiers::empty(),
             kind: KeyEventKind::Press,
+            state: KeyEventState::empty(),
         }
     }
 }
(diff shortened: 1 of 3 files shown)


src/event/sys/unix/parse.rs
```diff
@@ -312,7 +403,9 @@ pub(crate) fn parse_csi_u_encoded_key_code(buffer: &[u8]) -> Result<Option<Inter
         };
 
     let keycode = {
-        if let Some(c) = char::from_u32(codepoint) {
+        if let Some(special_key_code) = translate_functional_key_code(codepoint) {
```
TimonPost, line 406: Before this statement, we gather modifier information. Can you perhaps check in this function if shift, alt, META were pressed and add this to the `KeyModifiers` bitflag?

TimonPost: Just such that we dont get two seperate ways to tell the user about what modifier key was pressed

src/event.rs
```diff
@@ -643,6 +757,31 @@ pub enum KeyCode {
...
+    /// Num Lock key.
+    NumLock,
+    /// Print Screen key.
+    PrintScreen,
+    /// Pause key.
+    Pause,
+    /// Menu key.
+    Menu,
+    /// A key on the keypad.
+    Keypad(KeypadKeyCode),
```
TimonPost, line 777: I am a bit dubious on whether it is a problem if Keypad contains UP/DOWN/LEFT/RIGHT and that KeyCode has them as well. Same for modifiers.

TimonPost: Could you add here:

    /// The [`KeyboardEnhancementFlags::REPORT_ALL_KEYS_AS_ESCAPE_CODES`] flag is required to
    /// read these keys.

src/event.rs
```diff
@@ -643,6 +757,31 @@ pub enum KeyCode {
...
+    /// A media key.
+    Media(MediaKeyCode),
```
TimonPost, line 779: Could you add here:    

     /// The [`KeyboardEnhancementFlags::REPORT_ALL_KEYS_AS_ESCAPE_CODES`] flag is required to
    /// read these keys.

TimonPost:
Like you said it is a bit odd now that KeyCode and Keypad contain similar key events and that modifiers now are duplicated in KeyCode::Modifiers and KeyModifiers struct. I would love to see if we can bring the two closer together. I prefer no duplication as it only confuses the user on where to find certain events. e.g. it might match on KeyCode::Up but rather KeyCode::Keypad(Up) is sent. Or both are sent which can also cause confusion. Similar to modifiers, if you get  KeyCode::Modifier but  the KeyEvent.modifiers don't contain this modifier it might be confusing as well.

TimonPost:
One issue can be resolved quite easily. Just add some entries to KeyModifiers and make sure it is correctly filled.

pianohacker:
The modifier mask being set for the modifier's keypress event is an interesting one. The precedent is to not set the modifier bits for the keypress, which I do think makes sense. Take the following example (from `kitty +kitten show_key -m kitty`, a demo for the protocol). Here, I press Right Alt, then Left Alt, then release Right Alt, then Left Alt:

```
RIGHT_ALT PRESS 
CSI 57449 u

Alt+LEFT_ALT PRESS 
CSI 57443 ; 3 u

Alt+RIGHT_ALT RELEASE 
CSI 57449 ; 3 : 3 u

Alt+LEFT_ALT RELEASE 
CSI 57443 ; 3 : 3 u
```

The `Alt` modifier is only reported for the second Alt press, where I'm already holding down another Alt key.

It is somewhere else to look for the info, and it is confusing. However, it does match the precedent of kitty (and X11, as reported by `xev`). Which would you prefer an easier-to-understand API or one that matches APIs?

As far as the `Keypad` keys go, I agree. The duplication is confusing. When I was working on the extra modifiers PR, I had a thought about a fourth field for `KeyEvent`: `extra`, `state`, or something similar. The kitty keyboard protocol reports the of Caps Lock and Num Lock as additional modifiers, but I didn't want to add those to the `modifiers` field, as it seems like it would lead to confusion. ("My events that used to match don't match unless I mask out the state of Num Lock?") Maybe this extra field could have a flag like `KeyEventState::KEYPAD`?

TimonPost:
But can you maybe do something like this: 

```
let modifiers: KeyModifiers;

match modifier_key_code {
   LeftAlt | RightAlt => modifiers.set(Alt),
   LeftShift | RightShift =>  modifiers.set(Shift),
   LeftMeta | RightMeta => modifiers.set(Meta)
}
```

Or is that what you try to argue isnt possible?

TimonPost:
I agree those so called modifiers Caps Lock, Num Lock, Super, Hyper are to vaque should not be added to the key modifiers. I want to reserver that for SHIFT, CTRL, ALT, META keys.

TimonPost:
> Which would you prefer an easier-to-understand API or one that matches APIs

I prefer consistency and less duplication. I want the user to be able to match on the `modifiers` of key_event and be able to tell if they pressed shift, alt, meta, ctrl. You can still keey the 'Left/Right' pairs in `ModifierKeyCode`. I think we might even try to support that on windows if possible.

src/event.rs
```diff
@@ -643,6 +757,31 @@ pub enum KeyCode {
...
+    /// Caps Lock key.
+    ///
+    /// **Note:** this and all following keys can only be read if
+    /// [`KeyboardEnhancementFlags::DISAMBIGUATE_ESCAPE_CODES`] has been enabled with
+    /// [`PushKeyboardEnhancementFlags`].
+    CapsLock,
+    /// Scroll Lock key.
+    ScrollLock,
+    /// Num Lock key.
+    NumLock,
```
TimonPost, line 769: Is this technically keypad key?

src/event.rs
```diff
@@ -643,6 +757,31 @@ pub enum KeyCode {
     Null,
     /// Escape key.
     Esc,
+    /// Caps Lock key.
+    ///
+    /// **Note:** this and all following keys can only be read if
```
TimonPost, line 762: I prefer a redundancy here. If you will please copy and paste the same text for each key to which this comment is applicable to. It's just to be clear towards the user..

TimonPost: If we update the documentation we'll have to do find replace, but at least we are clear.

TimonPost: Would be awesome to see if windows supports those key combinations

TimonPost: <html>
<body>


VK_SNAPSHOT | 0x2C | PRINT SCREEN key
-- | -- | --



</body>
</html>
<html>
<body>


VK_CAPITAL | 0x14 | CAPS LOCK key
-- | -- | --



</body>
</html>

<html>
<body>


VK_NUMLOCK | 0x90 | NUM LOCK key
-- | -- | --
VK_SCROLL | 0x91 | SCROLL LOCK key


</body>
</html>

https://docs.microsoft.com/en-us/windows/win32/inputdev/virtual-key-codes

Actually pretty easy. I might implement that after this pr is merged.

pianohacker:
I just pushed a couple of commits.

- https://github.com/crossterm-rs/crossterm/pull/691/commits/763ef18cf3a4ba728753ea4a085cd97ffa8a94c9 makes the discussed change, including the modifier mask in the key press event.
- https://github.com/crossterm-rs/crossterm/pull/691/commits/a9d8b334c37a1ba5465659d4ab74cf54e7214d65 makes a change to keypad key presses. They're now mapped to normal keycodes. There's a few things to note here:
  - `Decimal` is mapped to `.`, and `Separator` is mapped to `,`. This is a locale-specific thing.
  - The `Begin` key might be a historical artifact... but it's the one thing on the keypad that doesn't have a normal equivalent.
  - There's a new `state` field, which currently only indicates whether the key press was on the keypad. I think it could also hold Num Lock and Caps Lock.

Speaking of, Caps Lock feels worth adding in a followup PR. It would be nice for something reading a password, for instance, to be able to warn if Caps Lock is on.

src/event.rs
```diff
@@ -507,6 +507,14 @@ pub enum KeyEventKind {
     Release,
 }
 
+bitflags! {
+    /// Represents extra state about the key event.
+    #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+    pub struct KeyEventState: u8 {
+        const KEYPAD = 0b0000_0001;
```
TimonPost, line 514: /// The key event origins from the keypad.

src/event.rs
```diff
@@ -517,6 +525,8 @@ pub struct KeyEvent {
     pub modifiers: KeyModifiers,
     /// Kind of event.
     pub kind: KeyEventKind,
+    /// Keyboard state.
```
TimonPost, line 528:
```suggestion
    /// Keyboard state.
    ///
    /// Only present if [`KeyboardEnhancementFlags::DISAMBIGUATE_ESCAPE_CODES`] has been enabled with
    /// [`PushKeyboardEnhancementFlags`].
```

src/event/sys/unix/parse.rs
```diff
@@ -311,32 +412,54 @@ pub(crate) fn parse_csi_u_encoded_key_code(buffer: &[u8]) -> Result<Option<Inter
...
-    let input_event = Event::Key(KeyEvent::new_with_kind(keycode, modifiers, kind));
+    if let KeyCode::Modifier(modifier_keycode) = keycode {
+        match modifier_keycode {
+            ModifierKeyCode::LeftAlt | ModifierKeyCode::RightAlt => {
+                modifiers.set(KeyModifiers::ALT, true)
+            }
+            ModifierKeyCode::LeftControl | ModifierKeyCode::RightControl => {
+                modifiers.set(KeyModifiers::CONTROL, true)
+            }
+            ModifierKeyCode::LeftShift | ModifierKeyCode::RightShift => {
```
TimonPost, line 453: I am not sure what IsoLevel3Shift, IsoLevel5Shift are. But should we consider adding those here?

TimonPost (approved):
Just a few comments but is good to go.

src/event/sys/unix/parse.rs
pianohacker: I don't think so; those are extra modifiers used in keyboard layouts like [Neo](https://en.wikipedia.org/wiki/Neo_(keyboard_layout)) to switch to extra punctuation/number layers.

src/event.rs
pianohacker: I don't think so; they're off the keypad and not marked as keypad keys in the protocol.

TimonPost:
Thanks a lot @pianohacker !

PR #693, opened 2022-07-28
Add bracketed paste parsing

groves:
This is another attempt at adding [bracketed paste parsing](https://github.com/crossterm-rs/crossterm/issues/545), like https://github.com/crossterm-rs/crossterm/pull/557. This version keeps from removing the `Copy` trait on `Event` and does all of the paste parsing in `parse_event`, so I'm hoping we can get it to an acceptable state.

Since pasting needs to read in a variable length string, I couldn't find a way to stuff the result in `Event` since it has `Copy`. Instead I've added an `input` module with its own read function. That read returns an `Input` enum that either contains `Event` or the pasted data. What I was thinking is that you'd use the `input` module if you wanted to parse terminal input that included variable length data, or you could keep using the `event` module if you only need the fixed length events. I didn't want to break all existing API clients and I wanted to let people pass around Events on the stack, and this was the best compromise I came up with.

I need to document this and give it a spin in Helix. I wanted to get some feedback on the approach before polishing it up though.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,8 @@ all-features = true
 # Features
 #
 [features]
-default = []
+default = ["bracketed-paste"]
+bracketed-paste = []
 event-stream = ["futures-core"]
 
 #
@@ -72,6 +73,10 @@ serde_json = "1.0"
 #
 # Examples
 #
+[[example]]
+name = "event-read"
+required-features = ["bracketed-paste"]
+
 [[example]]
 name = "event-stream-async-std"
 required-features = ["event-stream"]
--- a/src/event/sys/unix/parse.rs
+++ b/src/event/sys/unix/parse.rs
@@ -176,11 +176,15 @@ pub(crate) fn parse_csi(buffer: &[u8]) -> Result<Option<InternalEvent>> {
             } else {
                 // The final byte of a CSI sequence can be in the range 64-126, so
                 // let's keep reading anything else.
-                let last_byte = *buffer.last().unwrap();
+                let last_byte = buffer[buffer.len() - 1];
                 if !(64..=126).contains(&last_byte) {
                     None
                 } else {
-                    match buffer[buffer.len() - 1] {
+                    #[cfg(feature = "bracketed-paste")]
+                    if buffer.starts_with(b"\x1B[200~") {
+                        return parse_csi_bracketed_paste(buffer);
+                    }
+                    match last_byte {
                         b'M' => return parse_csi_rxvt_mouse(buffer),
                         b'~' => return parse_csi_special_key_code(buffer),
                         b'u' => return parse_csi_u_encoded_key_code(buffer),
@@ -650,6 +654,19 @@ fn parse_cb(cb: u8) -> Result<(MouseEventKind, KeyModifiers)> {
     Ok((kind, modifiers))
 }
 
+#[cfg(feature = "bracketed-paste")]
+pub(crate) fn parse_csi_bracketed_paste(buffer: &[u8]) -> Result<Option<InternalEvent>> {
+    // ESC [ 2 0 0 ~ pasted text ESC 2 0 1 ~
+    assert!(buffer.starts_with(b"\x1B[200~"));
+
+    if !buffer.ends_with(b"\x1b[201~") {
+        Ok(None)
+    } else {
+        let paste = String::from_utf8_lossy(&buffer[6..buffer.len() - 6]).to_string();
+        Ok(Some(InternalEvent::Event(Event::Paste(paste))))
+    }
+}
+
 pub(crate) fn parse_utf8_char(buffer: &[u8]) -> Result<Option<char>> {
     match std::str::from_utf8(buffer) {
         Ok(s) => {
@@ -773,6 +790,15 @@ mod tests {
             Some(InternalEvent::Event(Event::Key(KeyCode::Delete.into()))),
         );
 
+        // parse_csi_bracketed_paste
+        #[cfg(feature = "bracketed-paste")]
+        assert_eq!(
+            parse_event(b"\x1B[200~on and on and on\x1B[201~", false).unwrap(),
+            Some(InternalEvent::Event(Event::Paste(
+                "on and on and on".to_string()
+            ))),
+        );
+
         // parse_csi_rxvt_mouse
         assert_eq!(
             parse_event(b"\x1B[32;30;40;M", false).unwrap(),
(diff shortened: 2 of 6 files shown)


src/input.rs
```diff
@@ -0,0 +1,49 @@
...
+
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[derive(Debug, PartialOrd, PartialEq, Eq, Clone, Hash)]
+pub enum Input {
+    Event(Event),
+    Paste(String),
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct EnableBracketedPaste;
```
groves, line 23: I put these commands in here as they're only applicable if you're using the read function in this module. It's probably worth panicing or something like that if Paste event is emitted when you've called the `event::read` function as that means you're mixing the systems. Not the greatest API, but I don't see a way to do better without breaking compatibility.

src/event/sys/unix/parse.rs
```diff
@@ -527,6 +533,18 @@ fn parse_cb(cb: u8) -> Result<(MouseEventKind, KeyModifiers)> {
...
 }
 
+pub(crate) fn parse_csi_bracketed_paste( buffer: &[u8], ) -> Result<Option<InternalEvent>> {
+    // ESC [ 2 0 0 ~ pasted text ESC 2 0 1 ~
+    assert!(buffer.starts_with(b"\x1B[200~"));
+
+    if !buffer.ends_with(b"\x1b[201~") {
+        Ok(None)
+    } else {
+        let paste = String::from_utf8_lossy(&buffer[6..buffer.len() - 6]).to_string();
```
groves, line 543: This only works because [the parser is backed by a Vec buffer](https://github.com/crossterm-rs/crossterm/blob/069497b43b6b58b6ff2f3824850b403addf64cae/src/event/source/unix.rs#L181). It'll keep growing till the end of the bracket shows up. Maybe we should put a limit on the max paste? Not sure what that should be though.

src/input.rs
```diff
@@ -0,0 +1,49 @@
+use std::fmt;
+use crate::event::filter::{EventFilter};
+use crate::{csi, Command, Result};
+use crate::event::{Event, InternalEvent, read_internal};
+
+pub fn read() -> Result<Input> {
```
TimonPost, line 6: In the past, we thought about why we call something an input or an event. We concluded that events are more general and include input, resize events and more. In this scenario, an input contains an event, but is the resize event an input?

src/input.rs
TimonPost: Can you explain what the problem is? Like why would you need panicking or change event when you've called read?

src/event/sys/unix/parse.rs
TimonPost: Lets just allow pasting anything. There is eventually a limit on the clipboard size. Time will tell if we should place a more practical limit

TimonPost:
I think we should try to stick to one read call and one poll function and the Event rather than Input as a top-level structure that is passed to the user. Mixing input and events can be confusing for our users.

TimonPost:
I wonder if it is really a 'priority' to keep 'copy'. Maybe the users of this library should answer this. I will ask in the discord.

src/input.rs
groves: If you've used EnableBracketedPaste and then you call `event::read` instead of `input::read`, you'll never get the paste input you requested. It's programmer error to mix the two. I want some way to let them know there's a mismatch instead of having it silently fail. Maybe there's a better way?

src/event/sys/unix/parse.rs
groves: Works for me. I was a little worried about memory usage since the source module is being so careful with allocations. I'm happy to not worry about it.

groves:
> I prefer to stick with one read call and one poll function and the Event rather than Input as a top-level structure that is passed to the user. Mixing input and events can be confusing for our users.

Same here. I'm only introducing this weird structure to keep API compatibility.

> I wonder if it is really a 'priority' to keep 'copy'. Maybe the users of this library should answer this. I will ask in the discord.

At least in Helix we're passing Events on the stack all over the place. I tried doing this with Copy removed and it made for a pretty massive change in Helix.

I can make that change for Helix, but I didn't want to force it on all other library users. Thanks for asking!

sigmaSd:
I also vote for just breaking the api because:

- crossterm is still not 1.0, I think we should use this opportunity to keep iterating on the api
- the api introduced here would make crossterm very confusing to new users (should I use read from event module or input module , what  is the difference)

TimonPost:
Indeed. It is a bit unfortunate to the API that we require `.clone()` calls. Wonder if there is an other way that doesnt duplicate the api and doesnt require the removal of `Copy`

Cargo.toml
```diff
@@ -27,6 +27,7 @@ all-features = true
 #
 [features]
 default = []
+event-copy = []
```
TimonPost, line 30: Can you add this to the default features

TimonPost (commented):
I think we should defaulty have Copy:

´´´
#[cfg_attr(not(feature= "bracketed-paste", derive(Copy))]
#[derive(Debug, PartialOrd, PartialEq, Eq, Clone, Hash)]
pub enum Event { }
´´´

This way copy is always enabled unless you use the bracketed paste feature flag.

Also 

´´´
#[cfg(not(feature = "event-copy"))]
#[derive(Debug, Clone, Copy, PartialEq, Eq)]
pub struct EnableBracketedPaste;
´´´

can be

´´´
#[cfg(feature = "bracketed-paste")]
#[derive(Debug, Clone, Copy, PartialEq, Eq)]
pub struct EnableBracketedPaste;
´´´

lnicola:
>  `#[cfg_attr(not(feature= "bracketed-paste", derive(Copy))]`

Features are supposed to be additive, so enabling a feature shouldn't cause an impl to disappear.

TimonPost:
Yes indeed, it is a bit of a draw back with this method. We have to pick: Remove copy resulting in us having to call .clone() everywere which is annoying. Dont add this feature. Or change the entire API. Or do a mix and just add it as feature flag and mis use feature flags which is not ideal but has a bit of both sides.

lnicola:
Do most of the `crossterm` users (besides Helix) copy `Event`s that much?

sigmaSd:
I think the new cfg bracked-paste is a nice idea, those who wants to keep the old clone trait can just opt out of it.

That will certainly make the breaking change easier.

groves:
Sorry for the noise from all the half-complete pushes on this PR. I was getting it to run the Windows tests for me. I should be done with changes now.

I was thinking we could make a release with this `bracketed-paste` feature flag in place with a note in the changelog that it'll be removed in a future release. Since it's on by default, that'll get other people using `Copy` to notice the compilation breakage. They can turn off `bracketed-paste` to get compilation going immediately, and then remove or rework the use of `Copy` at their leisure.

It's still an obnoxious change, but hopefully that makes it less obnoxious.

TimonPost:
Sounds good to me.

TimonPost (approved):
Looks good now!

PR #696, opened 2022-08-01
Add extra modifiers/state from kitty keyboard protocol

pianohacker:
Add support for:
* The Super/Command/Windows modifier.
* The Hyper/Meta modifers (stored in the `state` field as discussed in a previous PR).
* The Num Lock/Caps Lock state.

Please especially review the solution for Hyper/Meta.

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -536,7 +536,7 @@ pub enum MouseButton {
 bitflags! {
     /// Represents key modifiers (shift, control, alt, etc.).
     ///
-    /// **Note:** `SUPER` can only be read if
+    /// **Note:** `SUPER`, `HYPER`, and `META` can only be read if
     /// [`KeyboardEnhancementFlags::DISAMBIGUATE_ESCAPE_CODES`] has been enabled with
     /// [`PushKeyboardEnhancementFlags`].
     #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
@@ -545,6 +545,8 @@ bitflags! {
         const CONTROL = 0b0000_0010;
         const ALT = 0b0000_0100;
         const SUPER = 0b0000_1000;
+        const HYPER = 0b0001_0000;
+        const META = 0b0010_0000;
         const NONE = 0b0000_0000;
     }
 }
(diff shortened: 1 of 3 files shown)


src/event.rs
```diff
@@ -555,10 +560,33 @@ pub enum KeyEventKind {
...
+    /// [`PushKeyboardEnhancementFlags`].
     #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
     pub struct KeyEventState: u8 {
         /// The key event origins from the keypad.
         const KEYPAD = 0b0000_0001;
+        /// The Hyper modifier was held for this key event.
+        ///
+        /// This modifier is uncommon; thus, it is not stored in `KeyEvent.state` to avoid
+        /// confusion or failures to match key events.
+        const MODIFIER_HYPER = 0b0000_0010;
```
groves, line 575: I've always heard hyper to be the combination of all mod keys i.e. all of shift, control, alt, and super pressed at once. See https://docs.folivora.ai/docs/1004_hyper_key.html or https://kaleidoscope.readthedocs.io/en/latest/NEWS.html#hyper-and-meh-keys for examples of that interpretation. https://askubuntu.com/questions/19558/what-are-the-meta-super-and-hyper-keys shows where it used to exist on an actual keyboard, but these modern usages have muddied the water.

Given the lack of modern keyboard support for hyper and meta, I'd vote for not including these. Having some mods be in KeyEventState and some in KeyModifiers feels confusing to me.

groves (commented):
Excited to see the addition of super! Always wanted to be able to use that from the terminal.

src/event.rs
pianohacker: I'd be fine with that. Didn't love my solution either. @TimonPost ?

TimonPost: If they are uncommon you could leave them out. I never used them but that should not be the leading factor. I agree it might be confusing to have modifier keys at two places. Also" "to avoid confusion or failures to match key events.", is this a real problem we face? In the sense will people not reading those keys have a harder time if it is within the `state` field?

TimonPost: Maybe just add them to the `KeyModifiers` even if they are uncommen.

pianohacker: Will do. I'd prefer that to having them in the state.

TimonPost (approved):
Looks awesome!!

PR #697, opened 2022-08-01
Add OSC 52 clipboard setting command and a Windows equivalent

groves:
I'm using OSC 52 in https://github.com/helix-editor/helix/pull/3220. I'd like to pull it up into crossterm since it's a generally useful thing.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,6 +49,7 @@ features = ["winuser", "winerror"]
 
 [target.'cfg(windows)'.dependencies]
 crossterm_winapi = "0.9"
+clipboard-win = { version = "4.4.2", features = ["std"] }
 
 #
 # UNIX dependencies
@@ -58,6 +59,7 @@ libc = "0.2"
 mio = { version = "0.8", features = ["os-poll"] }
 signal-hook = { version = "0.3.13" }
 signal-hook-mio = { version = "0.2.3", features = ["support-v0_8"] }
+base64 = "0.13.0"
 
 #
 # Dev dependencies (examples, ...)
--- a/examples/set-clipboard.rs
+++ b/examples/set-clipboard.rs
@@ -1,0 +1,54 @@
+//! Demonstrates how to set the clipboard.
+//!
+//! cargo run --example set-clipboard
+
+use std::io::stdout;
+
+use crossterm::{
+    event::{read, Event, KeyCode, KeyEvent},
+    execute,
+    terminal::{disable_raw_mode, enable_raw_mode, ClipboardKind, SetClipboard},
+    Result,
+};
+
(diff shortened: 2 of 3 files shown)


sigmaSd:
I think it should be behind a compile time flag since it adds new dependencies.

Does this work on wayland or only on x11?

groves:
> I think it should be behind a compile time flag since it adds new dependencies.

Forgive my lack of Rust knowledge, but is that generally what one does for new dependencies on a library? Or is it because in this case it's for a feature that not everyone will use?

On the corresponding Helix issue, it was suggested that we [add a simple base64 implementation rather than pulling in a dependency](https://github.com/helix-editor/helix/pull/3220#issuecomment-1201788480). I could do that here if it seems reasonable.

> Does this work on wayland or only on x11?

It does work on Wayland: https://wiki.gnome.org/Initiatives/Wayland/PrimarySelection. Maybe I should say "Linux" instead of X11? Not sure what the correct general term is.

sigmaSd:
It's not a general rust thing, it's just crossterm has strived to keep minimal dependencies, there is even a section on the readme explaining what each dependency is for.

Also projects using crossterm so far will pay the cost for new dependencies while not using the new feature.
I guess this is more relevant to crossterm since its a lowish level library and if we start acumlating dependencies at this level it will make the chain bloated.

If the manual implementation of base64 has no drawback I would vote for it.

Regarding terminology I'm not sure either, my opinion is if it works on wayland let's just remove x11 wording because its usually indicates that it relies on x11 server

TimonPost:
Thanks for opening a PR, and creating cool features. Yes, upto now very few people have requested the feature for over 4 years. So adding a feature that is an edge case and includes new dependencies for just that feature is a bit too much overhead. 

On another note, crossterm aims to be minimal in what functionality it provides. It provides on the nessesairily API for interfacing with the terminal. We might argue clipboard is not really something we want for crossterm. While it is a fun feature to have I do believe we might want to keep it in user space rather than in crossterm.

groves:
Makes sense to me. Closing!

One question that comes up now that I'll be putting this directly into Helix. The thing that got me to think about creating a Command was that I didn't see an obvious Command for printing arbitrary terminal codes. I [initially used crossterm::style::Print](https://github.com/groves/helix/blob/8a5c873fa77969fd09f89b5d5ceb760b6ff30799/helix-view/src/clipboard.rs#L212), but that seemed like a misuse of something for styling. Is there a better builtin command for writing arbitrary terminal codes?

TimonPost:
You can perfectly build your own commands in helix as you wish. The ´Command´ trait is exposed and you can just create an implementation for that as you desire.

PR #703, opened 2022-08-14
Derive strum `EnumIter` type for `Color`

Piturnah:
I found myself wanting to iterate over the variants of `Color`, thought this change might be useful upstream. Since most people will not need this I put it in a new optional feature.

I figure this might be controversial as adding an optional feature is potentially a big deal, so please let me know any feedback.

EDIT: if needed I can add the derive for other enums in the crate too to make proper use of the feature, as well as update [feature flags](https://github.com/crossterm-rs/crossterm#feature-flags) section of the readme. Didn't do it yet since I figured this could be an unlikely merge anyway 😅

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,6 +29,7 @@ all-features = true
 default = ["bracketed-paste"]
 bracketed-paste = []
 event-stream = ["futures-core"]
+strum = ["dep:strum", "strum_macros"]
 
 #
 # Shared dependencies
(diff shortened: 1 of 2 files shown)


Piturnah:
Example usage: https://github.com/Piturnah/tcols

TimonPost:
Thanks for filing the PR. If were to go through with this, I would indeed prefer having it directly for all enum structures in the crate. I do wonder how often people want to iterate through colors. Feels like a bit of an edge-case scenario. 

Why do you think it's a big deal?

Piturnah:
Big deal may have been poor phrasing, but I figured that adding a crate feature may mean more to maintain and as you say it is a bit of an edge case scenario.

TimonPost:
Indeed, let's not get it in, I dont think it's a useful feature to support. Tho tcols is a nice library to view the potential supported color schemes, good work!

PR #705, opened 2022-08-14
Add commas after commands in doc

Yusuto:
I found some missing and one misplaced commas on the main page of the Crossterm documentation, that I think are probably meant to be different.
I didn't test the changes because to my knowledge I really just added and removed commas, if that breaks something I would be *very* surprised.


Yusuto:
Actually I ended up still testing everything using `cargo doc --open`, `cargo test` and `cargo build --release` and everything seems to have worked fine, no errors or warning in sight!

TimonPost (approved):

PR #711, opened 2022-09-06
Use `select()` instead of `mio` polling in Unix EventSource

### LINKED ISSUE #500: /dev/tty does not work on macOS with kqueue

yyogo:
I've run into #500 (which is caused by https://github.com/tokio-rs/mio/issues/1377) several times now, so here is a shot at reimplementing `EventSource` using `select()` instead of `mio` as discussed in the issue. feedback is appreciated, I tested it a bit and it seems to work fine, but the code could use some cleanup.

I thought about using [`nix`](https://docs.rs/nix/latest/nix/sys/select/fn.select.html) but decided against adding another dependency since using `libc::select()` is pretty straightforward.

Fixes #500, unblocks #407

### DIFF
--- a/src/event/source/unix.rs
+++ b/src/event/source/unix.rs
@@ -5,6 +5,7 @@ use std::{collections::VecDeque, io, os::unix::net::UnixStream, time::Duration};
 use signal_hook::consts::SIGWINCH;
 use signal_hook::low_level::pipe;
 
+use crate::event::timeout::PollTimeout;
 use crate::event::Event;
 use crate::Result;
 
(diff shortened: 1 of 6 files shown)


src/event/source/unix.rs
```diff
@@ -45,128 +62,98 @@ impl UnixInternalEventSource {
...
-                                }
-                                _ => unreachable!("Synchronize signal registration & handling"),
-                            };
+                        Err(e) => {
+                            match e.kind() {
+                                // No more data to read at the moment. We will receive another event
+                                // once more data is available to read.
+                                io::ErrorKind::WouldBlock => break,
+                                io::ErrorKind::Interrupted => continue,
+                                _ => return Err(e),
```
yyogo, line 115: This branch was missing, I believe it was a mistake, but this could result in different behavior

yyogo:
So unfortunately this solution is partial and will only work with fds < 1024. There is a [workaround](https://code.saghul.net/2016/05/libuv-internals-the-osx-select2-trick/) for Darwin, I'm working on implementing it , but it won't be pretty. And it would mean using `poll/epoll()` on Linux and only falling back to `select` on macOS, so I'm wondering if we're better off keeping `mio` and running an auxiliary thread that polls the tty.

TimonPost:
> This solution is partial and will only work with fds < 1024

What does this imply?

>  I'm wondering if we're better off keeping mio and running an auxiliary thread that polls the tty

If the potential workaround is a dirty one, I would argue to stay with mio, and hope macOS supports tty in kqueue. Tho if you argue the workaround outweighs the benefits of switching to poll, please provide the arguments why you think so, and then we can consider continuing the work here.

yyogo:
Hey @TimonPost , I'll try to summarize the situation as best I can.


As you said, the core issue is that `kqueue` does not support ttys. This is a [long standing issue](https://lists.apple.com/archives/darwin-dev/2006/Apr/msg00064.html) and I doubt it will be fixed any time soon, so I believe having a proper workaround is a good idea. `poll` and `epoll` are implemented in macos using `kqueue`, however `select` uses the older legacy implementation which does work with ttys. `mio` also uses `kqueue` and [will not implement a workaround](https://github.com/tokio-rs/mio/issues/1377#issuecomment-962452088).

So the idea behind this PR was to switch both the Linux and macOS implementations to use `select`. However, it being a rather old syscall, it only support fds that are less than 1024 in value:

> WARNING: select() can monitor only file descriptors numbers that
>       are less than FD_SETSIZE (1024)—an unreasonably low limit for
>       many modern applications—and this limitation will not change.
>       All modern applications should instead use [poll(2)](https://man7.org/linux/man-pages/man2/poll.2.html) or [epoll(7)](https://man7.org/linux/man-pages/man7/epoll.7.html),
>       which do not suffer this limitation.
[(from Linux's `select(2)` man page)](https://man7.org/linux/man-pages/man2/select.2.html)

The implication of this limitation is that apps that open more than 1024 files will not be able to use the event stream **after** those files have been opened, since `/dev/tty/` could potentially get a file descriptor larger than 1024. a hypothetical example of such a scenario would be an editor that is started with many filenames as arguments, and opens them before initializing the UI. Adding this limitation in macOS seems reasonable, since it won't break existing applications, however for linux it is a regression.

(In MacOS, there is a way to work around this limitation by compiling with different compile-time flags,  which causes a different version of `select` to be linked, which can support arbitrary sized FD sets. However this is not implemented in `libc` or any other rust library that I could find.)


In short, we are left with the following options as I can tell:
1. Use `select` universally and break fds >= 1024 in Linux [(this is what rustty does)](https://github.com/cpjreynolds/rustty/blob/master/src/core/terminal.rs#L581)
2. Implement `poll/epoll` for Linux and `select` for macOS
3. A hybrid solution with `mio` (e.g spinning an extra thread on mac that polls using select and wakes mio)
4. Do nothing

I'm willing to put in the work, just give me the go ahead on what you think is the best choice. I am leaning towards option 2, however there might be something I'm missing.

Cheers

TimonPost:
I like the solution two as well. 
1) it uses the proper not deprecated api's on the platforms that can have problems with the deprecated one 
2) Using select on macos while doing the work around you speak of
3) Removing mio which is quite a large dependency (eventho we will have to reinvent the weel a bit). 

Perhaps, we could do a smooth transition by just creating a new event source like you did, and exposing it behind a feature flag. Just needs good manual testing on various platforms if we are going to do the polling our selfs. Atm i only have windows, but soon, I will possess a macbook. Linux for me is only accessible trough wsl 2. What platforms can you test on?

conradludgate:
I'm trying to use crossterm in my tui app and this is a blocker for me, so I'm happy to help test. I have access to Linux and Mac environments

conradludgate:
I already am in testing. It works great for me right now (using select). I won't merge it like this though

sigmaSd:
@yyogo thanks for your work, any update on this? is this still a draft or considered done?
This is currently probably the most requested feature probably in crossterm by unblocking https://github.com/crossterm-rs/crossterm/pull/407/files

sigmaSd:
Also my personal opinion is to use select for macos *without* the hack.

I think as a fist cut that would be a nice step and the limitation seems theoretical. We can improve things when it gets tested in real world and people complain.

yyogo:
@sigmaSd this is still a draft, but I recently found [filedescriptor's poll](https://docs.rs/filedescriptor/latest/filedescriptor/fn.poll.html), which is used in `termwiz` and implements the polling logic I intended to. I already have a local branch with it and am currently doing some cleanups and preparing to open a new PR with it, expect it within a few days.

PR #714, opened 2022-09-20
Make Windows resize event match `terminal::size`

### LINKED ISSUE #713: Resize events yield different results than `terminal::size` on Windows
**Describe the bug**
On Windows, the Resize event yields results that are 1 less than the results from `terminal::size`.

**To Reproduce**
1. Use crossterm on Windows
2. Call `terminal::size`; store results 
3. Enter raw mode so events are processed
4. Enter alternate screen so that Resize events are fired properly (required on CMD and PowerShell, not required on Windows Terminal)
5. Call `event::read` and store the results of the first `Event::Resize`
6. Leave alternate screen (if it was entered) and disable raw mode 
7. Compare results from `terminal::size` and `Event::Resize`

**Expected behavior**
Results from Resize events match results from `terminal::size`

**OS**
 Windows 10 (10.0.19044 Build 19044)

**Terminal/Console**
CMD, PowerShell, and Windows Terminal

generic-user1:
Previously, results from a Resize event were 1 less than results from `terminal::size` in both x and y dimensions.

This PR fixes #713

### DIFF
--- a/src/event/source/windows.rs
+++ b/src/event/source/windows.rs
@@ -64,7 +64,8 @@ impl EventSource for WindowsEventSource {
                             mouse_event
                         }
                         InputRecord::WindowBufferSizeEvent(record) => {
-                            Some(Event::Resize(record.size.x as u16, record.size.y as u16))
+                            // windows starts counting at 0, unix at 1, add one to replicate unix behaviour.
+                            Some(Event::Resize(record.size.x as u16 + 1, record.size.y as u16 + 1))
                         }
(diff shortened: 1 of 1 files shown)


TimonPost (approved):
Oh noes, good catch!

TimonPost:
Can you run cargo fmt?

generic-user1:
Yes! Added a commit after running cargo fmt

eminence:
Any guess on when this fix might be released on crates.io?

PR #716, opened 2022-10-03
Fix parsing of event types/modifiers with certain keys

pianohacker:
When keyboard enhancement is enabled, release/repeat events for certain keys (including Page Up, Page Down, Home, End, etc) are incorrectly parsed as press events.


TimonPost (approved):
good to have this fallback

PR #721, opened 2022-10-13
Add Chrome OS Crostini to tested terminals list

CenTdemeern1:
I have never had any issues with Crossterm on Crostini and I didn't even know it wasn't tested. Well, here you go.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,6 +72,7 @@ WARNING: Do not change following heading title as it's used in the URL by other 
 - (Arch) Kitty
 - Linux Mint
 - (OpenSuse) Alacritty
+- (Chrome OS) Crostini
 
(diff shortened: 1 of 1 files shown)


CenTdemeern1:
If there are any specific tests you'd like me to do to ensure 100% compatibility, let me know, I'd be more than willing to try it out for you.

CenTdemeern1:
[Screen recording 2022-10-23 04.30.34.webm](https://user-images.githubusercontent.com/24682280/197370322-5795f72a-9d82-4d0a-9faf-101ca799a2a2.webm)
The interactive demo works fine. The only thing I have noticed is that the cursor blink doesn't work, but that's because the preference setting in the terminal settings overrides it, and it's not an issue with crossterm.

TimonPost (approved):
That is fine, thanks for testing!

PR #732, opened 2022-11-26
Add a function for checking keyboard enhancement support

### LINKED ISSUE #717: Allow checking for keyboard enhancement support
The new keyboard enhancement feature is supported by a small (if growing) subset of terminals. It would be nice to know if the current terminal supports the protocol. This allows adjusting expected keyboard inputs, changing shortcuts,  etc.

The [suggested way](https://sw.kovidgoyal.net/kitty/keyboard-protocol/#detection-of-support-for-this-protocol) to check for support is to:
- send a query escape sequence for the current keyboard enhancement status
- send a query for the primary device attributes
- check if we got two answers, or just one

I'm not immediately sure a) how to reliably do this and b) how we could provide a helper in crossterm itself. Desired workflow would just be something like:

```rust
if crossterm::event::detect_keyboard_enhancement_support(&mut stdout)? {
...
```

I don't _think_ there's any other way to detect support. Enabling keyboard enhancement then checking the form of incoming events is possible, but that seems messy and unreliable.

the-mikedavis:
This follows the Kitty documentation's recommended way to check for progressive keyboard enhancement: query the flags and then query the primary device attributes (which is broadly supported). If we receive the response to the flags query, we return the flags. Otherwise, the terminal must not support progressive keyboard enhancement.

Closes #717

### DIFF
--- a/src/terminal/sys/unix.rs
+++ b/src/terminal/sys/unix.rs
@@ -1,7 +1,9 @@
 //! UNIX related logic for terminal manipulation.
 
 use std::fs::File;
+use std::io::Write;
 use std::os::unix::io::{IntoRawFd, RawFd};
+use std::time::Duration;
 use std::{io, mem, process};
 
 use libc::{
@@ -11,7 +13,9 @@ use libc::{
 use parking_lot::Mutex;
 
 use crate::error::Result;
+use crate::event::filter::{KeyboardEnhancementFlagsFilter, PrimaryDeviceAttributesFilter};
 use crate::event::sys::unix::file_descriptor::{tty_fd, FileDesc};
+use crate::event::{poll_internal, read_internal, InternalEvent};
 
 // Some(Termios) -> we're in the raw mode and this is the previous mode
 // None -> we're not in the raw mode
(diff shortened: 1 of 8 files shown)


src/terminal/sys/unix.rs
```diff
@@ -88,6 +92,65 @@ pub(crate) fn disable_raw_mode() -> Result<()> {
...
+
+fn read_supports_keyboard_enhancement_raw() -> Result<bool> {
+    // This is the recommended method for testing support for the keyboard enhancement protocol.
+    // We send a query for the flags supported by the terminal and then the primary device attributes
+    // query. If we receive the primary device attributes response but not the keyboard enhancement
+    // flags, none of the flags are supported.
+    //
+    // See <https://sw.kovidgoyal.net/kitty/keyboard-protocol/#detection-of-support-for-this-protocol>
+
+    let mut stdout = io::stdout();
```
TimonPost, line 122: wonder if there are usecases when we need to read from an other file descriptor?

TimonPost (approved):
Awesome, this is very useful to have 🎊

TimonPost (commented):
Oh what, had not yet pressed submit review, whoops.

TimonPost (approved):

PR #735, opened 2022-12-12
Replace mio polling with filedescriptor's poll()

### LINKED ISSUE #500: /dev/tty does not work on macOS with kqueue

yyogo:
Dince mio doesn't handle ptys correctly macOS, use filedescriptor's poll() which falls back to select().

This PR replaces #711.

Fixes #500, unblocks #407.

## Implementation notes

Since poll() doesn't support waking from signals or otherwise, I create two unix socket pairs, one which is written to on `SIGWINCH`  and one which is used by the `Waker`, and poll on those as well as the tty fd.


src/event/source/unix.rs
```diff
@@ -45,128 +68,128 @@ impl UnixInternalEventSource {
...
-
-                            if let Some(event) = self.parser.next() {
-                                return Ok(Some(event));
-                            }
-                        }
+            match poll(&mut fds, timeout.leftover()) {
+                Err(filedescriptor::Error::Io(e)) => return Err(e),
+                res => res.expect("polling tty"),
+            };
+            if fds[0].events & POLLIN != 0 {
```
YS-L, line 139: Should these conditions be checking for `revents` since that is the output indicator of the events that occurred whereas `events` is an input parameter (based on [1] and [2])?

[1] https://docs.rs/filedescriptor/latest/filedescriptor/fn.poll.html
[2] https://man7.org/linux/man-pages/man2/poll.2.html

yyogo: you're right, good catch! The incorrect code works because it just always assumes there is input when the poll returns.

YS-L:
As a data point, I tested this with `csvlens` on multiple platforms (linux, macOS, Windows) and it works like a charm. It fixes the tty issue on macOS (https://github.com/YS-L/csvlens/issues/20). Thank you for working on this!

TimonPost (commented):
Gonna give it a go on mac os now! Unamomento. Wonder if we should consider having a seperate `EventSource` asside from mio event source. Then it should be a matter of a feature toggle to enable or disable this behavior. This might be the safe way to transition away from mio. Do you think this might be worth the effort?

TimonPost:
Also validated this works on macos

TimonPost (approved):
Lets get this merged. Looks good! Thanks for the contribution

yyogo:
@TimonPost I agree a feature flag sounds like a safe way to introduce this change. Maybe the flag could be used to introduce #407 as well.

yyogo:
I added the feature flag though it made the PR a bit messy. let me know if you want to merge it as-is, undo the feature gate, or something else

TimonPost:
Yes this is good. WIll do a review asap

Cargo.toml
```diff
@@ -29,6 +29,7 @@ all-features = true
 default = ["bracketed-paste"]
 bracketed-paste = []
 event-stream = ["futures-core"]
+use-dev-tty = ["filedescriptor"]
```
TimonPost, line 32: dont we want to make a freature flag: mio and refer to the mio dependency, and make this enabled by default?

src/event/source/unix.rs
```diff
@@ -1,241 +1,11 @@
-use std::{collections::VecDeque, io, time::Duration};
+#[cfg(feature = "use-dev-tty")]
+pub(crate) mod tty;
 
-use mio::{unix::SourceFd, Events, Interest, Poll, Token};
-use signal_hook_mio::v0_8::Signals;
+#[cfg(not(feature = "use-dev-tty"))]
```
TimonPost, line 4: Then here we can use the mio feature flag

Cargo.toml
yyogo: which would you like to be enabled by default?

 I think default features are a bit of a pain to deal with since you can't disable them individually, so if we keep mio as the default by making it a default feature it would be troublesome for users to opt out, as well as possibly breaking. this seemed cleaner since it makes the new behavior opt in and the feature can be deprecated easily later. Also it's intended to include #407.

TimonPost: Good point. So then we temp can not opt out from mio but can opt in for the new feature.

TimonPost (approved):

TimonPost:
Great work @yyogo !

Issue #646, opened 2022-04-02
Add a function to get the cursor style (shape, blinking)


Cl00e9ment:
Currently, it's possible to modify the cursor shape and whether it's blinking or not, but the previous state can't be restored. A function that can retrieve the cursor style would be useful, because one could save the old style and restore it when the program exits, so users can keep their terminal the way they prefer it.

I don't know if this is possible. For now, [I forked the project](https://github.com/Cl00e9ment/crossterm/tree/feat-get-cursor-style) and I've a proof of concept that works only on Xterm compatible terminals using DECRQSS with DECSCUSR (see [this](https://invisible-island.net/xterm/ctlseqs/ctlseqs.html#h3-Device-Control-functions)).

What works:
- GNOME Terminal (Manjaro Linux)
- Xterm (Manjaro Linux)

What doesn't:
- CLion terminal (Manjaro Linux)
- cmd.exe (Windows 10)
- PowerSheel (Windows 10)
- iTerm2 (macOS)

I've 2 questions:
- Do anyone know how to implement something like that on other terminals?
- If it's not possible to make it work on other terminals, should this be included in crossterm?

sigmaSd:
According to this pr https://github.com/microsoft/terminal/pull/11152 requesting DECSCUSR via DECRQSS might land in a future pr.

Personally I think xterm should be the reference to judge if a feature should be added or not, windows terminal is catching up with the spec, and I imagine macos have other terminals where this work.

TimonPost:
Maybe for windows systems we could get some kind of idea by querying: https://docs.microsoft.com/en-us/windows/console/console-cursor-info-str

Cl00e9ment:
It's interesting to note that, on Windows (tested with cmd.exe and PowerShell), if the cursor style was modified in the alternate screen, it is automatically restored upon leaving the alternate screen. Maybe we should implement that on other terminals, when possible, to have an uniform behavior.

Cl00e9ment:
> Maybe for Windows systems, we could get some kind of idea by querying: https://docs.microsoft.com/en-us/windows/console/console-cursor-info-str.

Idk why, but whatever the shape of the cursor, the reported `dwSize` is always `25` (see [this](https://github.com/Cl00e9ment/crossterm/blob/d0f598e9318dcfd4915b94f39ce8897d7b0d1bf3/src/cursor/sys/windows.rs#L232)).

TimonPost:
[...]

Issue #740, opened 2023-01-09
DisableBlinking does not work for line cursor shape in Konsole


notfirefox:
**Describe the bug**
When using a line cursor shape, disable blinking does not work in Konsole.

**To Reproduce**
Compile and run this minimal example using Konsole: 
```rust
use crossterm::{cursor, event, execute, style, terminal};
use std::io::{stdout, Write};

fn main() -> Result<(), std::io::Error> {
    execute!(
        stdout(),
        terminal::EnterAlternateScreen,
        style::ResetColor,
        cursor::MoveTo(0, 0)
    )?;
    stdout().flush()?;
    terminal::enable_raw_mode()?;

    execute!(
        stdout(),
        cursor::SetCursorShape(cursor::CursorShape::Line),
        cursor::DisableBlinking,
    )?;

    loop {
        if let event::Event::Key(event::KeyEvent { code, .. }) = event::read()? {
            if code == event::KeyCode::Char('q') {
                break;
            }
        }
    }

    terminal::disable_raw_mode()?;
    execute!(stdout(), terminal::LeaveAlternateScreen, style::ResetColor)?;

    Ok(())
}
```

**Expected behavior**
The cursor shows up as a line and does not blink.

**OS**
 - Arch Linux

**Terminal/Console**
 - Konsole

notfirefox:
Related: #650

TimonPost:
Fixed by: https://github.com/crossterm-rs/crossterm/pull/742

PR #742, opened 2023-01-11
Add SetCursorStyle command

TimonPost:
Adds a command to set the cursor style.

Fixes: #646 , maybe #740

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,11 @@
+# Version 0.26.0
+## Added ⭐
+
+- ⚠️ Add `SetCursorStyle` to set the cursor style. 
+
+## Breaking
+
+- Remove `SetCursorShape` in vavour of `SetCursorStyle`.
 
 # Version 0.25.0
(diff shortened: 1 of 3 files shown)


Cl00e9ment:
I was just about to create a pull request but you were faster!

I don't know if it's useful, but here is my version:
https://github.com/Cl00e9ment/crossterm/tree/feat-set-cursor-style

TimonPost:
Oh whoops, what a coincidence, im sorry, let me look at that.

TimonPost:
I decided to keep enable / disable blinking. Those are very specific and a user might not know which cursor shape it uses. So allowing to have this separately may allow a user to control blinking for the current active cursor

Cl00e9ment:
> I decided to keep enable / disable blinking. Those are very specific and a user might not know which cursor shape it uses.

Yes but these commands are broken on certain terminals so a note should be added.

> If you want to follow up and improve there is time to do that!

I'll do that!

PR #744, opened 2023-01-11
SetCursorStyle improvements

Cl00e9ment:
Followup to #742

- Added a note to `EnableBlinking` and `DisableBlinking` to warn of cross-compatibility issues.
- Removed "It uses two types of escape codes, one to control blinking, and the other the shape." because it's false. One escape code controls shape+blinking and the other was just a fall-back for blinking.
- Removed the fall-back escape codes since during my tests on different terminals/OS a didn't find any occurrence were they would be needed.
- Updated the doc.

### DIFF
--- a/src/cursor.rs
+++ b/src/cursor.rs
@@ -328,13 +328,11 @@ impl Command for Show {
 }
 
 /// A command that enables blinking of the terminal cursor.
-///
-/// See `SetCursorStyle` which is more advanced and better supported.
-///
+/// 
 /// # Notes
 ///
 /// - Some Unix terminals (ex: GNOME and Konsole) as well as Windows versions lower than Windows 10 do not support this functionality.
-///   Use SetCursorStyle for better cross-compatibility.
+///   Use `SetCursorStyle` for better cross-compatibility.
 /// - Commands must be executed/queued for execution otherwise they do nothing.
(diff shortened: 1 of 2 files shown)


src/cursor.rs
```diff
@@ -354,7 +357,10 @@ impl Command for EnableBlinking {
 ///
```
TimonPost, line 357: And the line above here

src/cursor.rs
```diff
@@ -333,7 +333,10 @@ impl Command for Show {
...
```
TimonPost, line 333: can you remove the line above this one? Has a similar message

src/cursor.rs
```diff
@@ -394,13 +399,13 @@ pub enum SetCursorStyle {
 impl Command for SetCursorStyle {
     fn write_ansi(&self, f: &mut impl fmt::Write) -> fmt::Result {
         match self {
-            SetCursorStyle::DefaultUserShape => f.write_str("\x1b[?12h\x1b[0 q"),
```
TimonPost, line 397: I was thinking perhaps a little redundancy could not hurt. But actually since we do have Enable/DisableBlinking it is a bit uneasily.

TimonPost:
Nice!

Issue #695, opened 2022-07-30
Support key release events


jjant:
**Is your feature request related to a problem? Please describe.**
I'm making a piano-like UI and I need to detect key up events to know when to stop playing the note for that key.

**Describe the solution you'd like**
An additional field in `KeyEvent` for `Press/Release`.

**Additional context**
Just FYI, termion doesn't seem to provide this either.

sigmaSd:
It is is possible in master(not released) yet using the kitty protocol.
See https://github.com/crossterm-rs/crossterm/blob/master/src/event.rs#L357 it also contains the list of supported terminals.

To use it just add:
```rs
    execute!(stdout, PushKeyboardEnhancementFlags(KeyboardEnhancementFlags::REPORT_EVENT_TYPES))?;
```
and run it in a supported terminal.

David-Else:
This feature was release in 0.25, the issue can be closed.

TimonPost:
Its partially possible for those that have a kitty terminal

PR #745, opened 2023-01-11
Support key release events for windows.

TimonPost:
Adds support for key release event on windows.

@matkaas was there a reason that you added: https://github.com/crossterm-rs/crossterm/compare/timon/key-up-events-windows?expand=1#diff-0fc06b629fbd961b93f1888ab8546e9089e7359747d2fbe9e8b7aabd30983603L238? 

Fixes: https://github.com/crossterm-rs/crossterm/issues/695

### DIFF
--- a/src/cursor.rs
+++ b/src/cursor.rs
@@ -328,7 +328,7 @@ impl Command for Show {
 }
 
 /// A command that enables blinking of the terminal cursor.
-/// 
+///
 /// # Notes
 ///
 /// - Some Unix terminals (ex: GNOME and Konsole) as well as Windows versions lower than Windows 10 do not support this functionality.
--- a/src/cursor/sys/windows.rs
+++ b/src/cursor/sys/windows.rs
@@ -135,8 +135,7 @@ impl ScreenBufferCursor {
             return Err(io::Error::new(
                 io::ErrorKind::Other,
                 format!(
-                    "Argument Out of Range Exception when setting cursor position to X: {}",
-                    x
+                    "Argument Out of Range Exception when setting cursor position to X: {x}"
                 ),
             ));
         }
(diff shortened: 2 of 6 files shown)


matkaas:
Hi @TimonPost

I rather moved it than added it. The check used to happen here:
https://github.com/crossterm-rs/crossterm/pull/651/commits/0fcce0dcb431dc1e903b04e3cefa7260c28668b4#diff-0fc06b629fbd961b93f1888ab8546e9089e7359747d2fbe9e8b7aabd30983603L35

And I moved it where you linked:
https://github.com/crossterm-rs/crossterm/pull/651/commits/0fcce0dcb431dc1e903b04e3cefa7260c28668b4#diff-0fc06b629fbd961b93f1888ab8546e9089e7359747d2fbe9e8b7aabd30983603R115

The reason I moved it one function call further in was that I needed to check the (Alt) key release for alt+numpad key combinations:
https://github.com/crossterm-rs/crossterm/pull/651/commits/0fcce0dcb431dc1e903b04e3cefa7260c28668b4#diff-0fc06b629fbd961b93f1888ab8546e9089e7359747d2fbe9e8b7aabd30983603R94

I see no problem with removing it in principle, but beware that the alt+numpad key events from the terminal are sorta weird. The generated character gets delivered as part of the Alt key release event. I don't think the case here:
https://github.com/crossterm-rs/crossterm/pull/745/files#diff-0fc06b629fbd961b93f1888ab8546e9089e7359747d2fbe9e8b7aabd30983603R224
will ever happen for `key_down == true`.
(In particular also because we have a check for that in the if-expression of the block https://github.com/crossterm-rs/crossterm/commit/0fcce0dcb431dc1e903b04e3cefa7260c28668b4#diff-0fc06b629fbd961b93f1888ab8546e9089e7359747d2fbe9e8b7aabd30983603R96).

Since the alt+numpad character events are synthetically generated, they don't really have a press+release event pair, so it's hard to say how that is best handled. Perhaps synthesize both the press and release at the same time? Or simply accept that the character only gets a release event. It's an intrinsic problem anyhow that press+release events are not guaranteed to be delivered to applications in pairs, e.g. if a user presses down a key and moves focus away from the window before releasing the key, so that the press and release events go to two different windows. I also think the keyboard repeat function (holding down a key to repeatedly trigger it) will deliver several press events and one final release event, so not in pairs. I'm not sure if that applies to terminal applications though, since they're really a text-based protocol.

Cheers!

PR #751, opened 2023-01-30
Remove early return in keyboard enhancement check

the-mikedavis:
This block in the `if` sends the query over stdout if the query couldn't be sent over `/dev/tty` successfully. If we send it over stdout we should still query for the response though: the early return prevents that so the response to the query might end up not being consumed.

This fixes detection for me on Linux. I will test with macOS as well. (Edit: looks good on macOS too)


archseer:
Nice catch! Looks like it was broken in https://github.com/crossterm-rs/crossterm/pull/746

TimonPost (approved):
Ah nice one!

PR #765, opened 2023-03-18
Unwrap `crossterm::Result<T, ErrorKind>`  to `std::io::Result`.

gibbz00:
Introducing superficially wrapped Error types increases in most cases error handling complexity, both internally and externally.

https://github.com/fdehau/tui-rs/blob/fafad6c96109610825aad89c4bba5253e01101ed/src/backend/crossterm.rs#L125-L127

Generally, it becomes a nuisance every time one is working inside a function that returns a `Result<T, io::Error>`. Any expression that returns a `crossterm::Result` leads to a mismatched type compile time error, even if the errors are practically the same.

Unwrapping the type should also increase the transparency of the error origin.

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,7 +109,7 @@ use crossterm::{
     event,
 };
 
-fn main() -> Result<()> {
+fn main() -> std::io::Result<()> {
     // using the macro
     execute!(
         stdout(),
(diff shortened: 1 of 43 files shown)


TimonPost (approved):
Agree in the past we had some custom errors but as crossterm is close to io operations it should likely reflect that.

TimonPost:
code is not compiling

Issue #766, opened 2023-03-23
Put windows deps in (default-included) feature


jhartzell42:
Due to idiosyncrasies of how cargo resolves dependencies, all the Windows dependencies show up in Cargo.lock files even on Linux -- and in Yocto builds, which is making our build process more difficult. Features do not have the same issue.

There should be a "windows" feature, included by default to avoid breaking existing clients, that can be explicitly disabled if the windows dependencies are not going to be used. This will make building easier on Yocto and other build systems.

I am willing to (and going to) write a PR for this; I'm just adding an issue first for discussion in principle.

sigmaSd:
Is there an issue about this in cargo repo ?

jhartzell42:
https://github.com/rust-lang/cargo/issues/1197 is very close.

jhartzell42:
The fix I have is #767 , which takes my project from not building on Yocto to building on Yocto, I think without hurting anyone else's builds on Windows or Unix.

jhartzell42:
#767 being merged addresses the issue.

joshka:
Hey @jhartzell42

I'm looking at @TimonPost's comment on https://github.com/crossterm-rs/crossterm/pull/943/files#r1845623030 about the changes that you made here and trying to understand what the underlying problem was.

The windows deps appearing in the lockfile is intended cargo behavior, as the lockfile doesn't specify which dependencies to use, only which versions of them to use if called for by the manifest. The **`Cargo.toml`** file is responsible for choosing which deps to use. The  previous configuration without the feature flag correctly gated the windows deps with conditions that only evaluate as true on windows. This works fine on my Mac to not enable the winapi stuff.

I'm curious why the contents of the lock file caused problems in your build process. Is Yocto doing something special?

joshka:
Digging a bit deeper https://github.com/meta-rust/cargo-bitbake/issues/58 suggests that removing the windows deps from the Yocto side is the correct thing to do, instead of marking the correctly specified deps as optional and behind a feature flag like this. It's possible I'm misunderstanding this in some way...

jhartzell42:
I don't work on the Yocto project anymore and don't have access to it and cannot troubleshoot, unfortunately.

joshka:
Thanks for the reply. My understanding is that the problem you saw was in generating package config right, i.e. dev / config time problems not a hard fail right? No problem if you can't recall.

PR #767, opened 2023-03-23
Make direct winapi deps optional

jhartzell42:
This is to make building easier on platforms like Yocto so we can actually remove it from the calculated dependencies. Would address #766 .

### DIFF
--- a/.github/workflows/crossterm_test.yml
+++ b/.github/workflows/crossterm_test.yml
@@ -63,8 +63,12 @@ jobs:
       run: cargo test --all-features -- --nocapture --test-threads 1
       continue-on-error: ${{ matrix.can-fail }}
     - name: Test no default features
+      if: matrix.os != 'windows-2019'
       run: cargo test --no-default-features -- --nocapture --test-threads 1
       continue-on-error: ${{ matrix.can-fail }}
+    - name: Test no default features with windows feature enabled
+      if: matrix.os == 'windows-2019'
+      run: cargo test --no-default-features --features windows -- --nocapture --test-threads 1
     - name: Test Packaging
       if: matrix.rust == 'stable'
       run: cargo package
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,7 @@ all-features = true
 # Features
 #
 [features]
-default = ["bracketed-paste"]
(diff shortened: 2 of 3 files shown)


TimonPost:
Not sure how I feel about that if you disable a feature flag you can break compilation. We should likely also update the places in code that use the windows crate and program some logic for what happens if it is disabled

I had this idea for some time, and its how it used to be in the past, but to put input/events behind a feature flag. Because on all > windows  10 systems that is one of the main user of the windows API. Maybe we also have to do something with getting cursor pos, and terminal size.

jhartzell42:
As this feature isn't designed to be disabled on Windows, perhaps it should just use the compile_error macro to explain that this feature is mandatory on Windows, giving a more explicit error rather than just failing to compile more haphazardly.

The idea would be that the feature flag would only be disabled on platforms that know they're only going to run on Unix-based platforms, as a hack to work around the Cargo constraint. I can't build a project that depends on this in Yocto without a change to make the crossterm_winapi and windows deps optional in some way.

If you would still rather a feature mapping that accomplishes this but leaves a binary able to build in Windows, I'm honestly not sure why -- I'd always want to build on Windows with these crates included -- but I can code that up too. I'd just need a little more detail on what you're thinking.

TimonPost:
Im fine with the first suggestion of adding a cleaner message when disabling 'windows' flag on windows. I am only concerned about devs disabling default features and then ending up with a bunch of compile errors without direct instructions on what causes it.

jhartzell42:
I added the appropriate error messages. I don't have a Windows machine to test generation of these errors, however. Could you help me confirm that these errors are in fact triggered if these features are left out on Windows?

TimonPost:
image

It seems to work correctly!

TimonPost (approved):

jhartzell42:
@TimonPost I made the formatting fix that the CI wanted, but now it apparently needs approval to run the workflow again.

jhartzell42:
@TimonPost I had to make a CI change because it was failing testing Windows with no features. Instead, I test it with no default features, but the Windows feature still enabled. Please re-run the work-flow! I hope I got it right!

TimonPost (approved):

Issue #772, opened 2023-04-01
Double key input on Windows


jasiukiewicztymon:
**Describe the bug**
When I use the event::read function I have double input, and it does like I have clicked a key two times. When I tried it on my WSL it worked..

**To Reproduce**
1. Use the Rust tui lib tab example
2. Run the code on windows
3. See error

**Expected behavior**
Fix this

**OS**
 - Windows

**Terminal/Console**
 - CMD
 - PowerShell
 - Windows Terminal

TimonPost:
Its like key press/release

jasiukiewicztymon:
Ok.. but why it doesn't occur on every terminal but only on Window's one?

TimonPost:
good question, and actually, I think this needs to be reverted. On UNIX systems its supported under the kitty protocol. But as its not supported all over it should likely be disabled by default and enabled behind feature flag

jasiukiewicztymon:
ok, waiting to be fixed so... But it would be a big change no?

TimonPost:
not really. In the meantime just match on the https://docs.rs/crossterm/latest/crossterm/event/struct.KeyEvent.html `kind` field. So you can easily always make sure you are expecting key-press only..

lesleyrs:
I was just trying out `event-poll-read` as well using both wezterm and windows terminal on windows, wezterm doesn't actually have the double input problem. But it still has the enter release event at the start.

However I noticed that on WSL, there aren't any release events on either terminal. Am I not understanding something here?

lesleyrs:
Hey I spend some time figuring this out and I explained it in my PR.

But to sum it up:

cmd on conhost works as intended, then any other combination of cmd, powershell, conhost and wezterm (maybe others) works but you get the unintended "release" event at the start.

The only one which doesn't work well is windows terminal regardless of shell which means you can't use it for detecting held keys. They might get it fixed eventually as well, it's being tracked here https://github.com/crossterm-rs/crossterm/pull/775#issuecomment-1498296749. But it still works fine as long as you check for either press or release.

lesleyrs:
@TimonPost this one is already open for the same issue https://github.com/crossterm-rs/crossterm/issues/752

TimonPost:
Closing as duplicate then

Issue #769, opened 2023-03-27
Key stroke listener, windows vs mac os


lcmgh:
Hi!

I am using this code taken from the `tui` crate examples https://github.com/fdehau/tui-rs/blob/master/examples/demo/crossterm.rs#L45 to listen for Key events. However the behavior on Mac OS vs. Windows is different. While on MacOS one can easily navigate Windows users report that when they type a key they move across two menu entries instead of one.

Do you have any idea why this is the case?

Thanks

sigmaSd:
https://github.com/crossterm-rs/crossterm/issues/752

lesleyrs:
This one is more related to https://github.com/crossterm-rs/crossterm/issues/772#issuecomment-1498380935.

You need to check for KeyEventKind, either press or release. I think currently mac and linux only show presses right now?

lcmgh:
My problem was solved by downgrading the version to 0.25.0.

Issue #759, opened 2023-02-15
Windows 10 bugs from interactive-demo and examples


lesleyrs:
**Describe the bug**
- [ ] after closing the `interactive-demo` 1st `cursor` option it doesn't restore cursor correctly and stays as underline or block. https://github.com/crossterm-rs/crossterm/pull/775
- [ ] In the `is_tty` example Windows terminal does not resize screen. This messes up the screen until you resize the window and press enter as it thinks it's resized. https://github.com/microsoft/terminal/issues/5094 (not gonna bother tracking)
- [ ] the 3rd `attribute` option does not load at all in Windows terminal (it just blinks) but it works on the normal Console host. https://github.com/crossterm-rs/crossterm/pull/775
- [ ] This one isn't related to Windows terminal, but Console host seems to have no delay between key presses. Meaning that you can't view 1 screen at a time and it skips over them in 2nd `color` option for example. https://github.com/crossterm-rs/crossterm/pull/775
- [ ] Both on windows terminal and console host `tab` for some reason works as arrow keys. https://github.com/crossterm-rs/crossterm/pull/775 (can't reproduce)

**To Reproduce**
1. `cargo run` in interactive-demo directory, try the examples

**Expected behavior**
above

**OS**
Windows 10

**Terminal/Console**
Windows terminal

Issue #752, opened 2023-02-03
Strange input event on windows


C-BJ:
**Describe the bug**
We had a strange problem developing REPL for the Erg language: https://github.com/erg-lang/erg/pull/362

When we run a program through the command line, we get an Enter event first, and then whatever we type, it gets two of these input events

This bug only appears at `0.26.0`, not `0.25.0`
This issue only appears on Windows (not WSL)

**OS**
 - Windows10 / 11

aschey:
Looks like this was caused by this PR: https://github.com/crossterm-rs/crossterm/pull/745 and the fix is to check the `KeyEventKind` first:
```rust
if let Event::Key(key) = event {
    if key.kind == KeyEventKind::Press {
        ...
    }
}
```

TimonPost:
Thats possible, does it have the event_kind set?

dsherret:
This bug only occurs in crossterm 0.26.0 and above.

**To Reproduce**

```rs
pub fn main() {
  crossterm::terminal::enable_raw_mode().unwrap();
  let result = crossterm::event::read();
  eprintln!("RESULT: {:?}", result);
}
```

This will immedately return the enter key press:

```
> cargo run
    Finished dev [unoptimized + debuginfo] target(s) in 0.03s
     Running `target\debug\rust-test.exe`
RESULT: Ok(Key(KeyEvent { code: Enter, modifiers: NONE, kind: Release, state: NONE }))
```

**Expected behavior**

It should wait for a key press.

**OS**
Windows 10.0.19045 Build 19045

**Terminal/Console**

Windows terminal

lesleyrs:
I just did some testing on various terminals and shells and the "release" event does not appear on command prompt as long as you are using conhost instead of wezterm or windows terminal.

No clue why this happens, needs to be looked in to. Powershell on conhost has the release event as well.

TimonPost:
Interesting finding! So this issue seems to have to things at play:
- The release event kind introduced in 0.26 was unexpected and caused some confusion.
- There is an enter key code being send 

The second issue was introduced by #745 where I removed:

```
  if !key_event.key_down {
        return None;
    }
```

I wasn't sure what the purpose of this was but likely this was the reason it exited in the code. It early returns if the key was down whereafter it does a check on virtual key codes.

lesleyrs:
Does kitty protocol actually make release events work on linux? I thought it was only used for extra modifier keys.

[...]

PR #775, opened 2023-04-05
Fix examples on windows

lesleyrs:
Hey so the examples/demos have been broken since I found this crate, this stops "double presses" from happening by only looking for presses instead of all events. It seems to make demos work in conhost, windows terminal, and wezterm.

Edit: the "release" event does not seem to appear on cmd, so the problem might not be crossterm and I removed this hack.

Fixes https://github.com/crossterm-rs/crossterm/issues/759, closes https://github.com/crossterm-rs/crossterm/issues/772, closes https://github.com/crossterm-rs/crossterm/issues/769

Related issue: https://github.com/crossterm-rs/crossterm/issues/752

Also I had a problem where rust-analyzer didn't work in the `interactive-demo` sub-crate, it just shows a message "file not included in module tree". I had to move the folder out of crossterm and change the path to it in order to make it work. I already tried a bunch of stuff like changing where I open the folder in helix etc. Anyone else have this?

### DIFF
--- a/examples/interactive-demo/src/main.rs
+++ b/examples/interactive-demo/src/main.rs
@@ -2,6 +2,7 @@
 
 use std::io::{self, Write};
 
+use crossterm::event::KeyEventKind;
 pub use crossterm::{
     cursor,
     event::{self, Event, KeyCode, KeyEvent},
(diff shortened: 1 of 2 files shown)


lesleyrs:
Some more notes: Wezterm eats key presses that are assigned to terminal shortcuts, but besides that it works perfectly along with conhost for KeyEventKind (press, release). While windows terminal shortcuts only apply while on the command line, yet it does not work correctly and keys that produce letters immediately send `release`.

jasiukiewicztymon:
Great job, waiting for approvement! 🎉

TimonPost (approved):

Issue #770, opened 2023-03-28
Support bitflags version 2


dzfrias:
**Is your feature request related to a problem? Please describe.**
Bitflags v2 provides a ton of conveniences for types implementing Bitflags. Currently, both the [`KeyModifiers`](https://docs.rs/crossterm/latest/crossterm/event/struct.KeyModifiers.html) and [`KeyboardEnhancementFlags`](https://docs.rs/crossterm/latest/crossterm/event/struct.KeyboardEnhancementFlags.html) types use the `bitflags!` macro. It would be nice to use version 2 so their APIs get access to the new features.

This includes [a few breaking changes](https://github.com/bitflags/bitflags/releases/tag/2.0.0), so there a couple ways to handle it.

Let me know if you want me to submit a PR for any of the options!

**Describe the solution you'd like**
Use Bitflags v2 for the [`KeyModifiers`](https://docs.rs/crossterm/latest/crossterm/event/struct.KeyModifiers.html) and [`KeyboardEnhancementFlags`](https://docs.rs/crossterm/latest/crossterm/event/struct.KeyboardEnhancementFlags.html) types. This would require an increment of the minor version of this crate.

**Describe alternatives you've considered if any**
Either to not go through with this change or gate it with a feature flag.

PR #777, opened 2023-04-07
Upgrade to bitflags 2

TimonPost:
Fixes: #770


kalkin:
Would be nice if a crossterm version would be released, containing this commit. Currently crossterm is the only project in my dependency tree using bitflags@v1 instead of v2

PR #788, opened 2023-06-03
Add support for scrolling left and right.

Gronis:
crossterm does not currently support left and right scrolling. This PR fixes that.

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -575,6 +575,10 @@ pub enum MouseEventKind {
     ScrollDown,
     /// Scrolled mouse wheel upwards (away from the user).
     ScrollUp,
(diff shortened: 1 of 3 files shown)


TimonPost (commented):
~Would be nice if we can also add support for windows, do you have a windows machiene, otherwise i can push to this pr https://learn.microsoft.com/en-us/windows/win32/inputdev/wm-mousehwheel~

I added support for windows. Altough, for some reason, I dont get any events from winapi if I use my third mouse button to scroll horizontally. Not sure why it is not working.

TimonPost (approved):
Should be fine for now, thanks!

PR #801, opened 2023-07-28
Feature: Parse hex colour codes in `serde::de::Deserialize`

Piturnah:
With this PR Crossterm will be able to deserialize `Color` from strings like `"#d69bea"`.


TimonPost (approved):
cool!

PR #802, opened 2023-07-28
Add style::force_color_output() API to override NO_COLOR.

kevin-vigor:
As per comments by @Piturnah in https://github.com/crossterm-rs/crossterm/pull/782#issuecomment-1654742215 we should provide an API to override the NO_COLOR setting.


Piturnah:
Thank you for implementing this fix 😁

PR #806, opened 2023-07-29
Add command for the terminal bell sound

### LINKED ISSUE #730: Allow for ringing the terminal bell
There is no event for ringing the terminal bell, at least not from what I can find.

I would like to see a queue-able and executable event added that rings the terminal bell.

I could ring the terminal bell by printing the relevant escape code, but isn't that what crossterm tries to avoid?

**Additional context**
https://en.wikipedia.org/wiki/Bell_character

Piturnah:
Adds a command that allows for ringing the terminal bell by printing the bell character.

Technically it's not an ansi code, so it felt slightly strange to put it under `Command::write_ansi`, but I don't think it makes sense to implement this separately from the Command API.

I think it's somewhat justified if you reinterpret `write_ansi` as "write to the terminal potentially including ansi codes", but then `is_ansi_code_supported` doesn't make a lot of sense.

But, this is one edge case so I think it's fairly reasonable. What do you think? I think this functionality - even though it's really as simple as printing a char - has a place in the Crossterm library

This should close #730


TimonPost:
Well I do think its not appropriate. We use command strictly for writing ansi sequences to the writer. For things that do not fit into that structure, we expose custom functions e.g terminal_size, cursor_pos, etc....

Piturnah:
Is this API more appropriate? Or would you rather a function like `terminal::bell()`?

TimonPost:
A function is better I think, that works similar to the other api's we have that do not fit the box

jvantuyl:
Any idea what's needed to get this done?  I'm working on an app where it would be very handy.

Piturnah:
> Any idea what's needed to get this done? I'm working on an app where it would be very handy.

For now, you can inline it very easily as all it is does is print the char `0x07`.

PR #810, opened 2023-08-07
Add `CSI 14 t` sequence support

sxyazi:
I noticed that https://github.com/crossterm-rs/crossterm/pull/790 is merged, but some terminals (e.g. VSCode, [Hyper canary](https://github.com/vercel/hyper/issues/7375#issuecomment-1668365392) which are [`node-pty`](https://github.com/microsoft/node-pty) based) do not implement the `ioctl` system call to get the terminal's width and height -- always zeros are returned, but they can get the correct pixel size by `CSI 14 t`.

I'm working on a [terminal file manager](https://github.com/sxyazi/yazi) where getting the correct width and height is critical for resizing images to be previewed, and I'll continue to add tests to the PR if it makes sense!

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -947,6 +947,9 @@ pub(crate) enum InternalEvent {
     /// A cursor position (`col`, `row`).
     #[cfg(unix)]
     CursorPosition(u16, u16),
+    /// Terminal window size in pixels (`width`, `height`).
+    #[cfg(unix)]
+    WindowSize(u16, u16),
     /// The progressive keyboard enhancement flags enabled by the terminal.
     #[cfg(unix)]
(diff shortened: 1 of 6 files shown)


TimonPost:
Awesome!

sxyazi:
Done!

sxyazi:
I changed the timeout to be configurable so that it's more flexible, e.g. a session through ssh may need to wait for a bit large time.

Goooyi:
I came across this when try to use yazi in vscode+tmux and see this pr being mentioned [here](https://yazi-rs.github.io/docs/image-preview/#size) , will this still be merged?

XhstormR:
Why wasn't it merged？

Issue #768, opened 2023-03-27
Keycode::Right not recognized on Windows Git Bash shell


lcmgh:
The following Left, Up, Right, Down events are not being recognized on Windows 10 when user launches the app within Git Bash. It works when using Powershell.

```rust
if crossterm::event::poll(timeout)? {
    if let Event::Key(key) = event::read()? {
        // info!("Key pressed {:?}", &key.code);
        match key.code {
            KeyCode::Char(c) => app.on_key(c),
            KeyCode::Left => app.on_left(),
            KeyCode::Up => app.on_up(),
            KeyCode::Right => app.on_right(),
            KeyCode::Down => app.on_down(),
            _ => {}
        }
    }
}
```

lcmgh:
Update: According to a report of our user, it looks like no key codes are being recognized.

lcmgh:
Does anyone know what is the root cause in here or in which direction one could investigate?

PR #815, opened 2023-08-27
windows: improve/fix raw mode

Atrox:
Raw mode didn't quite work for me as I would have expected on windows. For example: function and arrow keys didn't work at all.

I took inspiration from the raw console mode set in the [containerd/console](https://github.com/containerd/console/blob/8f6c4e4faef5a326d2cd907097d04c0239ee5e2f/console_windows.go#L181-L195) library. After setting those verbatim, everything worked as expected on my side.

Please take those changes with a grain of salt, I'm not a windows guy.


sigmaSd:
This is the first time I see an issue about this, and I used crossterm on windwos for years now with no issues

I think its better if you open an issue first with what you encountered (+reproduction + os version) before attempting changes with no clear benefit and that in-itself isn't clear

Atrox:
To reproduce my issue on windows: Enable raw mode with `crossterm::terminal::enable_raw_mode()`, press `arrow up` on the keyboard. `Arrow up` is not getting sent to the application and instead processed by the terminal/windows.

lcmgh:
Might this also fix https://github.com/crossterm-rs/crossterm/issues/768 ?

szabotudor:
Sorry to bump, but is there a reason why this isn't merged yet?
Testing this merged into current master seems to fix issues I'm facing in cmd, ps and git bash.

PR #817, opened 2023-08-30
Add functions to allow constructing `Attributes` in a const context

kierdavis:
Previously, the only ways to construct new `Attributes` values were `Attributes::default()` or `Attributes::from(_)`, neither of can be made `const fn` due to limitations of these standard library traits.

### DIFF
--- a/src/style/attributes.rs
+++ b/src/style/attributes.rs
@@ -62,6 +62,26 @@ impl BitXor for Attributes {
 }
 
 impl Attributes {
+    /// Returns the empty bitset.
+    #[inline(always)]
+    pub const fn none() -> Self {
+        Self(0)
+    }
+
+    /// Returns a copy of the bitset with the given attribute set.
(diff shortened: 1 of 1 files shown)


kierdavis:
I tried to test with the MSRV in Cargo.toml, 1.58.0, but got:
```
[nix-shell:~/checkouts/crossterm]$ rm -rf Cargo.lock target
[nix-shell:~/checkouts/crossterm]$ cargo version
cargo 1.58.0 (7f08ace4f 2021-11-24)
[nix-shell:~/checkouts/crossterm]$ cargo test
error: failed to parse manifest at `/home/kier/checkouts/crossterm/Cargo.toml`

Caused by:
  namespaced features with the `dep:` prefix are only allowed on the nightly channel and requires the `-Z namespaced-features` flag on the command-line
```
which I don't think is related to my changes. All tests pass on Rust 1.60.0.

TimonPost (approved):

PR #821, opened 2023-09-18
Remove unsafe and unnecessary `size` argument from `FileDesc::read()`

martinvonz:
The `size` argument to `FileDesc::read()` is not checked against the length of the buffer, so `libc::read()` could end up writing past the buffer if we passed a size that's too large. However, we always pass exactly the size of the buffer, so that doesn't happen. Let's just remove the argument since it's not currently needed, thereby removing the risk of bugs if the function is used incorrectly by future callers.

This came up in review of `unsafe` Rust code at my company.

### DIFF
--- a/src/event/source/unix/mio.rs
+++ b/src/event/source/unix/mio.rs
@@ -93,7 +93,7 @@ impl EventSource for UnixInternalEventSource {
                 match token {
                     TTY_TOKEN => {
                         loop {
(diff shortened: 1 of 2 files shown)


Shnatsel:
Are there any blockers to merging this? Is there any way I can help expedite this PR?

TimonPost (approved):

PR #841, opened 2023-11-04
fix docs for window_size

benjajaja:
1. Fix external link formatting.
2. Be specific about all fields possibly being zero, and the pixel fields being documented for **linux** as "unused".
3. (I am not sure about TIOCGWINSZ bing posix).
4. Specify that `window_size()` always returns an error on windows.
5. Put the detailed docs in `WindowSize`.

@markus-bauer this addresses some points of your comment in #790 after it was merged. What do you suggest regarding windows not being supported? I have tried everything in that regard:
1. Getting the console font size, but it is in "logical units", so it would need the current DPI or something like that to get to the font pixel size, to ultimately get the window size by multiplying with `(columns,rows)`.
2. All of the [windows console API](https://learn.microsoft.com/en-us/windows/console/console-reference) only deals in `(columns,rows)`.
3. Using `GetClientRect`, which I almost thought would work, but even that gave me bogus values. Also seems a bit far off the track for a terminal library.

Perhaps this was a mistake, and `window_size()` should be removed from crossterm, after all it's not cross-platform now. Users who need the window size, on unix, can then just use some tty-ioctl lib, e.g. [rustix](https://docs.rs/rustix/latest/rustix/termios/fn.tcgetwinsize.html).

### DIFF
--- a/src/terminal.rs
+++ b/src/terminal.rs
@@ -130,13 +130,19 @@ pub fn disable_raw_mode() -> io::Result<()> {
     sys::disable_raw_mode()
 }
 
-/// Returns the terminal size `(columns, rows)`.
+/// Returns the terminal size in `(columns, rows)`.
 ///
 /// The top left cell is represented `(1, 1)`.
 pub fn size() -> io::Result<(u16, u16)> {
     sys::size()
 }
 
+/// Terminal size in both `(columns, rows)` and pixel `(width, height)`.
+///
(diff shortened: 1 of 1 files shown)


markus-bauer:
Since @TimonPost  merged it, I assume he's okay with adding it. I already said in the old PR, that there is probably no common solution and if you want it for unix, it's trivial to use rustix.
But it doesn't matter as long as the implementation and documentations is correct.

TimonPost (approved):

PR #851, opened 2023-12-22
alacritty supports the kitty protocol

LevitatingBusinessMan:
As of https://github.com/alacritty/alacritty/commit/cb03806e2ab85674c45e87e1bb24dfe2fd1a918c alacritty now suppots kitty's enhanced protocol. The rustdoc for the REPORT_ALTERNATE_KEYS was commented with // instead of /// so I also fixed that.

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -444,7 +444,7 @@ impl Command for DisableBracketedPaste {
 /// * [kitty terminal](https://sw.kovidgoyal.net/kitty/)
 /// * [foot terminal](https://codeberg.org/dnkl/foot/issues/319)
 /// * [WezTerm terminal](https://wezfurlong.org/wezterm/config/lua/config/enable_kitty_keyboard.html)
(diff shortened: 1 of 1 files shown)


src/event.rs
```diff
@@ -444,6 +444,7 @@ impl Command for DisableBracketedPaste {
 /// * [kitty terminal](https://sw.kovidgoyal.net/kitty/)
 /// * [foot terminal](https://codeberg.org/dnkl/foot/issues/319)
 /// * [WezTerm terminal](https://wezfurlong.org/wezterm/config/lua/config/enable_kitty_keyboard.html)
+/// * [alacritty terminal](https://github.com/alacritty/alacritty/commit/cb03806e2ab85674c45e87e1bb24dfe2fd1a918c)
```
joshka, line 447: What about linking to the PR instead?

```suggestion
/// * [alacritty terminal](https://github.com/alacritty/alacritty/pull/7125)
```
Or the issue
```suggestion
/// * [alacritty terminal](https://github.com/alacritty/alacritty/issues/6378)
```

joshka (approved):
This is a documentation only change and is reasonable to merge as is, or with either of the two suggested changes.

Issue #787, opened 2023-05-10
I noticed that crossterm::event::read handling of surrogate pairs was broken on Windows,


tasogare3710:
I would like to send pull request for this issue but the following issue is blocking it.

I noticed that crossterm::event::read handling of surrogate pairs was broken on Windows, so I fixed it and now this solution lets me know about illegal surrogates. We need to use PPYP pattern since such illegal input can poop(something terrible will happen with illegal input), but i can't decide when to pre-pooping pants. It is not friendly to crash as soon as it receives user input.

Do you have any good ideas?

In my opinion, crossterm::event::read may need a way to notify caller when it receives invalid input.

tasogare3710:
Sorry, it takes time to upload the repository.

tasogare3710:
Sorry, I'm late.

# TL;DR - main issue

Another issue is arises by the fact that the PPYP pattern have to used but cannot fail fast.

- - -

You will know that you received illegal input in the following [places in this file where](https://github.com/tasogare3710/crossterm/blob/fix_handling_of_surrogate_pairs/src/event/source/windows.rs) you call `missing_high_surrogate` and `two_high_surrogates`.

- https://github.com/tasogare3710/crossterm/blob/272dbd76cbd8f9f702d44673c34103ef8e037079/src/event/source/windows.rs#L78
- https://github.com/tasogare3710/crossterm/blob/fix_handling_of_surrogate_pairs/src/event/source/windows.rs#L97
- https://github.com/tasogare3710/crossterm/blob/fix_handling_of_surrogate_pairs/src/event/source/windows.rs#L116
- https://github.com/tasogare3710/crossterm/blob/fix_handling_of_surrogate_pairs/src/event/source/windows.rs#L139
- https://github.com/tasogare3710/crossterm/blob/fix_handling_of_surrogate_pairs/src/event/source/windows.rs#L160
- https://github.com/tasogare3710/crossterm/blob/fix_handling_of_surrogate_pairs/src/event/source/windows.rs#L181


It's hard to restore internal state gracefully when receiving illegal input, so it's ideal to fail fast, but calling panic as soon as the library receives input gives your application lose a chance to clean up. This may make the library less usable. But have to prevent worse things from happening.

In this case, there are two solutions I came up with.

1. function calls panic as soon as it receives illegal input
2. a function returns an error value representing that it received illegal input and prompts crate users to take appropriate action.

1&#41; has the issues already described and 2) has multiple issues with the current crossterm:

[...]

PR #857, opened 2024-01-10
Fixed Windows Terminal UTF-16 surrogate pair issue

### LINKED ISSUE #848: Windows Terminalでサロゲートペアが無視される
**Describe the bug**

On Windows Terminal, characters represented by UTF-16 surrogate pairs are not decoded properly and are ignored.

**To Reproduce**

1. Run the code https://gist.github.com/kazatsuyu/d112afeebac0490efac267bbf2355f84 on Windows terminal.
2. Copy and paste `"abc🐈def"`.
3. Confirm that the 🐈 is not displayed.

**Expected behavior**

The 🐈 is displayed.

**OS**

Windows 11 22621.2428

**Terminal/Console**

Windows Terminal

kazatsuyu:
Resolve #848

Windows Terminal seems to send the characters represented by surrogate pairs in the following order:

1. key down event of upper surrogate
2. key up event of the upper surrogate
3. key down event for the lower surrogate
4. key-up event for the lower surrogate

On the other hand, crossterm's surrogate pair handling does not distinguish between key-down/key-up events. This may be reasonable, as there are cases where another terminal sends only key-up events, but this results in an incorrect sequence of input from the Windows Terminal.

This fix uses only the last input in the case of a sequence of upper surrogates and ignores the input of lower surrogates with no upper surrogates; in the Windows Terminal, inputs 1 and 4 above are ignored and a single surrogate pair is created from inputs 2 and 3.

### DIFF
--- a/src/event/sys/windows/parse.rs
+++ b/src/event/sys/windows/parse.rs
@@ -61,19 +61,16 @@ pub(crate) fn handle_key_event(
 }
 
 fn handle_surrogate(surrogate_buffer: &mut Option<u16>, new_surrogate: u16) -> Option<char> {
-    match *surrogate_buffer {
-        Some(buffered_surrogate) => {
-            *surrogate_buffer = None;
-            std::char::decode_utf16([buffered_surrogate, new_surrogate])
-                .next()
-                .unwrap()
-                .ok()
-        }
-        None => {
-            *surrogate_buffer = Some(new_surrogate);
-            None
-        }
+    if new_surrogate < 0xDC00 {
+        // Discard any buffered surrogate value if another high surrogate comes.
+        *surrogate_buffer = Some(new_surrogate);
+        return None;
     }
+    let buffered_surrogate = surrogate_buffer.take()?;
+    std::char::decode_utf16([buffered_surrogate, new_surrogate])
+        .next()
+        .unwrap()
+        .ok()
 }
 
 impl From<&ControlKeyState> for KeyModifiers {


src/event/sys/windows/parse.rs
```diff
@@ -61,19 +61,16 @@ pub(crate) fn handle_key_event(
...
-            std::char::decode_utf16([buffered_surrogate, new_surrogate])
-                .next()
-                .unwrap()
-                .ok()
-        }
-        None => {
-            *surrogate_buffer = Some(new_surrogate);
-            None
-        }
+    if new_surrogate < 0xDC00 {
```
joshka, line 64: What does this value represent. A constant would help document this.

src/event/sys/windows/parse.rs
```diff
@@ -61,19 +61,16 @@ pub(crate) fn handle_key_event(
 }
 
 fn handle_surrogate(surrogate_buffer: &mut Option<u16>, new_surrogate: u16) -> Option<char> {
```
joshka, line 63: This method could do with an explanation that clarifies the intent as well as some unit tests for the same.

src/event/sys/windows/parse.rs
kazatsuyu: Used `LOW_SURROGATE_FIRST` and also named the existing constants.

src/event/sys/windows/parse.rs
kazatsuyu: Added test.

src/event/sys/windows/parse.rs
```diff
@@ -376,3 +373,42 @@ fn parse_mouse_event_record(
         modifiers,
     }))
 }
+
+const HIGH_SURROGATE_FIRST: u16 = 0xD800;
+const LOW_SURROGATE_FIRST: u16 = 0xDC00;
+const LOW_SURROGATE_LAST: u16 = 0xDFFF;
```
joshka, line 379: Constants generally should be at the top of the file

src/event/sys/windows/parse.rs
```diff
@@ -376,3 +373,42 @@ fn parse_mouse_event_record(
...
+                // The newer one will be stored
+                (Some(0xDBFF), None),
+                // A valid char will return
+                (None, Some('\u{10000}')),
+                // The high surrogate will be stored
+                (Some(0xDBFF), None),
+                // The low surrogate will be ignored
+                (None, None),
+            ]
+        );
```
joshka, line 412: Simplify (individual assertions, no testcase function, descriptions on the assertions, arranged in order that makes most sense (empty to full)):
```suggestion
        let mut buf = None;
        let next = handle_surrogate(&mut buf, 0xDC00);
        assert_eq!(
            (buf, next),
            (None, None),
            "Unexpected low surrogate is ignored"
        );

        let mut buf = None;
        let next = handle_surrogate(&mut buf, 0xDBFF);
        assert_eq!(
            (buf, next),
            (Some(0xDBFF), None),
            "High surrogate is buffered"
        );

        let mut buf = Some(0xD800);
        let next = handle_surrogate(&mut buf, 0xDBFF);
        assert_eq!(
            (buf, next),
            (Some(0xDBFF), None),
            "Consecutive high surrogates"
        );

        let mut buf = Some(0xD800);
        let next = handle_surrogate(&mut buf, 0xDC00);
        assert_eq!(
            (buf, next),
            (None, Some('\u{10000}')),
            "Valid Surrogate pair"
        );
```

joshka (approved):
LGTM - I'd suggest making the changes mentioned, mostly for reducing maintenance burden.

tasogare3710:
I am going to delete my fork, so read this. This issue related how to deal with the issue of an illegal bytes into the input.
I may not be able to help because I don't have time to spend on crossterm.

https://github.com/crossterm-rs/crossterm/issues/787

https://github.com/tasogare3710/crossterm/tree/fix_handling_of_surrogate_pairs
https://github.com/tasogare3710/crossterm/tree/illegal_surrogate_pairs_ppyp

joshka:
> I am going to delete my fork, so read this(#787). This issue related how to deal with the issue of an illegal bytes into the input. I may not be able to help because I don't have time to spend on crossterm.
> 
> https://github.com/tasogare3710/crossterm/tree/fix_handling_of_surrogate_pairs https://github.com/tasogare3710/crossterm/tree/illegal_surrogate_pairs_ppyp

I've pushed a backup copy of your branch to https://github.com/joshka/crossterm/tree/illegal_surrogate_pairs_ppyp (this branch contains both commits), and I'll keep it around, so you should be fine to delete your fork whenever you want. Thanks for the heads up.

PR #862, opened 2024-02-17
Implement Display for KeyCode and KeyModifiers

joshka:
Partially addresses #792

In Ratatui apps using Crossterm, something we'd like to be able to generally provide approaches for is help screens that show the key bindings that will be used. The responsibility for displaying each keys properly most likely belongs in Crossterm rather than each application.

I've tried to make some reasonable choices about displaying the keys that match the target platform (e.g. on Mac, Backspace is Delete, Delete is Fwd Del, Enter is Return, Alt is Option, Ctrl is Control, Super is Command, and on Windows Super is the Windows key). These choices are backed by the style guidelines from Apple and Microsoft where possible and intuition where not (e.g. Next Track instead of Track Next).

It would be nice to be able to have used something like `strum`'s `Display` derive macro to make the implementation of this simple, but there currently isn't a way to make the format for the function keys (`F1`) or the media / modifier key variants work with that, so a more manual approach is taken.

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -98,7 +98,7 @@ use crate::event::{
 };
 use crate::{csi, Command};
 use parking_lot::{MappedMutexGuard, Mutex, MutexGuard};
-use std::fmt;
+use std::fmt::{self, Display};
 use std::time::Duration;
 
 use bitflags::bitflags;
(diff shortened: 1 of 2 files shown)


joshka:
Ping @TimonPost for a review on this.

TimonPost (approved):
Sweet!

src/event.rs
```diff
@@ -835,11 +904,74 @@ pub enum ModifierKeyCode {
...
+
+            #[cfg(target_os = "macos")]
+            ModifierKeyCode::RightSuper => write!(f, "Right Command"),
+            #[cfg(target_os = "windows")]
+            ModifierKeyCode::RightSuper => write!(f, "Right Windows"),
+            #[cfg(not(any(target_os = "macos", target_os = "windows")))]
+            ModifierKeyCode::RightSuper => write!(f, "Right Super"),
+        }
+    }
+}
```
markus-bauer, line 968: This doesn't compile for me on linux:
```rust
922 |     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
    |        ---                                       ^^^^^^^^^^^ expected `Result<(), Error>`, found `()`
    |        |
    |        implicitly returns `()` as its body has no tail or `return` expression
    |
    = note:   expected enum `Result<(), std::fmt::Error>`
            found unit type `()`
```

joshka: Looks like line 922 is the problem, taking a look.

joshka: https://github.com/crossterm-rs/crossterm/pull/888

markus-bauer: It's just a cfg issue, because it doesn't contain any return for non-mac.

joshka: yep - there was an accidentally left in cfg, which I didn't notice because the checks didn't run and I was developing on a mac. I suspect it was also broken on windows.

Issue #784, opened 2023-04-13
How to trigger a response given a focus change event 


DaveTJones:
Conditionally executing print statements as a result of FocusGained or FocusLost from within alacritty/gnome terminal doesn't produce any output when switching between panes. Based on the way things are presented within examples, this looks like it should be fairly straightforward.

To reproduce the issue, run the following script in a terminal (I used alacritty on Pop!_OS). Changing window focus doesn't result in printing:  
  
```
use crossterm::event::{read, Event};

fn print_events() {
    loop {
        match read() {
            Ok(Event::FocusGained) => println!("FocusGained"),
            Ok(Event::FocusLost) => println!("FocusLost"),
            _ => (),
        }
    }
}

fn main() {
    print_events();
}
````

How do I go about getting this to work? Or am I misunderstanding what these events represent?

sigmaSd:
you need to enable focuschange https://github.com/crossterm-rs/crossterm/blob/master/examples/event-read.rs#L92

DaveTJones:
Thanks a tonne, that fixed it. Although that example is pretty clear, the one I initially used from the docs was slightly misleading https://docs.rs/crossterm/latest/crossterm/event/index.html

sigmaSd:
yeah that part could use a pr , there is a section about mouse events https://docs.rs/crossterm/latest/crossterm/event/index.html#mouse-events but its the same case with focus event and paste event

DaveTJones:
I'll submit one later if that's cool?

sigmaSd:
sure thanks

DaveTJones:
Also noticed that this is missing from at least one other example. https://github.com/crossterm-rs/crossterm/blob/master/examples/event-poll-read.rs looks like it implements focus gain and loss events but is missing the relevant execute statements.

PR #864, opened 2024-02-29
docs: update docs for focus change event (#784)

illiteratewriter:


illiteratewriter:
I was also thinking of adding the following lines to the example in the docs:
```
execute!(
        stdout,
        EnableBracketedPaste,
        EnableFocusChange,
        EnableMouseCapture,
    )?;
```
so that there is better clarity in the example. Let me know if this sounds like a good addition.

src/event.rs
```diff
@@ -17,11 +17,11 @@
...
 //! **Make sure to enable [raw mode](../terminal/index.html#raw-mode) in order for keyboard events to work properly**
 //!
-//! ## Mouse Events
+//! ## Mouse and Focus Events
 //!
-//! Mouse events are not enabled by default. You have to enable them with the
-//! [`EnableMouseCapture`](struct.EnableMouseCapture.html) command. See [Command API](../index.html#command-api)
-//! for more information.
+//! Mouse and focus events are not enabled by default. You have to enable them with the
+//! [`EnableMouseCapture`](struct.EnableMouseCapture.html)/[`EnableFocusChange`](struct.EnableFocusChange.html) command.
```
joshka, line 23: Small nit: this could do with a space in between the links.

joshka (approved):
This is a simple doc change which is correct and should be merged.

joshka:
> I was also thinking of adding the following lines to the example in the docs:

Yes - this would be a good idea. Don't forget the corresponding disable though.

joshka (approved):

joshka (approved):

TimonPost (approved):
Cool

PR #879, opened 2024-04-27
perf: speed up SetColors by ~15-25%

joshka:
The SetColors command was executing SetForegroundColor and then
SetBackgroundColor, which writes 2 extra characters per cell compared to
writing both colors in one command. This resulted in about 15-25% more
FPS (19->24 fps) on a fullscreen (171x51) app that writes every cell
with a different foreground and background color, compared to separately
using the SetForegroundColor and SetBackgroundColor commands (iTerm2, M2
Macbook Pro).

The app is the colors_rgb example in Ratatui, which writes every cell
with a different foreground and background color in a loop. The
CrosstermBackend was changed to use SetColors instead of
SetForegroundColor and SetBackgroundColor.

### DIFF
--- a/src/style.rs
+++ b/src/style.rs
@@ -279,13 +279,24 @@ pub struct SetColors(pub Colors);
 
 impl Command for SetColors {
     fn write_ansi(&self, f: &mut impl fmt::Write) -> fmt::Result {
-        if let Some(color) = self.0.foreground {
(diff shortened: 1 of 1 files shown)


joshka:
Before:
image

After:
image

TimonPost (approved):
Noice!

TimonPost:
We can fix CI later.

PR #884, opened 2024-05-04
Fix build break in events.rs

### LINKED ISSUE #883: Build break from #862
**Describe the bug**
Seems #862 broke the build by now always returning a result from `<ModifierKeyCode as Display>::fmt`. Also seems the `#[cfg]` attribute on the match itself isn't correct, as the individual match arms have conditions. Still, even without it still doesn't compile without some tweaks.

**To Reproduce**
1. Try to build.

**Expected behavior**
Build fails with:

```
    |
922 |     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
    |        ---                                       ^^^^^^^^^^^ expected `Result<(), Error>`, found `()`
    |        |
    |        implicitly returns `()` as its body has no tail or `return` expression
    |
    = note:   expected enum `Result<(), std::fmt::Error>`
            found unit type `()`
```

**OS**
 - Any

**Terminal/Console**
 - Any

heaths:
Fixes #883

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -920,7 +920,6 @@ impl Display for ModifierKeyCode {
     ///
     /// On other platforms, the super key is referred to as "Super".
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        #[cfg(target_os = "macos")]
         match self {
             ModifierKeyCode::LeftShift => write!(f, "Left Shift"),
             ModifierKeyCode::LeftHyper => write!(f, "Left Hyper"),


joshka:
Apologies, I didn't see this PR and raised https://github.com/crossterm-rs/crossterm/pull/888 which is the same fix. As I've had approvals before, I'm getting CI runs on that PR, which is highlighting that there's a few problems with the recent merges (and hence I'm fixing the build incrementally as each CI step fails).

TimonPost:
merged the other pr, this can be closed now right?

PR #885, opened 2024-05-04
Improve color detection across platforms

### LINKED ISSUE #882: available_color_count is wrong on Windows
**Describe the bug**
Windows pseudo-TTYs support true color, yes `crossterm::style::available_color_count` reports 8. You need to query if you're in a pseudo-terminal which, BTW, means you can do just about any VT100 sequence including keyboard support, though I'm not sure what "progressive keyboard support" means.

A better detection of what colors are supported can be found at <https://github.com/cli/go-gh/blob/dbd982eba2a041d88d398cce01cb708c4c3503f7/pkg/term/env.go#L70-L72>. I worked on that for the GitHub CLI (was refactored into cli/go-gh from cli/cli) along with colleagues on the Windows terminal team that own conhost and conpty - both of which support VT100 sequences since some version of Windows 10 (TH2, maybe? Anything older isn't supported anyway.)

**To Reproduce**
1. Write a simple bin:
   ```rust
   fn main() {
       println!("{}", crossterm::style::available_color_count());
   }
   ```
3. Run `cargo run`

**Expected behavior**
`256` is printed to stdout.

**OS**
 - Windows 11

**Terminal/Console**
 - Windows Terminal; though, conhost.exe supports this - which cmd.exe and powershell.exe/pwsh.exe use by default unless you default to wt.exe, which uses conpty.exe. Effectively, conpty is dead on built-in terminals and shells.

heaths:
Fixes https://github.com/crossterm-rs/crossterm/issues/882 by improving `style::available_color_count()`:
- Checks ANSI support on Windows.
- Uses the COLORTERM environment variable and falls back to the TERM environment variable.
- Supports "xterm-24bit" and "truecolor" values which return `u16::MAX`.

### DIFF
--- a/src/style.rs
+++ b/src/style.rs
@@ -161,9 +161,23 @@ pub fn style<D: Display>(val: D) -> StyledContent<D> {
 ///
 /// This does not always provide a good result.
 pub fn available_color_count() -> u16 {
-    env::var("TERM")
-        .map(|x| if x.contains("256color") { 256 } else { 8 })
-        .unwrap_or(8)
+    #[cfg(windows)]
+    {
+        // Check if we're running in a pseudo TTY, which supports true color.
+        // Fall back to env vars otherwise for other terminals on Windows.
+        if crate::ansi_support::supports_ansi() {
+            return u16::MAX;
+        }
+    }
+
+    const DEFAULT: u16 = 8;
+    env::var("COLORTERM")
+        .or_else(|_| env::var("TERM"))
+        .map_or(DEFAULT, |x| match x {
+            _ if x.contains("24bit") || x.contains("truecolor") => u16::MAX,
+            _ if x.contains("256") => 256,
+            _ => DEFAULT,
+        })
 }
 
 /// Forces colored output on or off globally, overriding NO_COLOR.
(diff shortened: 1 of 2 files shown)


heaths:
Contains fix for #883 to unblock testing. Tested in Windows Terminal with bash (WSL) and pwsh.exe (WT uses conpty) and cmd.exe (which uses conhost).

TimonPost:
if you can rebase that would be cool

src/style.rs
```diff
@@ -161,9 +161,24 @@ pub fn style<D: Display>(val: D) -> StyledContent<D> {
...
+
+    const DEFAULT: u16 = 8;
+    env::var("COLORTERM")
+        .or_else(|_| env::var("TERM"))
+        .map(|x| match x {
+            _ if x.contains("24bit") || x.contains("truecolor") => u16::MAX,
+            _ if x.contains("256") => 256,
+            _ => DEFAULT,
+        })
+        .unwrap_or(DEFAULT)
```
joshka, line 181:
```suggestion
        .map_or(DEFAULT, |x| match x {
            _ if x.contains("24bit") || x.contains("truecolor") => u16::MAX,
            _ if x.contains("256") => 256,
            _ => DEFAULT,
        })
```

joshka (approved):
LGTM

I'd suggest adding a small amount of details to the commit message body instead of just linking to the issue as this makes it easier for people to review what's changed in a release.

Something like:
```
`style::available_color_count()` now:
- checks support_ansi on windows
- uses COLORTERM environment variable with fallback to TERM
- supports 24bit / truecolor (reported as u16::MAX)
```

Would it be worthwhile to add unit tests for this? Perhaps using https://crates.io/crates/temp-env?

joshka:
LGTM

Issue #886, opened 2024-05-05
Build error with use-dev-tty


gwenn:
**Describe the bug**
Build fails with `use-dev-tty` feature

**To Reproduce**

```sh
% cargo check --no-default-features --features "use-dev-tty events"
...
error[E0061]: this method takes 1 argument but 2 arguments were supplied
  --> src/event/source/unix/tty.rs:83:18
   |
83 |         match fd.read(buf, buf.len()) {
   |                  ^^^^    -----------
   |                          | |
   |                          | unexpected argument of type `usize`
   |                          help: remove the extra argument
   |
note: method defined here
  --> src/terminal/sys/file_descriptor.rs:32:12
   |
32 |     pub fn read(&self, buffer: &mut [u8]) -> io::Result<usize> {
   |            ^^^^        -----------------
```

It seems to be related to #821

**Expected behavior**
The build should succeed.
And ideally, `use-dev-tty` should imply `events` but without `mio` dependencies.

**OS**
 - MacOs

joshka:
Fixed in https://github.com/crossterm-rs/crossterm/pull/888

gwenn:
Only partially fixed:
> And ideally, use-dev-tty should imply events but without mio dependencies.

joshka:
> Only partially fixed:
> 
> > And ideally, use-dev-tty should imply events but without mio dependencies.

Can you expand on this?

gwenn:
* `use-dev-tty` is useless without `events`
* but `events` implies `mio` while `mio` dependencies are useless with `use-dev-tty`

joshka:
Got it - makes sense.

PR #887, opened 2024-05-05
Test use-dev-tty feature

gwenn:
Build should failed until #886 is fixed.


joshka (approved):
LGTM

joshka:
I cherry-picked this into #888 to ensure that the combination of fixes tested this correctly.
I suspect that if the merge is squashed, then you'll lose the credit for this, but I wanted to make sure that all the build breaking issues were fixed at once, so I hope this is ok with you.

gwenn:
@joshka Thanks.

PR #888, opened 2024-05-05
fix: broken build issues

joshka:
Several issues were encountered after the merging of various PRs.
This was mostly caused by CI not being run on all of the PRs (I suspect
this was due to the GitHub first time repo contributor setting).

- An incorrect #[cfg] attribute introduced in #862 prevented compilation on unix
- Several formatting errors are fixed
- A for loop that (validly) only inspects the first element changed to if ....next() == Some(...)
- FileDesc.read now does not take a length parameter
- Key display example requires the "events" feature flag
- Adds use-dev-tty to testing matrix (credit gwenn <gtreguier@gmail.com>) in #887 
 
Fixes: https://github.com/crossterm-rs/crossterm/issues/883
Fixes: https://github.com/crossterm-rs/crossterm/issues/883
Supersedes: https://github.com/crossterm-rs/crossterm/issues/887
Supersedes: https://github.com/crossterm-rs/crossterm/issues/884

### DIFF
--- a/.github/workflows/crossterm_test.yml
+++ b/.github/workflows/crossterm_test.yml
@@ -66,6 +66,10 @@ jobs:
       if: matrix.os != 'windows-2019'
       run: cargo test --no-default-features -- --nocapture --test-threads 1
       continue-on-error: ${{ matrix.can-fail }}
+    - name: Test no default features with use-dev-tty feature enabled
+      if: matrix.os != 'windows-2019'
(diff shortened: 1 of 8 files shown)


joshka:
Looks like there's a few build failures in the current project that need to be cleared up. The CI makes it difficult to clear them up as a whole rather than one problem at a time due to failing at the first problem rather than failing multiple jobs in parallel.

TimonPost:
Great -- my idea was to follow up too, thanks for the quick PR to fix it.

PR #889, opened 2024-05-05
style: fix cargo formatting

joshka:


joshka:
Included in https://github.com/crossterm-rs/crossterm/pull/888 instead

PR #892, opened 2024-05-16
Use rustix instead of libc (additive only approach)

joshka:
- **use rustix instead of libc**
- **make rustix the default feature**

This is an purely additive alternative approach to #878. Instead of modifying the code paths that were hitting libc, this version instead adds feature gated additions (one path for not(libc), one for libc).

Closes: #847

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+# Unreleased
+
+- Use Rustix by default instead of libc. Libc can be re-enabled if necessary with the libc feature flag.
+- `FileDesc` now requires a lifetime annotation.
+
 # Version 0.27.1
 
 ## Added ⭐
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,7 +27,7 @@ all-features = true
 # Features
 #
 [features]
-default = ["bracketed-paste", "windows", "events", "libc"]
+default = ["bracketed-paste", "windows", "events"]
 windows = [
     "dep:winapi",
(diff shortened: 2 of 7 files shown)


joshka:
There's still probably going to be a few things to clear up based on the various OS tests for this that run in CI.

notgull:
It looks like `libc` is still in use in the event handling part of some backends, as part of signal handling (which `rustix` does not handle).

notgull:
(If you want help, add me to your forked repo and I will make changes)

joshka:
I can't add you right now (on my phone), but I wouldn't mind a hand with working out the more difficult of the two problems - what to do with the stream to fd conversion (sigwinch is the easy one). I'll take a look tomorrow sometime if you don't get it first.

joshka:
Changed in latest:
- use rustix version of sigwinch signal
- add a lifetime to FileDesc and replace FileDesc::Static to
  FileDesc::Borrowed. This made it necessary to either add a lifetime to
  the libc version of FileDesc or replace all the callers with multiple
  paths (libc, rustix). Changing FileDesc was more straightforward.
  There are no usages of FileDesc found in any repo on github, so this
  change should be reasonably safe.

notgull (approved):
Looks good to me!

joshka:
@TimonPost This is probably good to have your 👀 on it now.

TimonPost (approved):
Nice work, looks good to me!

PR #893, opened 2024-05-17
If ioctl TIOCGWINSZ returns 0,0 try to use env vars instead.

### LINKED ISSUE #891: crossterm::terminal::size() returns Ok((0, 0)) in eshell
**Describe the bug**

When running inside eshell, crossterm claims the size of the terminal is zero.

**To Reproduce**

1. Open eshell, the shell inside Emacs.
2. Run a rust program that invokes `crossterm::terminal::size()`
3. Observe that it returns Ok((0, 0))

**Expected behavior**

crossterm should respect the $COLUMNS and $ROWS shell variables, which are set.

**OS**
Arch Linux, kernel v6.8.2

**Terminal/Console**
eshell on Emacs 29.3

swilde:
For some terminals (for example Emacs eshell/eterm) the ioctl with TIOCGWINSZ might falsely return 0 columns and 0 rows.  If this happens we now try to use LINES and COLUMNS environment variables.

Fixes #891

### DIFF
--- a/src/terminal/sys/unix.rs
+++ b/src/terminal/sys/unix.rs
@@ -13,7 +13,7 @@ use std::fs::File;
 
 use std::os::unix::io::{IntoRawFd, RawFd};
 
-use std::{io, mem, process};
+use std::{env, io, mem, process};
 
 // Some(Termios) -> we're in the raw mode and this is the previous mode
 // None -> we're not in the raw mode
@@ -53,6 +53,10 @@ pub(crate) fn window_size() -> io::Result<WindowSize> {
     };
 
     if wrap_with_result(unsafe { ioctl(fd, TIOCGWINSZ.into(), &mut size) }).is_ok() {
+        if size.ws_row == 0 && size.ws_col == 0 {
+            size.ws_row = env::var("LINES").unwrap_or("0".to_string()).parse::<u16>().unwrap();
+            size.ws_col = env::var("COLUMNS").unwrap_or("0".to_string()).parse::<u16>().unwrap();
+        }
         return Ok(size.into());
     }
 


src/terminal/sys/unix.rs
```diff
@@ -53,6 +53,10 @@ pub(crate) fn window_size() -> io::Result<WindowSize> {
     };
 
     if wrap_with_result(unsafe { ioctl(fd, TIOCGWINSZ.into(), &mut size) }).is_ok() {
+        if size.ws_row == 0 && size.ws_col == 0 {
```
joshka, line 56: Add a comment that tells the future reader why this is necessary (like the PR message)

src/terminal/sys/unix.rs
```diff
@@ -53,6 +53,10 @@ pub(crate) fn window_size() -> io::Result<WindowSize> {
...
+            size.ws_row = env::var("LINES").unwrap_or("0".to_string()).parse::<u16>().unwrap();
+            size.ws_col = env::var("COLUMNS").unwrap_or("0".to_string()).parse::<u16>().unwrap();
```
joshka, line 58:
```suggestion
            size.ws_row = env::var("LINES").map_or(Ok(0), |v| v.parse()).unwrap_or(0);
            size.ws_col = env::var("COLUMNS").map_or(Ok(0), |v| v.parse()).unwrap_or(0);
```

joshka (commented):
Any idea if there are perf concerns to checking the environment variables compare with ioctl call. (Context: Ratatui checks window_size on every frame, I'd expect many apps might do something similar).

swilde:
> Any idea if there are perf concerns to checking the environment variables compare with ioctl call. (Context: Ratatui checks window_size on every frame, I'd expect many apps might do something similar).

I don't expect the operation to be expensive.  In addition it only happens when ioctl
fails.

src/terminal/sys/unix.rs
swilde: I added a comment.

src/terminal/sys/unix.rs
swilde: Thanks for the suggestion, I changed the code accordingly.

joshka (approved):

joshka:
> I don't expect the operation to be expensive. In addition it only happens when ioctl
> fails.

Did a quick benchmark. On a mac, this is ~100ns, so fast enough generally to not matter. If this turned out to be a hotspot (actually measured by a profiler), it would be easy to replace with a lazy static, but there's no need to do this upfront for now.

joshka:
@TimonPost this would probably be easy to merge

joshka:
> Don't mean to be a PITA but is there anything I could do to increase chances of getting the PR merged?

I've messaged @TimonPost about this on discord.

For context I'm just an interested party - not a maintainer of crossterm. Perhaps if you have some time, it might be good to get some extra eyes on the other PRs and help out from the perspective of testing / reviewing code. Does that sound like something that would work for you?

TimonPost:
Personally do not have a presence if ppl writing cli applications find a very usefull use to this im fine with getting it in.

TimonPost (commented):
I would like some comments in the `window_size` function to make this behavior explit. Do you think we should consider adding this to windows as well?

PR #904, opened 2024-07-31
Some fixes to my pull request for other underlines

AhoyISki:
A while ago, I sent a pull request to add undercurls and other types of underline to crossterm.
I just realized that I forgot to include methods for doing that to styles.

Now, `Stylize` will include the `double_underlined`, `undercurled`, `underdotted` and `underdashed` methods.

This commit also makes the printing of these underlines slightly faster by changing the check for `"4:{}"` to be just one `if` statement instead of two.

It also implements `Clone` and `Copy` for `SetCursorStyle`, I did that a while ago though... so I don't really remember why.


src/style/types/attribute.rs
```diff
@@ -175,7 +175,7 @@ impl Attribute {
     ///
     /// See <https://en.wikipedia.org/wiki/ANSI_escape_code#SGR_parameters>
     pub fn sgr(self) -> String {
-        if (self as usize) > 4 && (self as usize) < 9 {
+        if (self as usize) > 23 {
```
TimonPost, line 178: can you elaborate on this check

joshka: 23 makes it difficult to work out how this code works and makes the code fragile to changes. Would this work better? It still assumes that the underline types are sequential, but makes it easier to understand what bounds are in play here.
```suggestion
        if (self as usize >= Self::DoubleUnderlined as usize) && (self as usize <= Self::Underdashed as usize) {
```

src/style/types/attribute.rs
```diff
@@ -175,7 +175,7 @@ impl Attribute {
...
             return "4:".to_string() + SGR[self as usize].to_string().as_str();
```
joshka, line 179:
```suggestion
            return format!("{}:{}", Self::Underlined as usize, SGR[self as usize]);
```

src/style/types/attribute.rs
joshka: That's safe unless you add some other variant at the end, or you move the options back somewhere within the enum like they were originally. Putting upper and lower bounds on the check makes it clear that the rationale is that these options are contiguous in the enums variants. I'd guess that having this check is unlikely to be on the hot path (and is probably even compiled out in a release build).

Issue #903, opened 2024-07-31
Crossterm 0.28.0 on crates.io differs from this repo and cannot build on Linux/macOS


gyscos:
**Describe the bug**

Trying to build crossterm 0.28.0 from crates.io leads to failures related to `signal-hook-mio` missing the `support-v0_8` feature (and instead having `support-v1_0`).

Looking at the `Cargo.toml` from the crate downloaded from crates.io, it indeed differs from the one in this repository.

```bash
diff -w <(
    # The version from the git repo
    curl -sL https://raw.githubusercontent.com/crossterm-rs/crossterm/0.28/Cargo.toml
) <(
    # The version from crates.io
    curl -sL 'https://crates.io/api/v1/crates/crossterm/0.28.0/download' |
    tar Oxzf - crossterm-0.28.0/Cargo.toml.orig
)
```

```diff
84,86c84,86
< mio = { version = "0.8", features = ["os-poll"], optional = true }
< signal-hook-mio = { version = "0.2.3", features = [
<     "support-v0_8",
---
> mio = { version = "1.0", features = ["os-poll"], optional = true }
> signal-hook-mio = { version = "0.2.4", features = [
>     "support-v1_0",
93c93
< tokio = { version = "1.25", features = ["full"] }
---
> tokio = { version = "1.39", features = ["full"] }
```

The rest of the archive matches the current repository, only `Cargo.toml` and `Cargo.lock` differ.
So essentially it switched mio from 0.8 to 1.0 in `Cargo.toml`, but did not actually change the code to support mio 1.0:
https://github.com/crossterm-rs/crossterm/blob/5d50d8da62c5e034ef8b2787a771a2c0f9b3b2f9/src/event/source/unix/mio.rs#L4

**To Reproduce**

```bash
cargo new crossterm-test && cd crossterm-test
cargo add crossterm
cargo check
```
Output:
```
   Compiling libc v0.2.155
   Compiling autocfg v1.3.0
   Compiling parking_lot_core v0.9.10
   Compiling signal-hook v0.3.17
    Checking cfg-if v1.0.0
    Checking log v0.4.22
    Checking scopeguard v1.2.0
   Compiling rustix v0.38.34
    Checking smallvec v1.13.2
    Checking linux-raw-sys v0.4.14
    Checking bitflags v2.6.0
   Compiling lock_api v0.4.12
    Checking signal-hook-registry v1.4.2
    Checking mio v1.0.1
    Checking parking_lot v0.12.3
[...]

PR #907, opened 2024-08-01
Fix mio and signalhook

TimonPost:
Fixes: #903

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Unreleased
 
+# Version 0.28.1
+
+## Fixed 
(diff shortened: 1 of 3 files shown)


EdJoPaTo:
Heads up: this broke support of Rust 1.63 resulting in 0.28 no longer be possible to use for Debian. If that's ok, it should be updated in the `Cargo.toml`.

joshka:
I commented similarly recently on a Ratatui issue that Rust 1.63 is 2 years old in 6 days time (August 11). Debian needs to do better about updating software more regularly than that. Given Rust has a 6 week release schedule, 1.63 is 17 releases behind current. Software projects should not be held hostage to any standard that doesn't keep current like Debian.

PR #909, opened 2024-08-07
feat(cursor): Derive multiple traits for "SetCursorStyle"

linrongbin16:
This PR derives multiple useful traits for `SetCursorStyle` enum:

I checked other enums in `crossterm`, and seems they all have these traits. So I think it's ok to also add to `SetCursorStyle`.

It also adds 1 unit test to make sure they're working.

### DIFF
--- a/src/cursor.rs
+++ b/src/cursor.rs
@@ -370,7 +370,7 @@ impl Command for DisableBlinking {
 /// # Note
 ///
 /// - Commands must be executed/queued for execution otherwise they do nothing.
-#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum SetCursorStyle {
     /// Default cursor shape configured by the user.
     DefaultUserShape,
--- a/src/cursor.rs
+++ b/src/cursor.rs
@@ -433,7 +433,7 @@ mod tests {
 
     use super::{
         sys::position, MoveDown, MoveLeft, MoveRight, MoveTo, MoveUp, RestorePosition,
-        SavePosition, SetCursorStyle,
+        SavePosition,
     };
 
     // Test is disabled, because it's failing on Travis
--- a/src/cursor.rs
+++ b/src/cursor.rs
@@ -432,8 +432,7 @@ mod tests {
     use crate::execute;
 
     use super::{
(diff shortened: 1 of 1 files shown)


joshka:
I'm curious what use case there would be for ordering these? I can understand hashing them to store in some lookup table perhaps, but I can't think up a way that ordering them would be useful.

linrongbin16:
> I'm curious what use case there would be for ordering these? I can understand hashing them to store in some lookup table perhaps, but I can't think up a way that ordering them would be useful.

I simply read other enums in this repo, and found many of them (actually all of the others) derives the `PartialOrd` and `Ord`. So I also add them.

But you're right, after a second thought, the `PartialOrd` and `Ord` don't have any meaning here.

src/cursor.rs
```diff
@@ -501,4 +503,20 @@ mod tests {
...
+        assert_eq!(format!("{:?}", SetCursorStyle::SteadyBar), "SteadyBar");
+        let s1 = SetCursorStyle::DefaultUserShape;
+        let s2 = s1;
+        let s3 = s1.clone();
+        assert!(s2 == s1);
+        assert!(s3 == s1);
+        let mut hasher = DefaultHasher::new();
+        assert!(s1.hash(&mut hasher) == s2.hash(&mut hasher));
+        assert!(s1.hash(&mut hasher) == s3.hash(&mut hasher));
+    }
```
joshka, line 521: I think I wouldn't worry about testing the derived traits unless there's a need to maintain some sort of backwards compatibility or external guarantee that these are a particular way. Thoughts on removing this?

```suggestion
```

joshka (approved):
LGTM other than the comment about removing the tests.

src/cursor.rs
joshka: (I'd merge this with these tests btw if you feel strongly about it, it's just not code that I'd personally want to have lying around a project unless they're necessary)

linrongbin16:
> LGTM other than the comment about removing the tests.

Update: accepted the suggestion.

joshka (approved):

Schwa10:
Cómo ver algo prohibido sexo

linrongbin16:
> Cómo ver algo prohibido sexo

I think this is a spam message.

joshka:
@linrongbin16 thanks - reported user, deleted comment.

PR #914, opened 2024-08-15
make Attributes inner value pub

### LINKED ISSUE #774: expose Attributes inner value
For my application I'm storing CUI related information encoded using bincode, namely text Attributes are being stored. Currently the only way I've been able to store the relevant information is by checking each `Attribute` via the `Attributes.has(..)` function, then setting them in a different struct which I then encode with bincode. 

The current alternative that which I've implemented described in the first paragraph.. it's a bit cumbersome. 

It would be really nice to have the the inner value of attributes exposed. I see two basic options: 
1) make the value pub directly (`pub struct Attributes(u32);` -> `pub struct Attributes(pub u32);`) 
2) have `Attributes` implement a function to expose the value, maybe something such as  `pub const fn inner_value(&self) -> u32`. Additionally I could see a function like this being made only available under a `internals` feature flag if desired... however I'm noticing that crossterm does not yet have such a flag. 

Let me know your thoughts! THANK YOU for all the incredible work that's gone into this library :)

bogzbonny:
closes #774 
I'm open to the other method I mentioned in #774, just did the easiest approach to start. 
Let me know what you think

### DIFF
--- a/src/style/attributes.rs
+++ b/src/style/attributes.rs
@@ -4,7 +4,7 @@ use crate::style::Attribute;
 
 /// a bitset for all possible attributes
 #[derive(Debug, Default, Clone, Copy, PartialEq, Eq)]
-pub struct Attributes(u32);
+pub struct Attributes(pub u32);
 
 impl From<Attribute> for Attributes {
     fn from(attribute: Attribute) -> Self {


bogzbonny:
as per discussions in #774

PR #916, opened 2024-08-20
Add commands for scrolling region manipulation.

nfachan:
One command sets the scrolling region, the other resets it to be the whole screen.

I don't know what to do about the pre-windows-10 winapi.

Is this something that can be implemented using the winapi? If not, what happens then? I'd love some guidance.

### DIFF
--- a/src/terminal.rs
+++ b/src/terminal.rs
@@ -330,6 +330,46 @@ impl Command for ScrollDown {
     }
 }
 
+/// A command that sets the scrolling region.
+///
+/// # Notes
+///
+/// Commands must be executed/queued for execution otherwise they do nothing.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct SetScrollingRegion(pub u16, pub u16);
+
+impl Command for SetScrollingRegion {
+    fn write_ansi(&self, f: &mut impl fmt::Write) -> fmt::Result {
+        write!(f, csi!("{};{}r"), self.0, self.1)?;
+        Ok(())
+    }
+
+    #[cfg(windows)]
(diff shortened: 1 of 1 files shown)


joshka (changes requested):
Regarding the winapi side of this, there doesn't seem like there's an exact equivalent of this as a command in winapi. The desired behavior of scrolling a region can be done in a single call instead of one for setting the region and one to scroll. (see https://learn.microsoft.com/en-us/windows/console/scrolling-the-screen-buffer).

It might be possible to emulate this behavior, but it would be complex, likely have many edge cases and hence would probably cause a significant maintenance burden to get right. I wouldn't recommend doing so.

The better approach is to return an io error indicating that the feature is unsupported as this allows apps to gracefully do something else.

nfachan:
The calling application is setting the region, scrolling, then resetting the region. So it probably makes sense to just change the commands to be something like `ScrollRegionUp` and `ScrollRegionDown`.

I'll attempt to make that change.

Thanks for the pointer to the Windows API documentation.

joshka:
> The calling application is setting the region, scrolling, then resetting the region. So it probably makes sense to just change the commands to be something like `ScrollRegionUp` and `ScrollRegionDown`.
> 
> I'll attempt to make that change.
> 
> Thanks for the pointer to the Windows API documentation.

I think what you have here is reasonable - mainly as it maps directly to an actual command that would be sent. Abstracting the region stuff might still be reasonable to do (I think it's probably not), but should be implemented as some sort of  composite of the lower level commands you have here.

Putting my Ratatui hat on, having a scroll area work like this this might be really nice to experiment with.

PR #918, opened 2024-08-24
Add region-scrolling commands.

nfachan:
For terminals that support ANSI, these are implemented by temporarily setting the scrolling region, scrolling, then resetting the scrolling region.

For now, these are not implemented for non-ANSI windows terminals.

I don't think I need to implement `is_ansi_code_supported`, since it seems like the default implementation should be good. These ANSI escape sequences are supported by Windows.

I plan on following up with a windows implementation eventually.

### DIFF
--- a/src/terminal.rs
+++ b/src/terminal.rs
@@ -330,6 +330,90 @@ impl Command for ScrollDown {
     }
 }
 
+/// A command that scrolls the terminal screen a given number of rows up in a specific scrolling
+/// region.
+///
+/// # Notes
+///
+/// Commands must be executed/queued for execution otherwise they do nothing.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct ScrollUpInRegion {
+    /// The first row of the scrolling region.
+    pub first_row: u16,
+
+    /// The last row of the scrolling region.
+    pub last_row: u16,
+
+    /// The number of lines to scroll up by.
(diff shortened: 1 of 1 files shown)


joshka:
The rationale for splitting these into multiple commands would be that in an app you might have a scroll region, which you set once at the start of the app, scroll it whenever you need to, and reset it at the end of the app. That is to say, I liked the previous PR where this was broken out. This seems like it's possibly one layer of abstraction above where most of crossterm's commands are targeted. Do you agree, or have I missed something?

nfachan:
> The rationale for splitting these into multiple commands would be that in an app you might have a scroll region, which you set once at the start of the app, scroll it whenever you need to, and reset it at the end of the app. That is to say, I liked the previous PR where this was broken out. This seems like it's possibly one layer of abstraction above where most of crossterm's commands are targeted. Do you agree, or have I missed something?

I guess it depends on how you want to support the old windows API. If the plan is to never support it, then I think hewing close to the ANSI escape sequences is best (i.e. my original PR where there were just commands for setting and resetting the scrolling region). On the other hand, if you do want to support the old windows API, you'd need something like this that's implemented at a bit of a higher level.

I personally don't have a preference. I just want to do whatever needs to be done to get ratatui to support scrolling regions by default!

nfachan:
I create #923 as an alternative.

Issue #912, opened 2024-08-13
enable_raw_mode() error in version 0.28.0 under WSL and Android


tajkhan:
# Solution

rustix has released 0.38.36, which fixes this bug. Crossterm 0.28.1 is compatible with this version of rustix. Run `cargo update` to update your dependencies in your lock file.

---

# Problem

**Describe the bug**
Was following the [hecto tutorial](https://github.com/pflenker/hecto-tutorial/commit/0d30c3d9ea77beda16d19149b8720608378dd029) which uses crossterm and found that the `enable_raw_mode()` function in bash under WSL in the latest version "0.28.0" returns an error. Apparently was working fine in version "0.27.0".

**To Reproduce**
1. Go to bash in Ubuntu under WSL and create a new binary crate with the following code in src/main.rs:

```
use std::io::{self, Read};
use crossterm::terminal::enable_raw_mode;
use crossterm::terminal::disable_raw_mode;

fn main() {
    enable_raw_mode().unwrap();
    for b in io::stdin().bytes() {
        let c = b.unwrap() as char;
        println!("{}", c);
        if c == 'q' {
	    disable_raw_mode().unwrap();
            break;
        }
    }
}
```

2. Do `cargo add crossterm`. By default it will add the latest version "0.28.0".
3. Run `cargo run`
4. The output contains the error:  'called `Result::unwrap()` on an `Err` value: Os { code: 25, kind: Uncategorized, message: "Inappropriate ioctl for device" }'
5. Moving the dependency version to "0.27.0" as used by the tutorial works fine though.

**Expected behavior**
Calling `enable_raw_mode()` should return `Ok(())` and switch the terminal to raw mode.

**OS**
 I am running bash in Ubuntu in WSL under Windows 10 home edition; The terminal is xterm-256color.

joshka:
Please post a full error message including backtrace for this if you can.

Good to hear switching back to libc works. I pushed to keep the libc code in place when the rustix implementation was added, and was starting to second guess myself about being a little over cautious.

sunfishcode:
[...]

PR #926, opened 2024-09-06
fix: Update `rustix` to fix the `enable_raw_mode()` error on WSL/Android

sxyazi:
Fixes https://github.com/crossterm-rs/crossterm/issues/912

Related upstream PR: https://github.com/bytecodealliance/rustix/pull/1147

---

Should we limit `rustix` to just the mirror version, similar to what `libc` does, so that upstream fixes will automatically apply in the future?

https://github.com/crossterm-rs/crossterm/blob/b056370038416584746fd59f7576ecd218b29f8d/Cargo.toml#L77

Also if possible, could we release a patch version of `crossterm` that includes this fix?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -75,7 +75,7 @@ crossterm_winapi = { version = "0.9.1", optional = true }
 # Default to using rustix for UNIX systems, but provide an option to use libc for backwards
 # compatibility.
 libc = { version = "0.2", default-features = false, optional = true }
-rustix = { version = "0.38.34", default-features = false, features = [
+rustix = { version = "0.38.36", default-features = false, features = [
     "std",
     "stdio",
     "termios",


joshka:
> Should we limit `rustix` to just the mirror version, similar to what `libc` does, so that upstream fixes will automatically apply in the future?

Being explicit about versions like this means that when cargo does dependency resolution, it will not resolve a set of versions that does not work. `version = "0.38.36"` means any version of `0.38.y` where y > = 36. I'd suggest not moving to using `version = "0.38"` because 0.38.35 is specifically broken and should not be used.

Apps that don't specifically lock the rustix version can just run `cargo update` to get the updated version of rustix even without this change, so this PR is fine to merge and then release when it makes sense.

joshka (approved):
LGTM

sxyazi:
> it will not resolve a set of versions that does not work. `version = "0.38.36"` means any version of `0.38.y` where y > = 36

TIL about this, it makes total sense. Thanks for the excellent explanation!

joshka:
@TimonPost this is a simple version bump and is pretty safe to merge.

voronind-com:
Gently pinging @TimonPost this PR so Yazi could be used on Android again. Thanks!

joshka:
> Gently pinging @TimonPost this PR so Yazi could be used on Android again. Thanks!

This is not a blocker for building anything downstream. Specifying `version = "0.38.34"` is fully compatible with any consuming crate resolving that version to the semver compatible `version = "0.38.36"`. Yazi can update its Cargo.toml to the later version if it specifically needs to use that version to work, or at a minimum it can run `cargo update` and commit the change.

That said, I'm going to merge this.

TimonPost:
I'll see if i'll be able to do some release soon, thanks for the PR

PR #929, opened 2024-09-22
implement cell size pixels CSI and pixel mouse input example

### LINKED ISSUE #873: Support Pixel Coordinates in Mouse Events
Some terminals offer the ability to report pixel coordinates of mouse events along with the character coordinate. I want to make a UI based on the kitty image protocol (also blocked by #834) where pixel coordinates of events are important.

I would like crossterm to provide access to these modes and add the pixel coordinates to `MouseEvent`.

**Additional context**
I found this website describing the [SGR-Pixels (1016)](https://invisible-island.net/xterm/ctlseqs/ctlseqs.html#:~:text=improvement%20over%201006.-,sgr-pixels%20(1016),-Use%20the%20same) mode.

bogzbonny:
Closes #873 

Based on conversations within https://github.com/crossterm-rs/crossterm/issues/873 I'd originally thought it'd be useful to be able to provide the mouse location in pixels and in cells, however upon further reflection this would require that crossterm hold state of what the most recent cell size was (which would change with each Resize) - hence I think the best way we can address this is to simply allow for users to get the mouse size in pixels and provide tooling for users to get the cell size in pixels themselves and hold this state, so if necessary the user is able to easily calculate the cell coordinates as well whenever they need it.

This PR has 3 main changes
 - Ansi for pixel capture mode (see `EnableMousePixelCapture`) 
 - Ansi for cell size in pixels (see `cell_size()`) [CAN MAYBE REMOVE]
 - An example (`cargo run --features="event-stream" --example event-stream-tokio-pixels`)
 
OPEN QUESTION: 
 - why does the cell size according window calculation differ (by 1 in width and height) from the CSI response cell size?
   - if you run the PRs example an hit `s` you will see the difference in both calculations
   
@bbb651 @joshka

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -123,6 +123,10 @@ name = "event-stream-tokio"
 required-features = ["event-stream", "events"]
 
 [[example]]
+name = "event-stream-tokio-pixels"
+required-features = ["event-stream", "events"]
+
+[[example]]
 name = "event-read-char-line"
 required-features = ["events"]
 
--- a/examples/event-stream-tokio-pixels.rs
+++ b/examples/event-stream-tokio-pixels.rs
@@ -1,0 +1,78 @@
+//! Demonstrates how to read events asynchronously with tokio.
+//!
+//! cargo run --features="event-stream" --example event-stream-tokio-pixels
+
+use std::{io::stdout, time::Duration};
+
+use futures::{future::FutureExt, select, StreamExt};
+use futures_timer::Delay;
+
+use crossterm::{
+    cursor::position,
+    event::{DisableMousePixelCapture, EnableMousePixelCapture, Event, EventStream, KeyCode},
+    execute,
+    terminal::{self, cell_size, disable_raw_mode, enable_raw_mode},
+};
+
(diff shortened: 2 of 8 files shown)


bogzbonny:
it seems that margins may be the culprit as to cell size difference... suggesting that we really should keep the `cell_size()` function

bogzbonny:
Additionally to this PR it would be good to have the ability to Request Ansi Mode (DECRQM) so the user could actually determine if 1016 is supported - I'm not sure but it seems as though this form of request is not supported anywhere in crossterm. 

for context ([ref](https://invisible-island.net/xterm/ctlseqs/ctlseqs.html#:~:text=improvement%20over%201006.-,sgr-pixels%20(1016),-Use%20the%20same)):
```
CSI Ps $ p
          Request ANSI mode (DECRQM).  For VT300 and up, reply DECRPM is
            CSI Ps ; Pm $ y
          where Ps is the mode number as in SM/RM, and Pm is the mode
          value:
            0 - not recognized
            1 - set
            2 - reset
            3 - permanently set
            4 - permanently reset
```

Issue #917, opened 2024-08-22
Wrong KeyEvent reported to yazi


SchuhBaum:
**Describe the bug**
See https://github.com/sxyazi/yazi/issues/1534.
It seems that crossterm reports `ctrl+alt+\` instead of just `\` when pressing `ctrl+alt+ß` on a german keyboard. This creates an issue because yazi destincts between `\` and `ctrl+alt+\`.

**To Reproduce**
See https://github.com/sxyazi/yazi/issues/1534.
Let me know if there is a way to check that indeed this dependency is at fault. The reported steps from the link use yazi.

**Expected behavior**
The keybinding `ctrl+alt+ß` is translated to the key event with `KeyCode::Char('\\')` without modifiers.

**OS**
 Windows

**Terminal/Console**
Powershell

joshka:
See https://github.com/crossterm-rs/crossterm/issues/685

joshka:
I'm going to reopen this one after doing a bit more reading as it's a little more complex than just a missing key.

On a German keyboard layout (I'm probably assuming / generalizing here), there is a third layer of keys on the standard keyboard which are accessed via either the AltGr key (which is often the right Alt button), or the Ctrl+Alt key combination on keyboards that don't do automatic AltGr translation.

On the layout, the `ß` key is to the right of the `0` key on the keyboard. This key has a `\` in its third layer.
So while both the control and alt keys are held down, these are used to instead select the layer not to modify the key.

This likely is something that would only be fixed by implementing a fairly deep fundamental change to interacting with the console in crossterm using [virtual terminal methods](https://learn.microsoft.com/en-us/windows/console/console-virtual-terminal-sequences#modifiers) instead of the more legacy low level methods. I'm not sure this would be reasonable to implement as a change in crossterm, but it's worth discussing. The main problem would be that the new methods don't continue to return the window size events etc, but they would handle stripping AltGr / Alt+Ctrl automatically.

SchuhBaum:
How feasible is it to just throw away `AltGr` / `Ctrl+Alt` when the detected code isn't a control code (see the TODO section below)?
Or to add a KeyCode category or some variable to the key event that says, hey we found a control code.

```
_ => {
            let utf16 = key_event.u_char;
            match utf16 {
                0x00..=0x1f => {
                    log_to_file(&format!("TEMP: Case 1 (control code or no u_char)."));

                    // Some key combinations generate either no u_char value or generate control
[...]

Issue #685, opened 2022-06-30
Tracking: Missing key combinations


TimonPost:
Parsing ANSI codes to input events is a pain. There is no good way to tell which terminals implement which key combinations. Therefore it might occur that some terminals do support certain key presses while others dont. This issue serves as a tracking issue and lists various keys that are reported missing. Either it might be that crossterm has not yet implemented it or that the terminal doesn't support the key combination. Please post your missing keys in this issue.

# How to debug

[This code](https://github.com/crossterm-rs/crossterm/blob/master/src/event/sys/unix/parse.rs#L26) is responsible for parsing a bytes buffer to an event. If you miss a key, it is useful if you can print the buffer with `println!("Buffer: {:?}, buffer)` in this function. Then report that value with the missing keys in this issue. That way we can trace back and see if crossterm should add parsing logic for that buffer. 

# Reported Missing

- `Ctrl + Enter` = modifiers::NONE (iTerm2, Terminal.app, gnome-terminal,VSCode Integrated Terminal) #669, #508
- `Shift + Enter` = modifiers::NONE (iTerm2, Terminal.app, gnome-terminal,VSCode Integrated Terminal) #669
- `Alt + Enter` = modifiers::ALT (iTerm2, Terminal.app, VSCode Integrated Terminal) #669
- `Alt + Enter` = modifiers::NONE (iTerm2, Terminal.app) #669
- `Cmd + Arrow Right` = Some gibberish events (IntelliJ IDEA Terminal)  #597
- `ALT + [` = doesnt work.
- `ALT + ]` Doesnt work.
- `Ctrl+Backspace`  = Missing modifier #575, #504
- `Alt+Backspace` = Missing modifier #575, #504
- Shift + F3 Key Event not read (ubunutu)  #513
-

cloudultima3164:
@TimonPost 
Sorry for the sudden comment. I was just writing up a simple tui using crossterm as a backend, and I ran into the issue. I'm not very experienced with coding, so I could be overlooking something, but it looks like the function for parsing events that you linked above never tries to match Enter with some key modifiers based on input. Again, I'm not really familiar, but could adding a few more match cases could take care of most enter-related issues?

TimonPost:
It depends if the terminal supports escape codes for certain special key combinations. It indeed might be the case that it can be mapped but isn't implemented yet, in that case, it's a simple addition to the existing parsing code. If you have any in mind feel free to add them.

cloudultima3164:
[...]

PR #937, opened 2024-10-27
Manage AltGr on Windows

### LINKED ISSUE #820: AltGr keys produce alt and ctrl modifier on windows
**Describe the bug**
This issue was first reported in https://github.com/helix-editor/helix/issues/7129 including a reprocutions case.
AltGr handling seems a mess on windows but I do think this may be a bug in crossterm.
Disclaimer that i don't have a windows machine or a keyboard that uses altgr at hand myself so I didn't test this myself but the original reproduction case was confirmed by multiple people in helix and we don't do any processing of crossterm key events.

I cameup with the reproduction case below but I can't test it myself:

**To Reproduce**
1. Open windows terminal
2. run the `event-read` exmaple
3. use a german keyboard layout and press `altgr-7` to type `{`
4. Observe modifier `Alt | Ctrl` in the key event (with char `{`)

**Expected behavior**
A key event for `{` without modifier is emitted.

**OS**
Windows

**Terminal/Console**
Windows Terminal (stable and preview) and Git Bash (+winpty)

Geobert:
Fixes #820

Tested with local build of Yazi and Broot

### DIFF
--- a/src/event/sys/windows/parse.rs
+++ b/src/event/sys/windows/parse.rs
@@ -79,18 +79,20 @@ fn handle_surrogate(surrogate_buffer: &mut Option<u16>, new_surrogate: u16) -> O
 impl From<&ControlKeyState> for KeyModifiers {
     fn from(state: &ControlKeyState) -> Self {
         let shift = state.has_state(SHIFT_PRESSED);
-        let alt = state.has_state(LEFT_ALT_PRESSED | RIGHT_ALT_PRESSED);
+        let l_alt = state.has_state(LEFT_ALT_PRESSED);
+        let r_alt = state.has_state(RIGHT_ALT_PRESSED);
         let control = state.has_state(LEFT_CTRL_PRESSED | RIGHT_CTRL_PRESSED);
+        let alt_gr = control && r_alt;
 
         let mut modifier = KeyModifiers::empty();
 
         if shift {
             modifier |= KeyModifiers::SHIFT;
         }
-        if control {
+        if control && !alt_gr {
             modifier |= KeyModifiers::CONTROL;
         }
-        if alt {
+        if (l_alt || r_alt) && !alt_gr {
             modifier |= KeyModifiers::ALT;
         }
 


fnuttens:
Tested with local build of Helix, this solves related issues (https://github.com/helix-editor/helix/issues/5718 and https://github.com/helix-editor/helix/issues/7129) 🎉 

Thank you so much for your work @Geobert!

TimonPost (commented):
Do you have a link to some documentation that this is the correct way to check if altgr is pressed? Like some resource from Microsoft, since I have difficulties looking up the spec for this and no keyboard to try on.

Geobert:
> Do you have a link to some documentation that this is the correct way to check if altgr is pressed? Like some resource from Microsoft, since I have difficulties looking up the spec for this and no keyboard to try on.

Hi, thanks for looking at this.

I don’t have such spec unfortunately, but this is from one of MS team https://github.com/microsoft/Terminal/blob/660d31ac522186e615ec243de2a434c273464828/src/terminal/input/terminalInput.cpp#L400-L419

and other sources quoted in the issue have similar code to manage AltGr :)

Let me know if you need more :)

joshka:
This is also related to https://github.com/crossterm-rs/crossterm/issues/917 and comments in https://github.com/crossterm-rs/crossterm/issues/685#issuecomment-2305442764

Geobert:
It might for combinations using Right alt+Ctrl keys I guess

aschey:
I tested this out on Windows and it does swallow ctrl + right alt key combinations as expected. I did a bit more digging and it seems both wezterm and Windows Terminal have some more complex methods for detecting this.

Windows Terminal [tracks the time between ctrl and alt presses](https://github.com/microsoft/terminal/blob/1040035b5530ede83b4ac032a6c1edc86868a861/src/terminal/input/terminalInput.cpp#L185-L204) (from [this PR](https://github.com/microsoft/terminal/pull/16511)) and decides that it's AltGr if the time between the two events is less than 50 milliseconds.

Wezterm [iterates through unicode key codes](https://github.com/wez/wezterm/blob/4bf0007cefff7f1ad139d89b54f074d5ad7d2184/window/src/os/windows/window.rs#L2218-L2249) with ctrl + alt enabled and checks if any iteration returns a single character.

Geobert:
I think I’m going for the Windows Terminal method, but it seems to be in `crossterm_winapi`, will keep this issue up to date

Geobert:
Printing `InputRecord::KeyEvent::control_key_state` and using `crossterm` interactive demo I get:
for `AltGr`:
```
[D:\Dev\crossterm-winapi\src\console.rs:174:13] &e.control_key_state = ControlKeyState(
    0,
)
[D:\Dev\crossterm-winapi\src\console.rs:174:13] &e.control_key_state = ControlKeyState(
    8,
)
[D:\Dev\crossterm-winapi\src\console.rs:174:13] &e.control_key_state = ControlKeyState(
    265,
)
[D:\Dev\crossterm-winapi\src\console.rs:174:13] &e.control_key_state = ControlKeyState(
    2,
)
[D:\Dev\crossterm-winapi\src\console.rs:174:13] &e.control_key_state = ControlKeyState(
    256,
)
```

when pressing `LCtl` + `LAlt`:
```
[D:\Dev\crossterm-winapi\src\console.rs:174:13] &e.control_key_state = ControlKeyState(
    8,
)
[D:\Dev\crossterm-winapi\src\console.rs:174:13] &e.control_key_state = ControlKeyState(
    10,
)
[D:\Dev\crossterm-winapi\src\console.rs:174:13] &e.control_key_state = ControlKeyState(
    2,
)
[D:\Dev\crossterm-winapi\src\console.rs:174:13] &e.control_key_state = ControlKeyState(
    0,
)
```

Weirdly, pressing `RCtl` doesn’t do anything

So I’m thinking either do the timer thing like Windows Terminal, or check for these weird values 265 and 256 which are basically control and no mod but with an extra `1` bit at the 9th bit

Thoughts?

Geobert:
New approach, let me know if anything more is needed.

Geobert:
Hi @TimonPost, anyway to get this merged?

aschey:
Gave this a try on my US keyboard again and this is unfortunately still swallowing ctrl + right alt combos. Seems like there's probably no way around the hacky workarounds used by Wezterm/Windows Terminal.

PR #938, opened 2024-11-07
Update tui-rs link to point to ratatui

kmicklas:
The old library is archived and unmaintained, with Ratatui as the officially recommended successor.


joshka (approved):

joshka:
Thanks for this :)

PR #943, opened 2024-11-14
Fix warnings and clippy lints

joshka:
Remove winapi feature flag checks
    
>    These checks were impossible to trigger as there is no separate feature
    flag for winapi and crossterm_winapi. The windows feature flag
    automatically enables these dependencies.

Move InternalEventFilter into tests mod
    
>    This is only used in test code, so should be defined there

Fix clippy lints for byte char slices

> [https://rust-lang.github.io/rust-clippy/master/index.html\#byte_char_slices](https://rust-lang.github.io/rust-clippy/master/index.html/#byte_char_slices)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -252,9 +252,3 @@ pub(crate) mod macros;
 
 #[cfg(all(windows, not(feature = "windows")))]
 compile_error!("Compiling on Windows with \"windows\" feature disabled. Feature \"windows\" should only be disabled when project will never be compiled on Windows.");
-
-#[cfg(all(winapi, not(feature = "winapi")))]
-compile_error!("Compiling on Windows with \"winapi\" feature disabled. Feature \"winapi\" should only be disabled when project will never be compiled on Windows.");
-
-#[cfg(all(crossterm_winapi, not(feature = "crossterm_winapi")))]
-compile_error!("Compiling on Windows with \"crossterm_winapi\" feature disabled. Feature \"crossterm_winapi\" should only be disabled when project will never be compiled on Windows.");
--- a/src/event/filter.rs
+++ b/src/event/filter.rs
@@ -61,22 +61,22 @@ impl Filter for EventFilter {
     }
 }
 
-#[derive(Debug, Clone)]
-pub(crate) struct InternalEventFilter;
-
-impl Filter for InternalEventFilter {
-    fn eval(&self, _: &InternalEvent) -> bool {
-        true
-    }
-}
-
 #[cfg(test)]
 #[cfg(unix)]
 mod tests {
     use super::{
         super::Event, CursorPositionFilter, EventFilter, Filter, InternalEvent,
-        InternalEventFilter, KeyboardEnhancementFlagsFilter, PrimaryDeviceAttributesFilter,
+        KeyboardEnhancementFlagsFilter, PrimaryDeviceAttributesFilter,
     };
+
+    #[derive(Debug, Clone)]
+    pub(crate) struct InternalEventFilter;
+
+    impl Filter for InternalEventFilter {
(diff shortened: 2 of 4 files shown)


src/lib.rs
```diff
@@ -252,9 +252,3 @@ pub(crate) mod macros;
 
 #[cfg(all(windows, not(feature = "windows")))]
 compile_error!("Compiling on Windows with \"windows\" feature disabled. Feature \"windows\" should only be disabled when project will never be compiled on Windows.");
-
-#[cfg(all(winapi, not(feature = "winapi")))]
```
TimonPost, line 256: https://github.com/crossterm-rs/crossterm/issues/766

Okay with the other changes but not sure if this is desired since it was implemented to give a more concrete error when developers accidently disable the windows feature flag on windows, and that feaure flag existed for the reasons described in that issue. What do you think?

joshka: Maybe I'm reading this wrong / missing something. There is no `winapi` / `crossterm_winapi` config predicate* (https://doc.rust-lang.org/reference/conditional-compilation.html), only a `windows` one, so these are impossible to trigger. I considered perhaps this might have just been a typo and should have been `#[cfg(all(windows, not(feature = "winapi")))]`, but that would also be impossible to trigger without also triggering the previous compile error as the windows feature enables the winapi and crossterm_winapi deps.

```
warning: unexpected `cfg` condition name: `winapi`
   --> src/lib.rs:256:11
    |
256 | #[cfg(all(winapi, not(feature = "winapi")))]
    |           ^^^^^^
    |
    = help: expected names are: `clippy`, `debug_assertions`, `doc`, `docsrs`, `doctest`, `feature`, `fmt_debug`, `miri`, `overflow_checks`, `panic`, `proc_macro`, `relocation_model`, `rustfmt`, `sanitize`, `sanitizer_cfi_generalize_pointers`, `sanitizer_cfi_normalize_integers`, `target_abi`, `target_arch`, `target_endian`, `target_env`, `target_family`, `target_feature`, `target_has_atomic`, `target_has_atomic_equal_alignment`, `target_has_atomic_load_store`, `target_os`, `target_pointer_width`, `target_thread_local`, `target_vendor`, `test`, `ub_checks`, `unix`, and `windows`
    = help: consider using a Cargo feature instead
    = help: or consider adding in `Cargo.toml` the `check-cfg` lint config for the lint:
             [lints.rust]
             unexpected_cfgs = { level = "warn", check-cfg = ['cfg(winapi)'] }
    = help: or consider adding `println!("cargo::rustc-check-cfg=cfg(winapi)");` to the top of the `build.rs`
    = note: see <https://doc.rust-lang.org/nightly/rustc/check-cfg/cargo-specifics.html> for more information about checking conditional configuration
    = note: `#[warn(unexpected_cfgs)]` on by default

warning: unexpected `cfg` condition value: `winapi`
   --> src/lib.rs:256:23
    |
256 | #[cfg(all(winapi, not(feature = "winapi")))]
    |                       ^^^^^^^^^^^^^^^^^^
    |
    = note: expected values for `feature` are: `bracketed-paste`, `default`, `event-stream`, `events`, `filedescriptor`, `libc`, `serde`, `use-dev-tty`, and `windows`
    = help: consider adding `winapi` as a feature in `Cargo.toml`
    = note: see <https://doc.rust-lang.org/nightly/rustc/check-cfg/cargo-specifics.html> for more information about checking conditional configuration
```
(and same for crossterm_winapi)

So the first check makes sense (windows), the second and third (winapi, crossterm_winapi) don't.

*there are ways to add these (e.g. https://crates.io/crates/cfg_aliases or `rustc --cfg`), but these aren't in use in crossterm AFAICT.

PR #947, opened 2024-11-22
Document feature flags in lib.rs

joshka:
Use the [document-features](https://docs.rs/document-features/latest/document_features/) crate to automatically add documentation about the features to the crate documentation at the crate level.


joshka:
image

PR #951, opened 2024-12-04
feat: Derive Ord for Event and sub-types

nick42d:
Might not be the best practice, but I wanted to use `KeyCode` in a BTreeMap and noticed it implements PartialOrd but not Ord. From what I understand from the Rust docs, the Ord derive is exactly the same as the PartialOrd derive, so I'm proposing that we add it in this PR.

I propagated the change to `Event` and all it's children for consistency.

This change should be non-breaking.

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -546,7 +546,7 @@ impl Command for PopKeyboardEnhancementFlags {
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 #[cfg_attr(feature = "derive-more", derive(IsVariant))]
 #[cfg_attr(not(feature = "bracketed-paste"), derive(Copy))]
-#[derive(Debug, PartialOrd, PartialEq, Eq, Clone, Hash)]
+#[derive(Debug, PartialOrd, Ord, PartialEq, Eq, Clone, Hash)]
 pub enum Event {
     /// The terminal gained focus
     FocusGained,
@@ -773,7 +773,7 @@ impl Event {
 /// combinations for all mouse event types. For example - macOS reports
 /// `Ctrl` + left mouse button click as a right mouse button click.
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
-#[derive(Debug, PartialOrd, PartialEq, Eq, Clone, Copy, Hash)]
+#[derive(Debug, PartialOrd, Ord, PartialEq, Eq, Clone, Copy, Hash)]
 pub struct MouseEvent {
     /// The kind of mouse event that was caused.
     pub kind: MouseEventKind,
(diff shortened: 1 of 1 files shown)


joshka:
I questioned the rationale for ordering a while back in https://github.com/crossterm-rs/crossterm/pull/909#issuecomment-2274879296. Mostly my concern was that there's nothing that seems obvious to me why this would be needed. I'm curious if you can expand on the use case for this a bit more.

I'd be inclined to instead remove even the `PartialOrd` impls from these types.

`KeyCode` seems like the only one that really has an obvious reason to be ordered, and that's mainly so that it would be reasonable to show the keys sorted in UIs. I suspect that the current default top-to-bottom ordering might not be the best there however. Any thoughts on that?

nick42d:
> I questioned the rationale for ordering a while back in https://github.com/crossterm-rs/crossterm/pull/909#issuecomment-2274879296. Mostly my concern was that there's nothing that seems obvious to me why this would be needed. I'm curious if you can expand on the use case for this a bit more.
> 
> I'd be inclined to instead remove even the `PartialOrd` impls from these types.
> 
> `KeyCode` seems like the only one that really has an obvious reason to be ordered, and that's mainly so that it would be reasonable to show the keys sorted in UIs. I suspect that the current default top-to-bottom ordering might not be the best there however. Any thoughts on that?

Got it, I guess to expand on this a little;
I am storing a Map of keybinds (specifically, a struct containing KeyCode + KeyModifiers) and I indeed do want to display them in a UI. For this the actual sort order is not a priority for me, more that the ordering is consistent. For that, the derived ordering Rust provides is ideal, as it is automatically updated and generating my own version would require a lot of boilerplate.

I am wondering if you'd be more amenable to this change if it was limited to just KeyCode and KeyModifiers, perhaps with the addition of some documentation similar to [here](https://docs.rs/enumflags2/latest/enumflags2/struct.BitFlags.html) - as actually using comparison operators on these structs doesn't make much sense.

For the other structs I could personally take it or leave it, but it still doesn't make much sense for them to derive PartialOrd but not Ord. I wonder if anyone downstream is actually using the PartialOrd impls.

TimonPost:
tbh i dont mind implementing whatever derives on it, even if the usecase sounds far-fetched, it is a relatively straightforward derive and does not seem like there are any downsides to it really.

nick42d:
I'm still pro this change, have now rebased back on main.

joshka:
The only (very small) downside I could see for this would be if there's serialized data that relies on the order (and that order changes between crossterm versions due to addition or reordering of variants). It's minor enough, unlikely enough, and it's already there for PartialOrd. So this doesn't make it any worse. So my concerns on this are allayed - merging this.

PR #954, opened 2024-12-11
Fix event-read-char-line example double read error & auto enter

hthuz:
I tried following the example in `event-read-char-line` to read keyboard character event without having to press enter. I met some problems when I tried this example.   

The first problem is double key input in Windows. I tracked issue #772 and it turns out `kind: KeyEventKind::Press` is not added when reading event in this example.  

The second problem is with `read_line()` function. It seems that it captures the pressed `Enter` command when I run `cargo run --example event-read-char-line`.   

The third problem is, I have to press Enter for `read_char()` to get my keyboard input in WSL. 

I played with this example for some time and compared with other examples. I made some modifications and they should fix these problems. For `read_line()`, I don't know exactly why the original `while` doesn't work but replacing it with `loop` does solve the second problem.

### DIFF
--- a/examples/event-read-char-line.rs
+++ b/examples/event-read-char-line.rs
@@ -5,12 +5,13 @@
 
 use std::io;
 
-use crossterm::event::{self, Event, KeyCode, KeyEvent};
+use crossterm::{event::{self, Event, KeyCode, KeyEvent, KeyEventKind}, terminal};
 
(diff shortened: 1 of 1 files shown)


hthuz:
Hi @TimonPost could you review it when you have a chance? I'd appreciate your comments and feedbacks. Thank you!

Issue #952, opened 2024-12-07
More granular query for keyboard enhancement protocol support


Iorvethe:
**Is your feature request related to a problem? Please describe.**
Some terminals could only support a subset of the keyboard enhancement protocol, but `crossterm`’s query for support only returns a boolean. This could be an issue when an application receives a positive answer for support, but the terminal doesn’t implement the entire protocol it expects (because of a *work-in-progress* or else). This issue was recently raised in Zellij (see https://github.com/zellij-org/zellij/issues/3789).

**Describe the solution you'd like**
The proposed solution by Zellij’s maintainer was to either improve the current `supports_keyboard_enhancement` function or to add an additional interface for a more granular query. The second would be probably better, in my opinion, to avoid breaking existing code, but the semantics of the current function should be changed to mean a complete support of the protocol.

**Describe alternatives you've considered if any**
Complete implementation of the protocol by terminals, but this could not always be guaranteed, needed or wanted.

PR #958, opened 2024-12-16
Add `query_keyboard_enhancement_flags` to read enabled flags

the-mikedavis:
This is mostly a refactor of `supports_keyboard_enhancement` and its helper functions. We previously discarded the response to the `CSI ? u` query but this returns the currently active flags. This response value can be useful for a client to know which flags to emit to disable the keyboard enhancement protocol or to know which flags a terminal supports (after the client has pushed flags).

Fixes https://github.com/crossterm-rs/crossterm/issues/952

### DIFF
--- a/src/terminal/sys/unix.rs
+++ b/src/terminal/sys/unix.rs
@@ -1,5 +1,7 @@
 //! UNIX related logic for terminal manipulation.
 
+#[cfg(feature = "events")]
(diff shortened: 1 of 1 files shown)


TimonPost (approved):

TimonPost:
Ty 👍

PR #970, opened 2025-02-14
Add a feature flag for `derive_more` impls

the-mikedavis:
This adds a feature flag to allow opting out of the `derive_more` dependency and derivations added in #949.

Currently the flag is enabled by default as it sounds useful for out-of-the-box experience for ratatui users for example - I don't have strong opinions on its default state either way.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ path = "src/lib.rs"
 all-features = true
 
 [features]
-default = ["bracketed-paste", "events", "windows"]
+default = ["bracketed-paste", "events", "windows", "derive-more"]
 
 #! ### Default features
 ## Enables triggering [`Event::Paste`](event::Event::Paste) when pasting text into the terminal.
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -44,9 +44,12 @@ serde = ["dep:serde", "bitflags/serde"]
 ## Enables raw file descriptor polling / selecting instead of mio.
 use-dev-tty = ["filedescriptor", "rustix/process"]
 
+## Enables `is_*` and `as_*` helpers for event enums.
+derive-more = ["dep:derive_more"]
```
joshka, line 48: small nit: should this be `derive_more` to match the crate name, or `derive-more` to align with the feature flag conventions used here?

joshka (approved):
I'd suggest adding a small note in the commit (for future readers) that the discussion of why this change was made is in the other PR and summarize the reason.

I'm fine with this change otherwise, but will defer to @TimonPost to merge.

Cargo.toml
the-mikedavis: I think it's better to match the existing feature flags here but I don't have a strong preference

joshka: Yeah, I'm 70/30 on that (i.e. I agree with you, but wanted to check that that this was correct because I wasn't sure).

TimonPost (approved):

PR #974, opened 2025-03-05
Add copying to clipboard using OSC52

naseschwarz:
Many terminal emulators support copying text to clipboard using ANSI OSC Ps; PT ST with Ps = 5 2, see
https://invisible-island.net/xterm/ctlseqs/ctlseqs.html#h3-Operating-System-Commands

This enables copying even through SSH and terminal multiplexers.

I'm not too sure whether you want this feature or whether `src/terminal.rs` is the right place for it. I had just implemented it for gitui, which uses crossterm, and thought, it would be nice to have it upstream too.

It does not come for free, as it adds a new dependency to base64 which is not yet required by any of crossterm's dependencies.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,7 +47,11 @@ use-dev-tty = ["filedescriptor", "rustix/process"]
 ## Enables `is_*` helper functions for event enums.
 derive-more = ["dep:derive_more"]
 
+## Enables interacting with a host clipboard via [`clipboard`](clipboard/index.html)
+osc52 = ["dep:base64"]
+
 [dependencies]
+base64 = { version = "0.22", optional = true }
 bitflags = { version = "2.3" }
 derive_more = { version = "1.0.0", features = ["is_variant"], optional = true }
 document-features = "0.2.10"
@@ -112,3 +116,7 @@ required-features = ["events"]
 [[example]]
 name = "key-display"
 required-features = ["events"]
+
+[[example]]
+name = "copy-to-clipboard"
+required-features = ["osc52"]
--- a/src/clipboard.rs
+++ b/src/clipboard.rs
@@ -1,0 +1,308 @@
+//! # Clipboard
+//!
+//! The `clipboard` module provides functionality to work with a host clipboard.
+//!
+//! ## Implemented operations:
+//!
+//! - Copy: [`CopyToClipboard`](struct.CopyToClipboard.html)
+use base64::prelude::{Engine, BASE64_STANDARD};
+
+use std::fmt;
+use std::str::FromStr;
+
+use crate::{osc, Command};
+
+/// Different clipboard types
+///
+/// Some operating systems and desktop environments support multiple buffers
+/// for copy/cut/paste. Their details differ between operating systems.
+/// See <https://specifications.freedesktop.org/clipboard-spec/latest/>
+/// for a detailed survey of supported types based on the X window system.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum ClipboardType {
+    /// Default clipboard when using Ctrl+C or Ctrl+V
+    Clipboard,
+
+    /// Clipboard on Linux/X/Wayland when using selection and middle mouse button
+    Primary,
+
+    /// Other clipboard type not explicitly supported by crossterm
+    ///
+    /// See
+    /// [XTerm Control Sequences](https://invisible-island.net/xterm/ctlseqs/ctlseqs.html#h3-Operating-System-Commands)
+    /// for potential values.
+    ///
+    /// Note that support for these in terminal emulators is very limited.
+    Other(char),
+}
+
+impl From<&ClipboardType> for char {
+    fn from(val: &ClipboardType) -> Self {
+        match val {
+            ClipboardType::Clipboard => 'c',
+            ClipboardType::Primary => 'p',
+            ClipboardType::Other(other) => *other,
+        }
+    }
+}
+
+impl From<char> for ClipboardType {
+    fn from(value: char) -> Self {
+        match value {
+            'c' => ClipboardType::Clipboard,
+            'p' => ClipboardType::Primary,
+            other => ClipboardType::Other(other),
+        }
+    }
+}
+
+/// A sequence of clipboard types
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct ClipboardSelection(
+    /// An ordered list of clipboards which will be the destination for the copied selection.
+    ///
+    /// Order matters due to implementations deviating from the
+    /// [XTerm Control Sequences](https://invisible-island.net/xterm/ctlseqs/ctlseqs.html#h3-Operating-System-Commands)
+    /// reference. Some terminal emulators may only interpret the first character of this
+    /// parameter. For differences, see
+    /// [`CopyToClipboard` (Terminal Support)](struct.CopyToClipboard.html#terminal-support).
+    pub Vec<ClipboardType>,
+);
+
+impl ClipboardSelection {
+    /// Returns a String corresponsing to the "Pc" parameter of the OSC52
+    /// sequence.
+    fn to_osc52_pc(&self) -> String {
+        self.0.iter().map(Into::<char>::into).collect()
+    }
+}
+
+impl fmt::Display for ClipboardSelection {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.write_str(&self.to_osc52_pc())
+    }
+}
+
+impl FromStr for ClipboardSelection {
+    type Err = ();
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        Ok(ClipboardSelection(
+            s.chars().map(From::<char>::from).collect(),
+        ))
+    }
+}
+
+/// A command that copies to clipboard
+///
+/// This command uses OSC control sequence `Pr = 5 2` (See
+/// [XTerm Control Sequences](https://invisible-island.net/xterm/ctlseqs/ctlseqs.html#h3-Operating-System-Commands) )
+/// to copy data to the terminal host clipboard.
+///
+/// This only works if it is enabled on the user's terminal emulator. If a terminal multiplexer
+/// is used, the multiplexer must support it, too.
+///
+/// Commands must be executed/queued for execution otherwise they do nothing.
+///
+/// # Examples
+///
+/// ```no_run
+/// use crossterm::execute;
+/// use crossterm::clipboard::CopyToClipboard;
+/// // Copy foo to clipboard
+/// execute!(std::io::stdout(), CopyToClipboard::to_clipboard_from("foo"));
+/// // Copy bar to primary
+/// execute!(std::io::stdout(), CopyToClipboard::to_primary_from("bar"));
+/// ```
+///
+/// See also examples/copy-to-clipboard.rs.
+///
+/// # Terminal Support
+///
+/// The following table shows what destinations are filled by different terminal emulators when
+/// asked to copy to different destination sequences.
+///
+/// | Terminal (Version)    | dest ''   | dest 'c'  | dest 'p' | dest 'cp'     | dest'pc'      |
+/// | --------------------- | --------- | --------- | -------- | ------------- | ------------- |
+/// | xterm (397) *3        | primary   | clipboard | primary  | clipb., prim. | clipb., prim. |
+/// | Alacritty (0.15.1) *3 | clipboard | clipboard | primary  | clipb.        | prim.         |
+/// | Wezterm (*1) *3       | clipboard | clipboard | primary  | clipb.        | clipb.        |
+/// | Konsole (24.12.3) *3  | clipboard | clipboard | primary  | clipb., prim. | clipb., prim. |
+/// | Kitty (0.40.0) *3     | clipboard | clipboard | primary  | clipb.        | clipb.        |
+/// | foot (1.20.2) *3      | clipboard | clipboard | primary  | clipb., prim. | clipb., prim. |
+/// | tmux (3.5a) *2 *3     | primary   | clipboard | primary  | clipb., prim. | clipb., prim. |
+///
+/// Asterisks:
+/// 1. 20240203-110809-5046fc22
+/// 2. set-clipboard set to [external](https://github.com/tmux/tmux/wiki/Clipboard#how-it-works),
+///    i.e. this is OSC52 pass-through.
+/// 3. This was tested on wayland with the
+///    [primary selection protocol](https://wayland.app/protocols/primary-selection-unstable-v1)
+///    enabled.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct CopyToClipboard<T> {
+    /// Content to be copied
+    pub content: T,
+    /// Sequence of copy destinations
+    ///
+    /// Not all sequences are equally supported by terminal emulators. See
+    /// [`CopyToClipboard` (Terminal Support)](struct.CopyToClipboard.html#terminal-support).
+    pub destination: ClipboardSelection,
+}
+
+impl<T: AsRef<[u8]>> Command for CopyToClipboard<T> {
+    fn write_ansi(&self, f: &mut impl fmt::Write) -> fmt::Result {
+        write!(
+            f,
+            osc!("52;{destination};{encoded_text}"),
+            destination = self.destination.to_osc52_pc(),
+            encoded_text = BASE64_STANDARD.encode(&self.content)
+        )
+    }
+
+    #[cfg(windows)]
+    fn execute_winapi(&self) -> std::io::Result<()> {
+        use std::io;
+
+        Err(io::Error::new(
+            io::ErrorKind::Unsupported,
+            "Copying is not implemented for the Windows API.",
+        ))
+    }
+}
+
+impl<T: AsRef<[u8]>> CopyToClipboard<T> {
+    /// Construct a [`CopyToClipboard`] that writes content into the
+    /// "clipboard" (or 'c') clipboard selection.
+    ///
+    /// # Example
+    ///
+    /// ```no_run
+    /// use crossterm::{execute, Command};
+    /// use crossterm::clipboard::CopyToClipboard;
+    /// execute!(std::io::stdout(), CopyToClipboard::to_clipboard_from("foo"));
+    /// ```
+    pub fn to_clipboard_from(content: T) -> CopyToClipboard<T> {
+        CopyToClipboard {
+            content,
+            destination: ClipboardSelection(vec![ClipboardType::Clipboard]),
+        }
+    }
+
+    /// Construct a [`CopyToClipboard`] that writes content into the "primary"
+    /// (or 'p') clipboard selection.
+    ///
+    /// # Example
+    ///
+    /// ```no_run
+    /// use crossterm::execute;
+    /// use crossterm::clipboard::CopyToClipboard;
+    /// execute!(std::io::stdout(), CopyToClipboard::to_primary_from("foo"));
+    /// ```
+    pub fn to_primary_from(content: T) -> CopyToClipboard<T> {
+        CopyToClipboard {
+            content,
+            destination: ClipboardSelection(vec![ClipboardType::Primary]),
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_clipboard_string_to_selection() {
+        assert_eq!(
+            ClipboardSelection::from_str("p").unwrap(),
+            ClipboardSelection(vec![ClipboardType::Primary])
+        );
+        assert_eq!(
+            ClipboardSelection::from_str("").unwrap(),
+            ClipboardSelection(vec![])
+        );
+        assert_eq!(
+            ClipboardSelection::from_str("cp").unwrap(),
+            ClipboardSelection(vec![ClipboardType::Clipboard, ClipboardType::Primary])
+        );
+    }
+    #[test]
+    fn test_clipboard_selection_to_osc52_pc() {
+        assert_eq!(ClipboardSelection(vec![]).to_osc52_pc(), "");
+        assert_eq!(
+            ClipboardSelection(vec![ClipboardType::Clipboard]).to_osc52_pc(),
+            "c"
+        );
+        assert_eq!(
+            ClipboardSelection(vec![ClipboardType::Primary]).to_osc52_pc(),
+            "p"
+        );
+        assert_eq!(
+            ClipboardSelection(vec![ClipboardType::Primary, ClipboardType::Clipboard])
+                .to_osc52_pc(),
+            "pc"
+        );
+        assert_eq!(
+            ClipboardSelection(vec![ClipboardType::Clipboard, ClipboardType::Primary])
+                .to_osc52_pc(),
+            "cp"
+        );
+        assert_eq!(
+            ClipboardSelection(vec![ClipboardType::Other('s')]).to_osc52_pc(),
+            "s"
+        );
+    }
+
+    #[test]
+    fn test_clipboard_copy_string_osc52() {
+        let mut buffer = String::new();
+        super::CopyToClipboard {
+            content: "foo",
+            destination: ClipboardSelection(vec![ClipboardType::Clipboard]),
+        }
+        .write_ansi(&mut buffer)
+        .unwrap();
+        assert_eq!(buffer, "\x1b]52;c;Zm9v\x1b\\");
+
+        buffer.clear();
+        super::CopyToClipboard {
+            content: "foo",
+            destination: ClipboardSelection(vec![ClipboardType::Primary]),
+        }
+        .write_ansi(&mut buffer)
+        .unwrap();
+        assert_eq!(buffer, "\x1b]52;p;Zm9v\x1b\\");
+
+        buffer.clear();
+        super::CopyToClipboard {
+            content: "foo",
+            destination: ClipboardSelection(vec![ClipboardType::Primary, ClipboardType::Clipboard]),
+        }
+        .write_ansi(&mut buffer)
+        .unwrap();
+        assert_eq!(buffer, "\x1b]52;pc;Zm9v\x1b\\");
+
+        buffer.clear();
+        super::CopyToClipboard {
+            content: "foo",
+            destination: ClipboardSelection(vec![]),
+        }
+        .write_ansi(&mut buffer)
+        .unwrap();
+        assert_eq!(buffer, "\x1b]52;;Zm9v\x1b\\");
+    }
+
+    #[test]
+    fn test_clipboard_copy_string_osc52_constructor() {
+        let mut buffer = String::new();
+        super::CopyToClipboard::to_clipboard_from("foo")
+            .write_ansi(&mut buffer)
+            .unwrap();
+        assert_eq!(buffer, "\x1b]52;c;Zm9v\x1b\\");
+
+        let mut buffer = String::new();
+        super::CopyToClipboard::to_primary_from("foo")
+            .write_ansi(&mut buffer)
+            .unwrap();
+        assert_eq!(buffer, "\x1b]52;p;Zm9v\x1b\\");
+    }
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Unreleased
 
+## Added ⭐
+
+- Copy to clipboard using OSC52 (#974)
+
 # Version 0.28.1
 
 ## Fixed 🐛
--- a/README.md
+++ b/README.md
@@ -154,6 +154,7 @@ features = ["event-stream"]
 | `events`        | Reading input/system events (enabled by default) |
 | `filedescriptor` | Use raw filedescriptor for all events rather then mio dependency |
 | `derive-more`  | Adds `is_*` helper functions for event types |
+| `osc52`        | Enables crossterm::clipboard                 |
 
 
 To use crossterm as a very thin layer you can disable the `events` feature or use `filedescriptor` feature. 
@@ -172,6 +173,7 @@ This can disable `mio` / `signal-hook` / `signal-hook-mio` dependencies.
 | `futures-core` | For async stream of events                                                       | only with `event-stream` feature flag |
 | `serde`        | ***ser***ializing and ***de***serializing of events                              | only with `serde` feature flag        |
 | `derive_more`  | Adds `is_*` helper functions for event types                                     | optional (`derive-more` feature), included by default |
+| `base64`       | Encoding clipboard data for OSC52 sequences in crossterm::clipboard              | only with `osc52` feature flag        |
 
 ### Other Resources
 
--- a/examples/copy-to-clipboard.rs
+++ b/examples/copy-to-clipboard.rs
@@ -1,0 +1,46 @@
+//! Demonstrates copying a string to clipboard
+//!
+//! This example uses OSC control sequence `Pr = 5 2` (See
+//! <https://invisible-island.net/xterm/ctlseqs/ctlseqs.html#h3-Operating-System-Commands)>
+//! to copy data to the terminal host clipboard.
+//!
+//! This only works if it is enabled on the respective terminal emulator. If a terminal multiplexer
+//! is used, the multiplexer will likely need to support it, too.
+//!
+//! ```no_run
+//! cargo run --example copy-to-clipboard -- --clipboard "Some String"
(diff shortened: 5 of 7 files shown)


extrawurst:
@joshka would love your input on this

src/terminal.rs
```diff
@@ -396,6 +397,26 @@ impl<T: fmt::Display> Command for SetTitle<T> {
...
+/// Commands must be executed/queued for execution otherwise they do nothing.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct CopyToClipboard<T>(pub T);
+
+impl<T: AsRef<[u8]>> Command for CopyToClipboard<T> {
+    fn write_ansi(&self, f: &mut impl fmt::Write) -> fmt::Result {
+        const OSC52_DESTINATION_CLIPBOARD: char = 'c';
+        write!(
+            f,
+            "\x1b]52;{destination};{encoded_text}\x07",
```
joshka, line 413: There's a csi! macro. Maybe we should also add an osc! macro?

src/terminal.rs
```diff
@@ -396,6 +397,26 @@ impl<T: fmt::Display> Command for SetTitle<T> {
     }
 }
 
+/// A command that copies to clipboard
+///
+/// # Notes
+///
+/// Commands must be executed/queued for execution otherwise they do nothing.
```
joshka, line 404: Would be nice to link to the spec for this here to close the loop. Use the docs to tel how this works a bit.

Cargo.toml
```diff
@@ -48,6 +48,7 @@ use-dev-tty = ["filedescriptor", "rustix/process"]
 derive-more = ["dep:derive_more"]
 
 [dependencies]
+base64 = "0.22"
```
joshka, line 51: Regarding the dependency, I'd suggest adding a clipboard feature flag that enables base64

src/terminal.rs
```diff
@@ -396,6 +397,26 @@ impl<T: fmt::Display> Command for SetTitle<T> {
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct CopyToClipboard<T>(pub T);
+
+impl<T: AsRef<[u8]>> Command for CopyToClipboard<T> {
+    fn write_ansi(&self, f: &mut impl fmt::Write) -> fmt::Result {
+        const OSC52_DESTINATION_CLIPBOARD: char = 'c';
+        write!(
+            f,
+            "\x1b]52;{destination};{encoded_text}\x07",
+            destination = OSC52_DESTINATION_CLIPBOARD,
```
joshka, line 414: Is it worth exposing the possible destinations as an enum?

src/terminal.rs
```diff
@@ -396,6 +397,26 @@ impl<T: fmt::Display> Command for SetTitle<T> {
...
 
+/// A command that copies to clipboard
+///
+/// # Notes
+///
+/// Commands must be executed/queued for execution otherwise they do nothing.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct CopyToClipboard<T>(pub T);
+
+impl<T: AsRef<[u8]>> Command for CopyToClipboard<T> {
```
joshka, line 408: Any idea if there's a windows api equivalent to this?

src/terminal.rs
```diff
@@ -565,4 +586,13 @@ mod tests {
...
         assert!(!is_raw_mode_enabled().unwrap());
     }
+
+    #[test]
+    fn test_copy_string_osc52() {
+        let mut buffer = String::new();
+        super::CopyToClipboard("foo")
+            .write_ansi(&mut buffer)
+            .unwrap();
+        assert_eq!(buffer, "\x1b]52;c;Zm9v\x07");
```
joshka, line 596: I wonder if there's a way to actually test that this works by checking the clipboard somehow?

joshka (commented):
General support for the idea with a few questions / thoughts. `terminal` seems like the right module for this unless you were to add a clipboard module, but there's only one command likely to end up there, so that feels like probably a bit unnecessary.

src/terminal.rs
naseschwarz: Of course, you could spin up a terminal emulator and test it - one can request the clipboard via Ps = 5 2, Pd = '?':

>           If the second parameter is a ? , xterm replies to the host
>           with the selection data encoded using the same protocol.  It
>           uses the first selection found by asking successively for each
>           item from the list of selection parameters.

Originally, I'd guessed that this is not in the scope of these tests. However, it seems like the CI is running unit tests in predefined terminals with nocapture enabled. So, maybe for crossterm-rs, it is?

src/terminal.rs
naseschwarz: Just for reference: Windows Terminal supports (or will support) OSC52: https://devblogs.microsoft.com/commandline/windows-terminal-preview-1-23-release/

As I cannot test any Windows implementation, any implementation would probably need to be contributed by someone else.

src/terminal.rs
naseschwarz: If there is interest in doing so, I could. However, multiple clipboards is a Linux thing. Is there interest in me doing so?

I'm also not sure which of the options that should be exposed. I guess, primary and clipboard are practical options. But all the cut buffers? I'm not sure if there is anybody who uses those.

Cargo.toml
naseschwarz: I thought about it, but didn't want to pollute the feature flag space. However, I'd gladly add it.

src/terminal.rs
naseschwarz: I should add that I would guess that having OSC52 paste mode enabled should probably be considered a security issue. I certainly would not expect untrusted hosts to extract clipboard data just because I SSH into them. Alacritty, for example, does not enable this feature by default:

image

Therefore it's maybe not worth it implementing OSC52 paste for anything other than testing.

src/terminal.rs
naseschwarz: For reference:

image

This article agrees that clipboard and primary are reasonably well defined and would be obvious choices: https://specifications.freedesktop.org/clipboard-spec/latest/

I personally don't have a use for "primary" style selections in the terminal. However, applications that support mouse interactions probably do.

src/terminal.rs
naseschwarz: I added `osc!`. Thank you for this hint.

src/terminal.rs
naseschwarz: Done

Cargo.toml
naseschwarz: Done

src/terminal.rs
naseschwarz: I've also added the option to copy to primary. I don't think any of the other options would find a user, but if so, they're easy to add later given the current interface.

src/terminal.rs
naseschwarz: Reviewing the other tests, I don't think this is practical at the moment. Other tests that actually test interaction with the terminal seem to be disabled at the moment, as they fail in the CI:

image

If there is actual interest in testing this (@TimonPost?), I'd gladly contribute to a more involved testing effort and maybe even sponsor the required compute time. For this OSC52 copy, however, I'd be quite content with the unit test against the specification. There's no space for free parameters that could break this feature here and an integration test with terminals would mostly test whether the terminals changed something. Thus, crossterm would only really profit if the tests were run against four or five popular terminals. As far as I can tell, this is not the case at the moment?

naseschwarz: That being sad, there is one free parameter, I have neglected: Size restrictions on the content. I'd wager terminals limit this size. I have not tested where this limit is, as I only intend to copy short strings. Thus, one interesting parameter to test in a system test would be to test those limits on different terminal emulators.

src/terminal.rs
naseschwarz: I've looked into this on an EC2 machine and concluded that it's unlikely that I will provide a winapi implementation here for two reasons:

1. The modern windows terminal does not require WinAPI copying anymore anyhow but supports OSC52, so it would be a very niche improvement in the first place.
2. Copying via OSC52 is really only relevant for applications running on remote hosts, e.g. via SSH, which the Windows API will never support. **Other applications should really directly use the host API** in order to avoid an unsupported, disabled or length-limited OSC52 on any host, be it windows, linux or macos.

src/terminal.rs
joshka: I use primary style selections all the time. When selecting text in my terminal, I expect that I should be able to immediately paste that somewhere else without having to manually trigger some sort of copy command. I suspect that there are many older linux users in this boat (I'm a mac user and enjoy this interaction). I don't know anything about the cut buffer / secondary options, so it seems fine to leave these out until someone complains about them here.

Cargo.toml
joshka: The background for this was recent push back of adding unnecessary deps for apps that won't use those deps. Here the base64 lib has about 10x the download count of crossterm, so it's much less likely to be a problem, but I can see that this same complaint would have been coming if this wasn't feature flagged, so thanks for adding it.

src/terminal.rs
joshka: > Therefore it's maybe not worth it implementing OSC52 paste for anything other than testing.

I think I would probably consider adding this more for apps, but make a caveat in the documentation about how this feature is often disabled by the terminal emulator, multiplexer or protocol in use due to security concerns.

joshka (commented):
Lots of extra comments after reading the docs more and giving this a deeper review. Apologies that my first review was fairly surface level in comparison.

At a high level, I support the addition of this. To summarize my thoughts across the various comments raised here, I have some concern around what level of abstraction Crossterm should provide for this. I don't have any immediately concrete suggestions though, so looking for some back and forth of ideas before settling on the right approach. Feel free to shoot down any of the perspectives I've thrown here with your viewpoints - you're likely just as much or more of an expert on these than I am.

---

Regarding the paste side of this (passing ? as the data). It seems like it could be worth adding this usage of OSC 52 at the same time as adding this feature. WDYT?

          If the second parameter is a ? , xterm replies to the host
          with the selection data encoded using the same protocol.  It
          uses the first selection found by asking successively for each
          item from the list of selection parameters.

Also there's a clear clipboard function - worth worrying about?

          If the second parameter is neither a base64 string nor ? ,
          then the selection is cleared.

naseschwarz:
Hi @joshka,

> Lots of extra comments after reading the docs more and giving this a deeper review. Apologies that my first review was fairly surface level in comparison.

I appreciate this very much. Thank you for taking the time. (Also, I know the feeling of having to double down on a review *very* well. No worries. :) )

> At a high level, I support the addition of this. To summarize my thoughts across the various comments raised here, I have some concern around what level of abstraction Crossterm should provide for this.

I've had these, too. This whole change feels odd as clipboard support breaks a barrier in so far as that it is very different from the usual presentation and input mechanisms supplied by terminals and in this case crossterm. It is only really relevant to applications as a means for copying from remote hosts, where the only line is a TTY. Thus, while odd, signalling the copy operation in-band is the go-to solution.

> Regarding the paste side of this (passing ? as the data). It seems like it could be worth adding this usage of OSC 52 at the same time as adding this feature. WDYT?

I don't have much to add to my previously stated opinion. There is essentially no need for paste (as pasting can be done through the terminal emulator and normal input) and normalizing its use is a huge security concern. You can't even use it to cross-check OSC52, as some terminal emulators disable OSC52 paste while having OSC52 copy enabled. (And after this has been merged I'll probably try to get alacritty's behavior into more emulators. I was quite surprised by how many emulators actually allow paste.)

My suggestion to consolidate this would be to actually move the clipboard code into a separate `clipboard.rs` and make rename `ClipboardDestination` to `ClipboardSelection` such that the clipboard features stand out a bit more and it'll be easy to add `PasteFromClipboard` later, if one wants to do so. Would that be alright?

> Also there's a clear clipboard function - worth worrying about?
> 
> ```
>       If the second parameter is neither a base64 string nor ? ,
>       then the selection is cleared.
> ```

I guess you're asking because a `CopyToClipboard("", _)` will base64-encode to an empty second parameter, i.e. clear the clipboard instead of copying an empty string? I would have guessed that from a user's standpoint an empty clipboard and a clipboard containing an empty string would be equivalent.

However, it is true that for clipboard protocols, there generally is a difference between being cleared and containing an empty string.

That being said, I'm not sure how to address this given OSC52's specification. A wider API would only be relevant if there is a plan to support other clipboard protocols (like wayland, X.org, winapi, whatever macos is doing and whatever we want to implement ourselves) in the future. My view as a new contributor is that this is not in crossterm's scope.

If you disagree, we should discuss this, though. Frankly, I do see a potential use case for copying a text/uri-list that contains sftp-URIs for pasting into your favorite file browser for file download through SSH. The only limit here is that associating a MIME type is not supported by OSC52. That does not really keep us from widening the interface to something resembling `CopyToClipboard(T, MimeType, ClipboardSelection)` in order to support that in the future.

I'll respond to the other items in line or implement them.

Cheers and thanks again,

Johannes

joshka:
> My suggestion to consolidate this would be to actually move the clipboard code into a separate clipboard.rs and make rename ClipboardDestination to ClipboardSelection such that the clipboard features stand out a bit more and it'll be easy to add PasteFromClipboard later, if one wants to do so. Would that be alright?

Sounds good - this also simplifies the feature flag checks to just include the entire module.

> That being said, I'm not sure how to address this given OSC52's specification. A wider API would only be relevant if there is a plan to support other clipboard protocols (like wayland, X.org, winapi, whatever macos is doing and whatever we want to implement ourselves) in the future. My view as a new contributor is that this is not in crossterm's scope.

Your logic there seems sound to me.

src/terminal.rs
```diff
@@ -396,6 +400,62 @@ impl<T: fmt::Display> Command for SetTitle<T> {
     }
 }
 
+/// Different clipboard classes
+#[cfg(feature = "clipboard")]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum ClipboardDestination {
```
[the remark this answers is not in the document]
naseschwarz: I intentionally narrowed the API here as I found that `alacritty`, my terminal emulator, does not respect that. However, I've surveyed other applications and all seem to implement this correctly, so I'll likely change the API to a vector.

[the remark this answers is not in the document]
naseschwarz: [And I've reported this as an issue to alacritty/vte.](https://github.com/alacritty/vte/issues/134)

src/terminal.rs
```diff
@@ -396,6 +400,62 @@ impl<T: fmt::Display> Command for SetTitle<T> {
...
+/// is used, the multiplexer will likely need to support it, too.
+///
+/// Commands must be executed/queued for execution otherwise they do nothing.
+///
+/// # Examples
+///
+/// See examples/copy.rs for a working example.
+#[cfg(feature = "clipboard")]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct CopyToClipboard<T>(pub T, pub ClipboardDestination);
```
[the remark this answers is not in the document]
naseschwarz: > I think that once you have more than 1 field in a struct like this, I'd tend to want to give the values names for better clarity in the API (e.g. what about naming the fields `selection` and `data`). Doing this gives a place to hang doc comments off to help a user understand what the intended values will look like.

I'll gladly change this. :)

> Additionally, based on the spec docs, it sounds like this can accept 0 or more destinations, so it might be worth making this a `Vec<ClipboardSelection>` to support that. It should be noted that the comment about how xterm handles copying when this is not provided by defaulting to copy to the clipboard/primary/cut buffer 0 suggests that this should probably be also able to work without passing the selection. This implies adding a constructor or something like that that doesn't require choosing the selection. WDYT?

I'm up for making this a `Vec<...>` and providing constructors - this should improve readability. 

> That said, there's some tension between the spec default (s 0 = selection buffer + cut buffer 0) and the suggestions from the clipboard spec doc (primary + clipboard), so I'm not sure how best to resolve those in a way that makes sense for users.

I'm concerned about implicitly exposing an internal and remote OSC52 default if `vec![]` is passed. This behavior at least warrants documenting - in particular as the concrete result much more depends on whether a terminal emulator implements this default correctly.

I was about to write that it's not worth worrying about this, as X is being phased out and wayland only supports [clipboard](https://wayland.app/protocols/wayland#wl_data_device) and [primary](https://wayland.app/protocols/primary-selection-unstable-v1#zwp_primary_selection_device_v1:request:set_selection). Then I learned that Mac OS X actually [still seems (or at least in 2012 did) to support cut buffers](https://osxdaily.com/2012/02/01/use-a-secondary-cut-and-paste-function-to-avoid-overwriting-clipboard-contents/).

Two suggestions. Either we disallow empty vectors as a clipboard selection parameter or we deal with it by documenting that the behavior when passing a `vec![]` is entirely defined by the terminal emulator.

> Would naming this `CopySelection` make sense here, and then the parameter be named `SelectionBuffer`?

Will do.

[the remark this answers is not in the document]
joshka: > deal with it by documenting that the behavior when passing a vec![] is entirely defined by the terminal emulator.

I think this is probably the right approach

src/terminal.rs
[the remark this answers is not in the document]
naseschwarz: Well, it was worth a try - wontfix from alacritty/vte. :shrug:

src/terminal.rs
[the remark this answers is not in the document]
naseschwarz: So, I have collected a bit of not so nice data: How terminal emulators interpret '', 'cp'  and 'pc' destinations is a wild thing:

The following table shows what destinations are filled by different terminal emulators when asked to copy to different destionation sequences.

| Terminal (Version)  | Test System | Copy to ''  | Copy to 'c'  | Copy to 'p' | Copy to 'cp'       | Copy to 'pc'       |
| ------------------- | ----------- | ----------- | ------------ | ----------- | ------------------ | ------------------ |
| xterm (397)         | 1           | primary     | clipboard    | primary     | clipboard, primary | clipboard, primary |
| Alacritty (0.15.1)  | 1           | clipboard   | clipboard    | primary     | clipboard          | primary            |
| Wezterm (*1)        | 1           | clipboard   | clipboard    | primary     | clipboard          | clipboard          |
| Konsole (24.12.3)   | 1           | clipboard   | clipboard    | primary     | clipboard, primary | clipboard, primary |
| Kitty (0.40.0)      | 1           | clipboard   | clipboard    | primary     | clipboard          | clipboard          |
| foot (1.20.2)       | 1           | clipboard   | clipboard    | primary     | clipboard, primary | clipboard, primary |
| tmux (3.5a)         | 1, *2       | primary     | clipboard    | primary     | clipboard, primary | clipboard, primary |

Test systems were the following:
1. Wayland with [primary selection protocol](https://wayland.app/protocols/primary-selection-unstable-v1) enabled.

Asterisks:
1. 20240203-110809-5046fc22
2. set-clipboard set to external, i.e. this is OSC52 pass-through

At this point, I'm unsure whether implementing destination sets is really a good idea. At least in the status quo, it will not work reliably and if an application actually needs to copy to both 'c' and 'p' it should issue two separate sequences.

src/terminal.rs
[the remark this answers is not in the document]
joshka: Makes me wonder a bit whether there should be a caniuse type site for terminal emulators.

src/terminal.rs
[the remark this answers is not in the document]
joshka: Would it be reasonable to make this a vec internally, but only provide constructors that set individual items (noting the reasons)?

src/terminal.rs
TimonPost: It is fine to land it like this we have other features that do not have a winapi equivalent. We are cross-platform but should not limit features to legacy IMO. As long there is a good message/docs/error its fine.

src/terminal.rs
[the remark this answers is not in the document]
naseschwarz: I have such an implementation ready. I have not pushed it as it feels like a lot of code for very little advantage. But if you think there will be a case where people use these, I'll publish it once I've looked it over once more.

src/terminal.rs
TimonPost: It looks like the windows terminal already has support for it they just exposed a option to configure it in windows terminal 1.23.

naseschwarz: Yes, it does. By default, at least in Windows 11, all applications are using new Terminal renderer, even `cmd.exe`. You'll have to manually switch over to the old one for the legacy terminal.

src/terminal.rs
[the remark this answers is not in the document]
TimonPost: I dont have a strong opionion on it, when making decisions like this I would like to look at what is the most common usecase. If its a simple clear API with overall support I prefer that over some more obscure support and complicated interface. e.g people can always run the command twice to copy to primary and clipboard right?

src/terminal.rs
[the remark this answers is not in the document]
naseschwarz: Well, terminfo sadly is incapable of dealing with software versions and the `Ms` variable (which is essentially OSC52) is documented only for weird terminals. But that might be a start. ;)

naseschwarz:
So, here's a new API that is a bit wider but offers and prioritizes simple constructors. I guess there will be criticism, I'm still open to any changes!

Commands are now primarily constructed using `CopyToClipboard::into_clipboard_from(T)` or `CopyToClipboard::into_primary_from(T)`. Thus, the interface is straightforward. Users can also build custom `CopyToClipboard`s and setting `destination` to any sequence of clipboards, even the empty sequence.

Now this covers the whole OSC52 copy feature set, I guess, without too many visible complications.

@joshka: Might I ask you to have another look at this if you can spare the time? Thank you, also for your excellent feedback in the previous two rounds!

kovidgoyal:
Just FYI, pasting is best supported with permissions, just as it is on the web. And copy/paste for different mimetypes is implemented using the OSC 5522 protocol, documented at:
https://sw.kovidgoyal.net/kitty/clipboard/

src/clipboard.rs
```diff
@@ -0,0 +1,294 @@
...
+}
+
+/// A sequence of clipboard types
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct ClipboardSelection(
+    /// This being an ordered sequence is important due to deviations from the
+    /// [XTerm Control Sequences](https://invisible-island.net/xterm/ctlseqs/ctlseqs.html#h3-Operating-System-Commands)
+    /// reference. Some terminal emulators may only interpret the first
+    /// character of this parameter.
+    /// For differences, see [`CopyToClipboard` (Terminal Support)](struct.CopyToClipboard.html#terminal-support).
```
joshka, line 63: This jumps into an apology for the field not being something without first stating what it is. There's a sort of implied context which is known if you know how this works already, but if you're a user trying to understand how this works may be missing.

Something like "An ordered list of clipboards which will be the destination for the copied selection.

Order matters because the support for ..."

src/clipboard.rs
```diff
@@ -0,0 +1,294 @@
...
+        Ok(ClipboardSelection(
+            s.chars().map(From::<char>::from).collect(),
+        ))
+    }
+}
+
+/// A command that copies to clipboard
+///
+/// # Notes
+///
```
joshka, line 85:
```suggestion
```

src/clipboard.rs
```diff
@@ -0,0 +1,294 @@
...
+/// This command uses OSC control sequence `Pr = 5 2` (See
+/// [XTerm Control Sequences](https://invisible-island.net/xterm/ctlseqs/ctlseqs.html#h3-Operating-System-Commands) )
+/// to copy data to the terminal host clipboard.
+///
+/// This only works if it is enabled on the respective terminal emulator. If a terminal multiplexer
```
joshka, line 90:
```suggestion
/// This only works if it is enabled on the user's terminal emulator. If a terminal multiplexer
```

src/clipboard.rs
```diff
@@ -0,0 +1,294 @@
...
+/// is used, the multiplexer will likely need to support it, too.
```
joshka, line 91:
```suggestion
/// is used, the multiplexer must support it, too.
```

src/clipboard.rs
```diff
@@ -0,0 +1,294 @@
...
+/// ```no_run
+/// use crossterm::execute;
+/// use crossterm::clipboard::CopyToClipboard;
+/// // Copy foo to clipboard
+/// execute!(std::io::stdout(), CopyToClipboard::into_clipboard_from("foo"));
+/// // Copy bar to primary
+/// execute!(std::io::stdout(), CopyToClipboard::into_primary_from("bar"));
+/// ```
+///
+/// See also examples/copy.rs.
```
joshka, line 106: If renaming, clipboard
```suggestion
/// See also examples/clipboard.rs.
```

src/clipboard.rs
```diff
@@ -0,0 +1,294 @@
...
+///
+/// | Terminal (Version)  | Test System | Copy to ''  | Copy to 'c'  | Copy to 'p' | Copy to 'cp'       | Copy to 'pc'       |
+/// | ------------------- | ----------- | ----------- | ------------ | ----------- | ------------------ | ------------------ |
+/// | xterm (397)         | 1           | primary     | clipboard    | primary     | clipboard, primary | clipboard, primary |
+/// | Alacritty (0.15.1)  | 1           | clipboard   | clipboard    | primary     | clipboard          | primary            |
+/// | Wezterm (*1)        | 1           | clipboard   | clipboard    | primary     | clipboard          | clipboard          |
+/// | Konsole (24.12.3)   | 1           | clipboard   | clipboard    | primary     | clipboard, primary | clipboard, primary |
+/// | Kitty (0.40.0)      | 1           | clipboard   | clipboard    | primary     | clipboard          | clipboard          |
+/// | foot (1.20.2)       | 1           | clipboard   | clipboard    | primary     | clipboard, primary | clipboard, primary |
+/// | tmux (3.5a)         | 1, *2       | primary     | clipboard    | primary     | clipboard, primary | clipboard, primary |
```
joshka, line 121: To make this narrower and fit better in the line width of 100 chars, drop the test system column (move it underneath?) and perhaps the "Copy to" part of the text

src/clipboard.rs
```diff
@@ -0,0 +1,294 @@
...
+///
+/// Test systems were the following:
+/// 1. Wayland with [primary selection protocol](https://wayland.app/protocols/primary-selection-unstable-v1) enabled.
+///
+/// Asterisks:
+/// 1. 20240203-110809-5046fc22
+/// 2. set-clipboard set to external, i.e. this is OSC52 pass-through
```
joshka, line 128: I'm not sure what this means

src/clipboard.rs
```diff
@@ -0,0 +1,294 @@
...
+    /// "clipboard" (or 'c') clipboard selection.
+    ///
+    /// # Example
+    ///
+    /// ```no_run
+    /// use crossterm::{execute, Command};
+    /// use crossterm::clipboard::CopyToClipboard;
+    /// execute!(std::io::stdout(), CopyToClipboard::into_clipboard_from("foo"));
+    /// ```
+    pub fn into_clipboard_from(content: T) -> CopyToClipboard<T> {
```
joshka, line 172: bikeshed: I think maybe to_clipboard / to_primary might be obvious enough names here. into_ gives greater connotations of conversion of the type into another type, which is not happening here. with_ would be the obvious builder style name, but that feels wrong here to me too.

naseschwarz:
> Just FYI, pasting is best supported with permissions, just as it is on the web. And copy/paste for different mimetypes is implemented using the OSC 5522 protocol, documented at: https://sw.kovidgoyal.net/kitty/clipboard/

Hi @kovidgoyal. Thank you for your input on paste, I appreciate it! It seems that, at the moment, this is only implemented in kitty, right?

kovidgoyal:
On Wed, Mar 19, 2025 at 11:45:04PM -0700, Johannes Agricola wrote:
> naseschwarz left a comment (crossterm-rs/crossterm#974)
>
> > Just FYI, pasting is best supported with permissions, just as it is on the web. And copy/paste for different mimetypes is implemented using the OSC 5522 protocol, documented at: https://sw.kovidgoyal.net/kitty/clipboard/
>
> Hi @kovidgoyal. Thank you for your input on paste, I appreciate it! It seems that, at the moment, this is only implemented in kitty, right?

Yes, so far as I know OSC 5522 is only supported in kitty, although, I
haven't really kept track.

However, since you are designing an API you might want to keep this in mind for the future.

examples/copy.rs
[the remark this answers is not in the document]
naseschwarz: Maybe this would be a good addition when paste get's implemented to show a fully interactive clipboard roundtrip. I've renamed it to `copy-to-clipboard.rs` for clarity, though.

I find this tool actually quite helpful as is. It allows me to quickly check crossterm's implementation against different terminal emulators without needing further interaction.

src/clipboard.rs
naseschwarz: I've removed and abbreviated a bit. I guess this is even more readable, as the most important point of this table is that only `Copy to 'c'` and `Copy to 'p'` are consistent across terminals, which is now even more prominent.

src/clipboard.rs
naseschwarz: Thanks, done.

src/clipboard.rs
naseschwarz: Thanks, done.

src/clipboard.rs
naseschwarz: Thanks, done.

src/clipboard.rs
```diff
@@ -0,0 +1,294 @@
...
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum ClipboardType {
+    /// Default clipboard when using Ctrl+C or Ctrl+V
+    Clipboard,
+
+    /// Clipboard on Linux/X/Wayland when using selection and middle mouse button
+    Primary,
+
+    /// Other clipboard type not explicitly supported by crossterm
+    /// See
```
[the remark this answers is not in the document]
naseschwarz: Thanks, done.

src/clipboard.rs
```diff
@@ -0,0 +1,294 @@
...
+//! - Copy: [`CopyToClipboard`](struct.CopyToClipboard.html)
+use base64::prelude::{Engine, BASE64_STANDARD};
+
+use std::fmt;
+use std::str::FromStr;
+
+#[doc(no_inline)]
+use crate::{osc, Command};
+
+/// Different clipboard types
```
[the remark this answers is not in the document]
naseschwarz: Thanks, done.

src/clipboard.rs
```diff
@@ -0,0 +1,294 @@
...
+//!
+//! ## Implemented operations:
+//!
+//! - Copy: [`CopyToClipboard`](struct.CopyToClipboard.html)
+use base64::prelude::{Engine, BASE64_STANDARD};
+
+use std::fmt;
+use std::str::FromStr;
+
+#[doc(no_inline)]
```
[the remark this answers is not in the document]
naseschwarz: Thanks, done.

naseschwarz:
Huge thanks again, @joshka. I've implemented all suggestions except for the more involved `clipboard.rs` example.

I also noticed that the CI isn't running clippy on all features and replaced my `ToString` implementation with `ClipboardSelection::to_osc52_pc()`.

joshka (approved):

archseer:
Worth noting that OSC52 support was previously rejected due to new dependencies (the exact same ones this PR is using): https://github.com/crossterm-rs/crossterm/pull/697

Cargo.toml
```diff
@@ -47,7 +47,11 @@ use-dev-tty = ["filedescriptor", "rustix/process"]
 ## Enables `is_*` helper functions for event enums.
 derive-more = ["dep:derive_more"]
 
+## Enables interacting with a host clipboard via [`clipboard`](clipboard/index.html)
+clipboard = ["dep:base64"]
```
TimonPost, line 51: Since its only osc52 perhaps we can name the feature flag to reflect that? Otherwise folks may think its a general clipboard feature which can have many kinds of implementations.

TimonPost (approved):
Yes indeed, previously dismissed the addition in the other PR, and similar concerns are raised here. In this PR we went for OSC52 support only without windows support, and compared the other PR this PR has done more work in documenting/testing the edge cases. On a second thought, with clear documentation and the feature behind a feature flag, I think its fine to have those features in the library given they are terminal protocols which crossterm aims to help simplifying access to. 

LGTM, just one comment regarding the feature flag.

Cargo.toml
naseschwarz: Thanks for your time and the approval, @TimonPost. I'll gladly yield to renaming the flag to `osc52`, see my latest change.

Apart from this, the latest change to the MR is rebasing onto master and adding an entry to `CHANGELOG.md`. I hope that's fine.

TimonPost:
Thanks for your patience and work!

PR #976, opened 2025-03-14
Fix clippy precedence hint

naseschwarz:
Stable clippy on windows understandibly warns about unclear precedence in this line.

This is currently beaking new CI runs if left unfixed.

### DIFF
--- a/src/cursor/sys/windows.rs
+++ b/src/cursor/sys/windows.rs
@@ -190,7 +190,9 @@ impl ScreenBufferCursor {
     fn save_position(&self) -> std::io::Result<()> {
         let position = self.position()?;
 
-        let bits = u64::from(u32::from(position.x as u16) << 16 | u32::from(position.y as u16));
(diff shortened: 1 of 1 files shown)


naseschwarz:
Warning for reference:

image

src/cursor/sys/windows.rs
```diff
@@ -190,7 +190,7 @@ impl ScreenBufferCursor {
     fn save_position(&self) -> std::io::Result<()> {
         let position = self.position()?;
 
-        let bits = u64::from(u32::from(position.x as u16) << 16 | u32::from(position.y as u16));
+        let bits = u64::from((u32::from(position.x as u16) << 16) | u32::from(position.y as u16));
```
joshka, line 193: Nit: Personally I'd generally add a small `let x = ...` to make it clearer (and avoid someone naively removing the parentheses here), but it's not a big deal.

joshka (approved):

joshka (approved):

PR #979, opened 2025-03-22
fix KeyModifiers Display impl

Flokkq:
### Fix KeyModifiers Display impl  

**Issue:**  
The `KeyModifiers` `Display` implementation incorrectly formats modifier keys due to a misplaced flag, causing the output to be concatenated without `+` separators.  

**Current output:** `"ShiftControlOption"`  
**Expected output:** `"Shift+Control+Option"`  

**Fix:**  
- Moved `first = false;` outside the if condition to ensure separator placement after the first iteration  
- Added a test (needs verification on Windows + Linux)

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -870,8 +870,9 @@ impl Display for KeyModifiers {
         for modifier in self.iter() {
             if !first {
                 f.write_str("+")?;
-                first = false;
             }
+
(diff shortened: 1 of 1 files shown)


joshka (approved):
Neat. Thanks :)

PR #981, opened 2025-03-28
CI: Add missing continue-on-error annotation

the-mikedavis:
It looks like the `continue-on-error` for windows with no default features was accidentally left off. I see the nightly build for this step failing today due to a transitive dev dependency: https://github.com/helix-editor/crossterm/actions/runs/14134310684/job/39602227594. Adding the continue-on-error like the other CI tasks papers over that


joshka:
LGTM - Thanks :)

PR #982, opened 2025-04-01
update rustix to 1.0

WindSoilder:
Hi, I noticed that the latest version of rustix is 1.0.4. I think it would be good to update it.

This way, if other crates depend on the latest rustix (e.g., tempfile), we can reduce duplicate compilation.


joshka:
There's a bunch of build failures. Can you please check these?

WindSoilder:
Hi, Sorry for the delay.  I renamed an enum variant and it passes the CI.

joshka:
No probs. Thanks for the PR

Issue #608, opened 2021-10-03
Consider using `spawn` and `channel` in specific async runtime.


Berrysoft:
I see that `EventStream` is using `std::thread` and `std::sync::mpsc` at now. It is of course an improvement according to #448. However, I'd like to suggest a possible improvement, with the usage of `spawn` and `channel`s in specific async runtime, for example, `tokio` or `async-std`. Therefore, the async runtime could manage the event stream thread.

The replacement in `tokio` is [`spawn`](https://docs.rs/tokio/1.12.0/tokio/fn.spawn.html) and [`mpsc`](https://docs.rs/tokio/1.12.0/tokio/sync/mpsc/index.html).

The replacement in `async-std` is [`spawn`](https://docs.rs/async-std/1.10.0/async_std/task/fn.spawn.html) and [`channel`](https://docs.rs/async-std/1.10.0/async_std/channel/index.html).

Both of them should be drop-in replacement, rewritting some functions to async ones. I'd like to open a pull request later, trying to add the implementation as an optional feature, and keep the original implementation as a fallback one.

BoolPurist:
I agree that the underlying stream should not used to spawn a naked OS-thread outside any async runtime.

The (previous attempt to fix this)[https://github.com/crossterm-rs/crossterm/pull/609] seems to go back quite a while.

I am willing to pick this up from here. 

However, I have a few questions before starting the rewrite.

1. Is there now any interest in the rewrite ?
2. How to do it in terms of the compile features ? I have the following suggestion.
 - Only supporting tokio under the flag "event-stream"
 - The flag "event-stream" is working the context of tokio. 
   One new compiler flag "asyn-std" is then using asyn-std.
 - Two new flags "tokio" and "asyn-std" in which tokio and asyn-std are respectively used. 
   In this variant, I would deprecate the flag "event-stream" and type EventStream simply 
   since the implementation of this type with naked OS thread should not be used anymore 
   going forward in my opinion.

Berrysoft:
> Is still there now any interest in the rewrite?

No, sorry. Now I think that it is easy enough for the users of tokio or other async runtimes to write another stream without spawing new threads.

BoolPurist:
> > Is still there now any interest in the rewrite?
> 
> No, sorry. Now I think that it is easy enough for the users of tokio or other async runtimes to write another stream without spawing new threads.

Fair enough.  
One question remains for me:  shouldn't the usage of this feature then be discouraged/deprecated ?

[...]

PR #987, opened 2025-04-25
calloop EventSource implementation

ogios:
I have read the issue https://github.com/crossterm-rs/crossterm/issues/608 which refers to `EventStream`, and i've also looked into ratatui-calloop, they both creates another thread for the events reading.

This implementation wraps the fd of `tty_fd` and `UnixStream` for `SIGWINCH` with `calloop::generic::Generic`, and implements `calloop::EventSource`.
It's a runtime specific impl so i created a separate mod for it.

Others to be done(under consideration):
- disable `INTERNAL_EVENT_READER` and global functions related to it, `read`, `poll`, `position`, `query_keyboard_enhancement_flags`.
- handle `InternalEvent` only public to crate, like `position()`, `query_keyboard_enhancement_flags()`

> currently i have InternalEvent public for testing with `cl_demo.rs`


ogios:
I cannot find a proper way to deal with `INTERNAL_EVENT_READER`, `read`, `poll`, `position`, `query_keyboard_enhancement_flags`, etc.
I would want to disable these global access and static variables while i can handle the event reader and the internal events in the runtime, but it's too much of a change, and not a call for me.

joshka:
Hey there, it would be useful to take a step back from the solution and talk a bit more about the context of the problem that's being solved here. I'm familiar only a little with calloop. I wrote ratatui-calloop as an experiment to play with ideas around how event loops should be structured in apps, but haven't really progressed it beyond that point. I'd expect others are probably even less familiar with calloop than me, so it's worth giving some background about why this is useful to have in crossterm.

I expect there probably to be a bit of push back on implementing this that's related to keeping things in crossterm simple and not increasing the number of dependencies, so it's worth considering alternatives to this solution.

I wonder if it would be possible to introduce this outside of crossterm rather than bringing it into the library. I haven't yet started looking at the code to understand whether this is a reasonable question though, so I'm guessing that probably not. But if there is, then I'd suggest making a secondary library that adds this functionality and exposing the necessary internals of crossterm if needed.

PR #988, opened 2025-04-30
README: Fix the name of GNOME Terminal

zeenix:
Since there is no such thing as a `Ubuntu Desktop Terminal`, I'm guessing this was meant to be `GNOME Terminal` (aka `gnome-terminal`) since that's what is shipped on Ubuntu Desktop by default. So let's be correct and specific. :)


joshka:
thanks

PR #991, opened 2025-05-19
feat: make `document-features` crate optional

mfontanini:
The `document-features` crate is unnecessarily included as a default dependency. This PR introduces a new feature flag `document-features` that makes it optional, which is part of the default feature flag set.

I would be inclined to remove this feature from the `default` set but I'll leave that decision to the maintainers.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ path = "src/lib.rs"
 all-features = true
 
 [features]
-default = ["bracketed-paste", "derive-more", "document-features", "events", "windows", "derive-more"]
+default = ["bracketed-paste", "derive-more", "events", "windows"]
 
 #! ### Default features
 ## Enables triggering [`Event::Paste`](event::Event::Paste) when pasting text into the terminal.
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -21,7 +21,7 @@ path = "src/lib.rs"
 all-features = true
 
 [features]
-default = ["bracketed-paste", "events", "windows", "derive-more"]
+default = ["bracketed-paste", "derive-more", "document-features", "events", "windows", "derive-more"]
```
joshka, line 24: It's probably reasonable to move this out of the default features as it's only really useful when compiling the documentation (which enables all-features).

src/lib.rs
```diff
@@ -227,7 +227,7 @@
 //! }
 //!```
 //! ## Feature Flags
-#![doc = document_features::document_features!()]
+#![cfg_attr(feature = "document-features", doc = document_features::document_features!())]
```
joshka, line 230: To make the section disappear when not enabled rather than having a heading which has no content, also slap this on the header:
```rus
#[cfg_attr(feature = "document-features", doc = "## Feature Flags")]
```

joshka (changes requested):
A couple of small changes

src/lib.rs
mfontanini: Smart! I noticed this and wanted to get rid of it but it didn't occur to me to do this.

joshka (approved):
LGTM

Issue #834, opened 2023-10-25
Unsupported input escape sequences reported as keypresses


Timmmm:
If you use a feature that crossterm doesn't support directly, for example Kitty's graphics protocol, then the terminal will send escape sequences that crossterm doesn't understand.

In that case I would expect `enum Event` to have a `UnrecognisedEscapeSequence(String)` variant. However instead it seems to just report all of the characters in the escape sequence as `Event::Key`.

To reproduce, modify the `event-read.rs` example to send an image using the Kitty graphics protocol. You'll end up with something like this:

![image](https://github.com/crossterm-rs/crossterm/assets/376842/66d80880-e6aa-4ca5-ab41-4e7d6ec2c355)

Note all the `KeyEvent`s are really the terminal's confirmation that the image has been shown.

PR #992, opened 2025-05-23
feat: ANSI APC implementation as a new event type + it's parsing implementation

monoamine11231:
**What is this?**
This is an implementation for parsing ANSI APC (Application Program Control) sequences into a new event type. The event type in the `enum Event` is named for `ApplicationProgramControl` and holds a string. ANSI APC sequences have the form of: 
```
<ESC>_<STRING><ESC>\
```
The parsing method is implemented only for the parser-unix submodule as it's used only in the UNIX-sphere (see below).  

**Why?**
The Kitty terminal uses APC sequences to encode status responses for the transfer and display of image data to and by the terminal. Currently there are no other solutions for parsing APC sequences by the `crossterm` package and such APC sequences sent, e.g by the Kitty terminal, are represented as a sequence of key-presses; see #834.

The only possibility for the end-user to parse such APC sequences with the `crossterm` package is to write an additional parser on top of the produced `KeyEvent` sequence which is not optimal.

**Discussion**
- More suitable name for the event type?
- Creating a more general event type for unrecognized escape sequences as suggested by #834?
-  Isn't really a cross-platform feature as only the Kitty terminal currently uses APC

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -54,6 +54,7 @@
 //!             #[cfg(feature = "bracketed-paste")]
 //!             Event::Paste(data) => println!("{:?}", data),
 //!             Event::Resize(width, height) => println!("New size {}x{}", width, height),
+//!             Event::ApplicationProgramCommand(command) => println!("New APC {}", command),
 //!         }
 //!     }
 //!     execute!(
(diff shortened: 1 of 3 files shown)


siennathesane:
@TimonPost are you able to take a look at this and see if it needs anything?

Issue #998, opened 2025-07-14
Refactor event module


joshka:
**Is your feature request related to a problem? Please describe.**
The event module has about 1750 LoC, which seems a bit large.

**Describe the solution you'd like**
It would be nice to split this out into smaller easier to read / navigate / maintain modules. These modules will have tests that are nearer the code that is being tested and generally have a high degree of coherence.
- internal (poll_internal, read_internal, InternalEvents etc)
- key_event
- key_code
- key_modifiers
- mouse_event
- commands (maybe)
- event (just for the Event enum)
- keyboard_flags

These would be for the purpose of developer happiness only and won't affect any external API changes. Do this by extracting the modules as private / pub(crate), and adding the necessary re-exports in the event module to maintain the existing api.

**Describe alternatives you've considered if any**

**Additional context**
Anything else?

PR #999, opened 2025-07-14
refactor: move internal items from event to event::internal

joshka:
This is a pure refactor with no change to the external API

Step 1 in #998


gavincrawford:
LGTM other than minor style stuff I commented about. Makes a lot of sense to refactor all of these.

joshka:
Makes sense. I figured that as well (didn't make the change as I wasn't 100% sure), having a second set of eyes on it brings the right amount of clarity.

PR #1007, opened 2025-08-04
terminal:handle zero dimensions from tcgetwinsize

alexs-sh:
Hello, 

This PR adds default (ANSI terminal) values in case when `tcgetwinsize` returns `Ok` with columns or rows equal to 0. In practice, Ok((0,0)) often results in crashes (frequently) or rendering issues (more rarely) when running applications over a serial line.

Meanwhile, applications like `vi` and `vim` work well in the same environment. The reason is they rely on multiple sources to determine the terminal size: tcgetwinsize, environment variables, terminfo, fallbacks to defaults, and more. They use an enhanced and layered mechanism for detecting the terminal dimensions.

Since implementing such a mechanism could take quite some time, I'd prefer to provide only default values for now:

- they can serve as a starting point for discussing the problem and solutions
- they can help prevent crashes and issues in existing applications
- they are simple and predictable

Some links on issues / PRs caused by Ok with zero columns/rows:

- Helix - https://github.com/helix-editor/helix/pull/14050
- Helix - https://github.com/helix-editor/helix/issues/14101
- aichat - https://github.com/sigoden/aichat/pull/1366

Some links on vi / vim:

- getting terminal size in vim: https://github.com/vim/vim/blob/b88f9e4a04ce9fb70abb7cdae17688aa4f49c8c9/src/os_unix.c#L4299
- using defaults in vi: https://github.com/mirror/busybox/blob/371fe9f71d445d18be28c82a2a6d82115c8af19d/editors/vi.c#L4814

### Note

This PR is mostly to demonstrate the problem, provide a quick (though likely not perfect) workaround, and help us find a correct and acceptable solution for everyone. It seems we should be careful here, as this kind of change could affect existing applications.

### Example 
A small proof. Helix before the change:

image

Helix with the change:
image

Please let me know if any additional info, test, information are needed.

Thank you

### DIFF
--- a/src/terminal/sys/unix.rs
+++ b/src/terminal/sys/unix.rs
@@ -98,7 +98,45 @@ pub(crate) fn window_size() -> io::Result<WindowSize> {
 #[allow(clippy::useless_conversion)]
 pub(crate) fn size() -> io::Result<(u16, u16)> {
     if let Ok(window_size) = window_size() {
-        return Ok((window_size.columns, window_size.rows));
+        // Let's add default (ANSI terminal) values in case when 'tcgetwinsize' returns Ok with columns and/or rows
+        // equal to 0. In practice, Ok((0,0)) often results in crashes (frequently) or rendering issues (more rarely)
+        // when running applications over a serial line.
+        //
+        // Meanwhile, applications like vi and vim work well in the same environment. The reason is they rely on
+        // multiple sources to determine the terminal size: tcgetwinsize, environment variables, terminfo, fallbacks to
+        // defaults, and more. They use an enhanced and layered mechanism for detecting the terminal dimensions, which
+        // seems to be implemented in this library.
+        //
+        // Since implementing such a mechanism could take quite some time, I'd prefer to provide
+        // only default values for now:
+        // - they can serve as a starting point for discussing the problem and solutions
+        // - they can help prevent crashes in existing applications
+        // - they are simple and predictable
+        //
+        // Some links on issues / PRs caused by Ok with zero columns/rows:
+        // - Helix - https://github.com/helix-editor/helix/pull/14050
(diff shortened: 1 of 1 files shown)


alexs-sh:
Agree. Returning `Err` for zero dimensions is a good approach, if `Ok(0,0)` was implemented by mistake and doesn’t have some hidden meaning

alexs-sh:
It seems we lose the zero-dimensions checks here: #790

benjajaja:
@alexs-sh mhh, hard to reconstruct, but I think the thought probably was: the new `window_size()` returns a `WindowSize` with pixel size on top of col-row-counts, and pixel size could be zero when col-row-counts is not zero, in some terminals. So the zero-check would perhaps be out of place in `window_size()`, but I don't see any reason why it shouldn't simply be in `size()`. It was probably a mistake.

On another note, I am not using `window_size()` anymore. `TIOCGWINSZ` (not via crossterm but termios directly) is just a fallback in my case. Perhaps `window_size()` should be removed or marked deprecated, IIRC only a few terminals actually return pixel sizes.

alexs-sh:
Seems the hardest part is that three levels may be responsible for handling it. 

- the lower level with `ioctl`. But it seems we can’t do anything there. 
- Crossterm itself. It could provide errors, default values, or extended logic to handle such situations
- the upper-level code. We can just delegate the problem (Ok(0,0))  there and preserve Crossterm as is. However, it seems the upper-level code expects more advanced or predictable behavior from the library.

benjajaja:
@alexs-sh I *think* it's straightforward: just return `Err` in the `size()` method if rows and cols are zero. That would reinstate previous behaviour, and yet `window_size()` would be consistent in returning `Ok(actual reported values)`.

Would also be worth looking into what `last_os_error()` returns when rows and cols are zero - I bet it's not really related.

PR #1017, opened 2025-11-17
Refactor `KeyCode` & `ModifierKeyCode` to new modules

gavincrawford:
Related to #998.
This PR migrates `KeyCode` and `ModifierKeyCode` to new modules, which reduces the LoC in `event.rs` and makes the code a bit more friendly.


joshka (commented):
Thanks - this has been on my someday list to do as I prefer smaller files that reduce the cognitive load too.

Can you please move the impl blocks and unit tests too?
Also feel free to do Mouse in this same PR if you'd like - it needs to happen in the same sometime bucket.

gavincrawford:
I moved the mouse event, mouse event kind, and mouse button structures to their own file, and moved the applicable tests for key codes and modifiers to their respective modules. Are there any other `impl`s I'm missing?

PR #1024, opened 2026-01-13
fix: make cursor `read_position_raw` skip older cursor pos events

playwmadness:
This PR is addressing issues https://github.com/nushell/nushell/issues/17179 and https://github.com/nushell/reedline/issues/860.

If cursor position was requested from the terminal by something other than crossterm itself (for example via `printf "\e[6n"` in nushell), then `cursor::position()` will be experiencing a permanent lag in cursor positions due to `read_position_raw` requesting a new pos, but consuming only an old event.

To fix this, I made `read_position_raw` discard all pending cursor pos events before requesting a new one.

### DIFF
--- a/src/cursor/sys/unix.rs
+++ b/src/cursor/sys/unix.rs
@@ -33,6 +33,9 @@ fn read_position() -> io::Result<(u16, u16)> {
 }
 
 fn read_position_raw() -> io::Result<(u16, u16)> {
+    // Discard any buffered cursor-position replies from earlier `ESC[6n` requests so the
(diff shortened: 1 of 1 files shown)


emmabastas:
My gut feeling is that race-conditions are inherent to the way the shell and the terminal emulator communicate.
Maybe best we can hope to achieve is to document under which conditions things break down.

Do we know anything about how other shells /  terminal manipulation libraries do it?

noverby:
Can this be merged as is, or should the `TERMINAL_MODE_PRIOR_RAW_MODE` guard be included?

PR #1026, opened 2026-01-23
fix: open /dev/tty in write mode to query keyboard enhancement flags

sandydoo:
The keyboard enhancement flags query would always fail (and fall back to querying stdout) because `/dev/tty` was being opened as read-only.

### DIFF
--- a/src/terminal/sys/unix.rs
+++ b/src/terminal/sys/unix.rs
@@ -230,10 +230,13 @@ fn query_keyboard_enhancement_flags_raw() -> io::Result<Option<KeyboardEnhanceme
     // ESC [ c          Query primary device attributes.
     const QUERY: &[u8] = b"\x1B[?u\x1B[c";
 
-    let result = File::open("/dev/tty").and_then(|mut file| {
-        file.write_all(QUERY)?;
-        file.flush()
-    });
+    let result = File::options()
+        .write(true)
+        .open("/dev/tty")
+        .and_then(|mut file| {
+            file.write_all(QUERY)?;
+            file.flush()
+        });
     if result.is_err() {
         let mut stdout = io::stdout();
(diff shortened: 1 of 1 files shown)


src/terminal/sys/unix.rs
```diff
@@ -230,10 +231,13 @@ fn query_keyboard_enhancement_flags_raw() -> io::Result<Option<KeyboardEnhanceme
     // ESC [ c          Query primary device attributes.
     const QUERY: &[u8] = b"\x1B[?u\x1B[c";
 
-    let result = File::open("/dev/tty").and_then(|mut file| {
-        file.write_all(QUERY)?;
-        file.flush()
-    });
+    let result = OpenOptions::new()
```
joshka:
```suggestion
    let result = File.options()
```

src/terminal/sys/unix.rs
```diff
@@ -216,6 +216,7 @@ fn query_keyboard_enhancement_flags_raw() -> io::Result<Option<KeyboardEnhanceme
         filter::{KeyboardEnhancementFlagsFilter, PrimaryDeviceAttributesFilter},
         internal::{self, InternalEvent},
     };
+    use std::fs::OpenOptions;
```
joshka:
```suggestion
```

joshka (approved):
nit: Do you mind changing this to use
https://doc.rust-lang.org/std/fs/struct.File.html#method.options instead of OpenOptions?

sandydoo:
> nit: Do you mind changing this to use [doc.rust-lang.org/std/fs/struct.File.html#method.options](https://doc.rust-lang.org/std/fs/struct.File.html#method.options) instead of OpenOptions?

Done!

joshka:
Thanks for the fix

PR #1027, opened 2026-01-25
fix: const fn for ContentStyle and StyledContent

### LINKED ISSUE #760: `const` constructor for (and usefulness of) `StyledContent<D>`
it's not currently possible to construct a `StyledContent<D>` in a `const` context.

there are several functions which are not `const` but could be marked `const`.
- `StyledContent::new`
- `StyledContent::content`
- `StyledContent::style`
- `ContentStyle::apply`

a reason not to do this is that it bumps the MSRV from `1.58.1` to `1.61.0`, which is technically a breaking change, but that might depend more on if a guarantee has previously been made about the MSRV.

N/A

**Additional context**
N/A

xtqqczze:
Closes https://github.com/crossterm-rs/crossterm/issues/760

feature(const_fn_trait_bound) is stabilised since rust 1.61

### DIFF
--- a/src/style/content_style.rs
+++ b/src/style/content_style.rs
@@ -18,16 +18,21 @@ pub struct ContentStyle {
 }
 
 impl ContentStyle {
-    /// Creates a `StyledContent` by applying the style to the given `val`.
+    /// Creates a [`StyledContent`] by applying the style to the given `val`.
     #[inline]
-    pub fn apply<D: Display>(self, val: D) -> StyledContent<D> {
+    pub const fn apply<D: Display>(self, val: D) -> StyledContent<D> {
         StyledContent::new(self, val)
     }
 
-    /// Creates a new `ContentStyle`.
+    /// Creates a new [`ContentStyle`].
     #[inline]
-    pub fn new() -> ContentStyle {
-        ContentStyle::default()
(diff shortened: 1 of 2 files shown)


src/style/attributes.rs
```diff
@@ -6,6 +6,14 @@ use crate::style::Attribute;
 #[derive(Debug, Default, Clone, Copy, PartialEq, Eq, Hash)]
 pub struct Attributes(u32);
 
+impl Attributes {
+    /// Creates a new [`Attributes`].
+    #[inline]
+    pub(crate) const fn new() -> Self {
```
xtqqczze: Any benefit to making this fully public?

TimonPost: it already is with none() function doing the same

TimonPost: i think this function is not needed here

TimonPost (approved):

xtqqczze:
This took a dependency on the [`const_mut_refs`](https://dev-doc.rust-lang.org/beta/unstable-book/language-features/const-mut-refs.html) feature from Rust 1.83. This is fine, but we could have reduced MSRV to 1.75 otherwise, see https://github.com/crossterm-rs/crossterm/pull/1102.

PR #1028, opened 2026-02-05
fix `execute!` double-expansion issue with e.g. RefCell

Qix-:
This tripped me up for a good 30 minutes tracking down where my `RefCell` was panicking. I'm probably using Crossterm in a somewhat niche way, but the panic was not making sense to me and I finally realized `execute!` was probably using the first argument's expression twice in a single statement, which would cause the `borrow[_mut]()` to be called twice. Sure enough, that was the case.

This should have no functional differences aside from releasing any locks that the first `queue!` took out on the expression before taking out another lock (if the expression has one) for the flush. I'm not sure if this is more on the "breaking" or the "fixing" side of the spectrum, but this at least lets lock-like expressions behave more like one would expect.

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -113,16 +113,36 @@ macro_rules! queue {
 ///   and can therefore not be written to the given `writer`.
 ///   Therefore, there is no difference between [execute](macro.execute.html)
 ///   and [queue](macro.queue.html) for those old Windows versions.
+///
+/// # Side Effects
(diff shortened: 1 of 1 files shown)


Qix-:
I also added a small comment. While this PR does fix trivial side effects (borrows/locks etc) it's probably worth mentioning that it might incur side effects twice.

Really, highly doubt anyone would ever run into issues with this, but more docs can't hurt I don't think. Let me know if you find it unnecessary and I can remove it!

PR #1030, opened 2026-02-06
feat: add ANSI parsing support on Windows (VT input hybrid)

### LINKED ISSUE #737: Bracketed paste doesn't work on windows
as the title says, trying to use bracketed paste on windows , just yields a series of Event::char

This works on termwiz though, so its probably something missing we can copy over

### LINKED ISSUE #962: Bracketed paste interference on Windows
**Describe the bug**
Enabling bracketed paste on alacritty on windows (even though bracketed paste doesn't work on windows #737) causes interference with the pasting of `\n` characters by adding the control modifier to the Enter key event.

Not using bracketed paste makes pasting work as expected with `\n` characters showing up as an Enter key press with no modifiers.

Could this just be a weird quirk of how terminals work, or is this a bug? Not entirely sure.

**To Reproduce**
1. Modify the `event-poll-read` example included with crossterm to enable bracketed paste
2. Run the modified `event-poll-read` example on alacritty on windows
3. Using the terminal emulator's paste shortcut, paste in some text that contains a newline in it
4. Observe that the newline has a CONTROL modifier

**Expected behavior**
The key event should be free from the CONTROL modifier, as it is on other terminals, even if bracketed paste is enabled but doesn't work.

**Screenshots**

In each example I pasted the string `a\nb` into the terminal using `ctrl+shift+v`

event-poll-read with bracketed paste enabled:
image

Notice the `CONTROL` on the `Enter` (unexpected behaviour, should be the same as if bracketed paste was not enabled)

normal event-poll-read with no bracketed paste:
image

No `CONTROL` on the `Enter` key (expected behaviour)

**OS**
 - e.g. Windows 11

**Terminal/Console**
 - e.g. alacritty

eitsupi:
Fix #737
Fix #962

## Summary

- Keep the existing Win32 input path as the default on Windows. Raw mode preserves `ENABLE_VIRTUAL_TERMINAL_INPUT` instead of enabling it implicitly.
- Enable VT input only for the lifetime of `EnableBracketedPaste`, allowing modern Windows terminals to deliver bracketed paste without changing keyboard semantics for every raw-mode user.
- Preserve ANSI-only bracketed paste in environments such as Git Bash and mintty, where `TERM` indicates ANSI support but no Win32 `CONIN$` console is available.
- Share the ANSI parser between Unix and Windows by moving it from `sys/unix/parse` to `sys/parse`.
- Keep `WindowsEventSource` construction side-effect free and select the VT or VK path from the actual console mode for each input batch.
- Bridge Windows mouse capture through both WinAPI mode flags and the VT `?1003;1006` representation, so mouse input continues across VT transport changes.
- Fall back to the existing `ReadConsoleInput` VK handling when VT input is inactive or unavailable.

## Scope on Windows

The source reads the console mode once per batch. Character-bearing key-down records use the shared ANSI parser only while VT input is active; other records continue through the Win32 path when the host supplies them.

| Area | Bracketed paste disabled | Bracketed paste enabled / VT input active |
|---|---|---|
| Keyboard | Existing Win32 translation, preserving Press and Release; a held key may be exposed as additional Press records | Character and ANSI-sequence presses use the shared parser. Host-supplied releases remain supported, but Windows Terminal/ConPTY may expose a Press-only VT transport |
| Bracketed paste (`Event::Paste`) | Not enabled | Decoded by the shared ANSI parser |
| Alt+numpad | Existing Win32 handling | Preserves host-supplied releases and suppresses only the adjacent duplicate character record observed on conhost |
| Mouse | Win32 `MOUSE_EVENT` path | SGR VT reports through the shared parser; WinAPI capture remains enabled for conhost and legacy fallback |
| Focus and resize | Existing `INPUT_RECORD` handling | Existing records remain supported when supplied by the host |
| UTF-16 surrogate pairs | Existing VK handling | Separate surrogate buffers for the VT and VK paths |

Keyboard input is still expected to be read in raw mode. Reading events without entering raw mode remains valid for resize, focus, and mouse events and does not change the console mode as a side effect.

**Explicitly out of scope: the kitty keyboard protocol.** `supports_keyboard_enhancement()` still returns `Ok(false)` on Windows and `PushKeyboardEnhancementFlags` still reports `ErrorKind::Unsupported`. Sharing the parser removes one decoding obstacle, but negotiating that protocol on Windows is separate work.

## Console-mode lifecycle

- `WindowsEventSource::new()` opens `CONIN$` but does not modify its mode.
- `enable_raw_mode()` captures the first-touch console mode, changes only the Windows raw-mode bits, and preserves the current VT-input bit.
- `disable_raw_mode()` restores the cooked-mode bits while likewise preserving the current VT-input bit.
- On ANSI-capable Windows terminals with a Win32 console, `EnableBracketedPaste` flushes prior output, saves the current VT-input bit in a bracketed-paste-specific state, enables VT input, writes `CSI ? 2004 h`, and flushes again before returning.
- `DisableBracketedPaste` writes `CSI ? 2004 l`, flushes it, and only then restores the VT-input bit observed immediately before the paired enable. It preserves every other current console-mode bit and consumes the saved state only after a successful restore.
- An unpaired disable leaves the console mode unchanged. If the paired enable found no `CONIN$` console, both commands still emit their ANSI sequences and safely skip the WinAPI mode side effects.
- Missing-console errors are recognized narrowly; access-denied and genuine console-mode failures still abort before an unsafe state transition.
- The paired bracketed-paste commands are synchronization boundaries on Windows and are not reference-counted; independent nested ownership is not supported.
- The event source does not cache VT support for its lifetime. `Console` and `ConsoleMode` share the same `CONIN$` handle, and the real VT and raw-mode state is read once per batch.

The ordering is deliberate: VT input must be active before enabling the terminal protocol, while the disable sequence must be flushed before restoring the input transport. Failures short-circuit later steps so the parser transport is not disabled before the terminal has accepted `CSI ? 2004 l`.

## Command execution on Windows

The hidden `Command` implementation hooks distinguish:

- the historical WinAPI fallback used when ANSI is unsupported;
- WinAPI side effects that run before ANSI;
- WinAPI side effects that run before ANSI and force a final flush;
- WinAPI side effects that run after ANSI has been flushed.

This preserves the existing legacy fallback while supporting the coordinated mouse and bracketed-paste operations. Hybrid commands are rejected by the internal `fmt::Write` formatting path because it cannot provide the required I/O flush ordering.

## Mouse transport

Windows mouse capture now keeps the WinAPI flags and, when ANSI is supported, also emits `?1003;1006h/l`.

- With VT input off, terminals continue to deliver Win32 `MOUSE_EVENT` records.
- With VT input on for bracketed paste, Windows Terminal/ConPTY can deliver SGR mouse reports through the shared parser.
- The two representations preserve conhost/legacy fallback and avoid the mouse regression described by microsoft/terminal#15296. The VT representation matches the bridge used by microsoft/terminal#9970.

## Key releases and Alt+numpad

When a host supplies character-bearing key-up records during VT input, releases are normalized through the same control-character and uppercase rules as the shared ANSI parser. This keeps the public Press and Release identities aligned.

Some hosts do not supply those records. In particular, Windows Terminal/ConPTY was observed to expose ordinary characters, function keys, modified characters, and Alt-code input as Press-only while VT input is active. This PR does not synthesize missing releases because doing so would duplicate real releases on other hosts and would invent hold, repeat, and ordering semantics.

For the conhost Alt+numpad trace, suppression is limited to the exact adjacent sequence: an Alt-bearing `VK_PACKET` release followed immediately by the same character as a key-down record with no virtual-key code. Mismatches, non-key records, and VT mode changes clear the candidate.

## Input batching and ordering

- Available records are fetched with `read_console_input()` and only the records actually returned are processed. This removes the additional per-record blocking race introduced by repeatedly calling `read_single_input_event()`. The pre-existing race where another console reader drains the entire queue between readiness and `ReadConsoleInputW` remains.
- Incomplete ANSI sequences survive input-batch boundaries.
- The parser records the queue position at which an incomplete sequence began. A later `flush()` or completion inserts the parsed event at that position, preserving its order relative to mouse, focus, resize, and VK-translated events.
- ANSI responses with no Windows consumer, such as cursor-position and keyboard-enhancement reports, are discarded at the Windows source boundary instead of remaining permanently in the shared reader queue.

## Esc and Alt sequence boundary

With VT input enabled for bracketed paste, Windows delivers keyboard input as the same byte stream used by Unix terminals. This carries the same inherent ambiguity between a standalone `Esc` key and the prefix of an Alt or CSI sequence.

The parser uses the currently available input batch: a quickly followed key can combine with `Esc` as an Alt-prefixed sequence, while a prefix split at a queue-empty boundary can be emitted as a standalone `Esc`. This PR intentionally keeps that Unix-compatible behavior rather than adding a Windows-only timing threshold.

Once the complete bracketed-paste opener (`ESC [ 200 ~`) has been recognized, temporary queue drains do not split the paste payload. A split at the opening `Esc` remains subject to the same boundary.

## Diagnostic example

`windows-vt-input` provides two phases in one run:

1. Raw mode with bracketed paste disabled, for ordinary/F-key/modified-key Press and Release behavior plus Win32 mouse input.
2. Bracketed paste enabled, for multiline `Event::Paste`, SGR mouse, physical-numpad Alt codes, and the documented Press-only host behavior.

It drains input already queued when the event source is constructed (typically the command-launch Enter release) before Phase 1 statistics begin. Its key accounting distinguishes additional Press records emitted while a key is held from genuinely unmatched identities, while retaining signed detection for release-first, extra-release, modifier-mismatch, and Press-only cases. It also reports narrowly matched Alt-code duplicate candidates, mouse activity without flooding moved events, mode transitions, partial-setup cleanup, and restoration of VT, mouse, window, and extended console-input flags. When VT input was already active before startup, Phase 1 advances on the transition-key press because ConPTY may not provide its release; otherwise it observes the press and release before enabling paste.

## Follow-up work

An unterminated bracketed paste can keep the shared parser waiting indefinitely and grow its buffer without a bound. This behavior already exists on Unix and is now reachable on Windows. Choosing a buffer limit and recovery behavior affects both platforms and is deferred to a separate parser-hardening change.

## Tested

- `cargo test --locked --all-targets --all-features -- --test-threads 1`: 153 library tests passed, 0 failed, 7 ignored, plus integration targets and 4 diagnostic-accounting example tests
- `cargo clippy --locked --all-targets --all-features -- -D warnings`
- `cargo check --locked --all-targets --no-default-features` (one pre-existing `file_descriptor::read` dead-code warning)
- `cargo test --locked --all-targets --no-default-features`: 52 passed, 0 failed, 1 ignored, plus integration and example targets
- The lifecycle-scoped implementation at `245d5c6` was manually validated on both Windows Terminal and modern conhost: raw mode preserved VT input off; Phase 1 retained Win32 Press/Release and mouse behavior; Phase 2 delivered multiline `Event::Paste`, SGR click/drag/wheel input, and one Alt+numpad character Press without a duplicate; cleanup restored the original VT, mouse, window, and extended input bits.
- The latest `94ed715` changes only diagnostic startup-event and held-key accounting based on those traces. Its four example tests cover the balance invariants, and a modern-conhost run confirmed that the startup Enter release is separated, additional held-key Press records do not inflate imbalance, and a Press-only VT identity still remains visible as one mismatch.


ofek:
Thanks for working on this! Could you please rebase to resolve the conflicts so that the maintainers can review?

eitsupi:
@ofek While it's certainly possible to resolve the conflict, I don't know what that would mean for a review.
This is because this PR had been left untouched for months even when there were no conflicts.

I don't think we can definitively say that resolving the conflict guarantees a review.

joshka:
I'm not generally using Windows these days (mostly macOS) and crossterm stuff is in my third tier of where I generally focus my attention on getting over the line (I volunteered to help out Timon a while back). I'd recommend having another contributor or two do a bit of an adversarial review here if they can. Once that's here I'll dive back and take a bit more of a look.

I'd encourage everyone to feel free to review code like this. It helps to have extra validation from people actually impacted by changes as the maintainers don't generally have a perfect coverage on every individual feature. Similar to the idea in that bevy uses https://shaping.systems/blog/open-code-review/ (there's a 20 min talk if that's more your style https://www.youtube.com/watch?v=6uRyFMM8vDI)

eitsupi:
To clarify, if you want to incorporate this branch, you don't need to specify the URL of my fork. Adding something like the following to your Cargo.toml should work.

```toml
[patch.crates-io]
crossterm = { git = "https://github.com/crossterm-rs/crossterm", rev = "03bb4f26610a5904b8e875fcb610d16bf7d00814" }
```

See my project, which motivated this PR, for a real-world example.
https://github.com/eitsupi/arf/blob/4b4de8b8d1333f394ae0c9a5d4197924f3dd4bc3/Cargo.toml#L95-L98

easyinplay:
Coming over from nushell/reedline#1132, where @eitsupi pointed me here. I ran this branch across the Windows terminals I have, against a control build, since a report with no baseline can't attribute anything to the PR.

## Method

Same diagnostic example built twice from the same source: once on `pr-1030` (this branch, merged with master at the time of testing), once on plain `origin/master` (`34636cf`). It enables raw mode and `EnableBracketedPaste`, then appends one JSON line per event with the payload's byte/char/line counts and a SHA-256 of the content normalized to LF, so the delivered text can be compared against what was put on the clipboard. Same machine, same shell, same payload. Windows 11 26200.

## Result: the PR is what makes bracketed paste work

| terminal | `origin/master` | this branch |
|---|---|---|
| Windows Terminal Canary 1.26.2182.0 | 659 `Key`, **0 `Paste`** | **`Paste`** |
| Warp v0.2026.08.05 | 657 `Key`, **0 `Paste`** | **`Paste`** |
| conhost (modern) | 337 `Key`, **0 `Paste`** | **`Paste`** |

On master every paste arrives as a stream of individual key events — inter-event gaps of 150–400 µs, i.e. the terminal's bracketed-paste markers are being delivered but nothing decodes them. On this branch the same paste in the same terminal produces a single `Event::Paste`. The log files differ by two orders of magnitude in size (97 KB vs 1.6 KB) for the same pasted text.

Terminals where this branch delivers `Event::Paste` (branch only, no control run):

- WezTerm nightly `20260812-070121-fe3006ae`
- Alacritty 0.17.0
- OxideTerm 2.0.19 — with a caveat, below

## Content integrity

Payload: three ASCII lines with a trailing newline, 47 chars, placed on the clipboard programmatically so the expected hash is known before the paste happens.

Windows Terminal, conhost and OxideTerm all delivered 47 chars with a normalized SHA-256 exactly matching the source (`c03e4a3e…`). No truncation, no doubling, no loss.

## Newlines arrive as CR, not LF

In all three, the line separators came through as lone `\r` (`lines: 0`, `crlf: 0`, content `case2 line one\rcase2 line two\r…`). The clipboard held LF.

That is the terminals' doing, not this PR's — `parse_csi_bracketed_paste` hands back `String::from_utf8_lossy(&buffer[6..buffer.len() - 6])`, the bytes verbatim, and translating a newline to CR is what a terminal does with a line break. Worth writing down anyway, because a consumer that checks `Event::Paste` content for `'\n'` will find none, and one that treats a lone `\r` as a submit will do the wrong thing on a multi-line paste. Consistent across three different terminals, so it looks like the norm rather than a quirk.

## OxideTerm behaves two ways

Its first run produced 394 individual `Key` events and no `Paste` at all — a 198-char paste arriving as a keystroke burst even on this branch. A later run in the same terminal produced one clean `Paste`. Both are real; the difference is most likely the paste gesture (its own `Ctrl+V` handling versus a gesture that emits bracketed paste), but I did not record which gesture was used for each, so that part is a hypothesis and not a result.

Either way it is a current, maintained Windows terminal where an application can still receive a paste as synthetic keystrokes, which is worth knowing for anyone assuming this PR removes that case entirely.

## Not covered

- Legacy console (`ForceV2=0`) — the fallback path in your scope table is untested here.
- The `disable_mouse_capture()` regression you describe. I built a toggle for it but never exercised it, so I can't confirm or deny that fix.
- Large pastes (I have a 5000-line payload wired up but did not run it) and pathological content (tabs, lone CR, an 8000-char unbroken line).
- CRLF on the clipboard, which is what a real Windows copy produces; everything above was LF.
- WaveTerm.

Happy to fill any of those in if it would help move the PR — say which matter to you and I'll run them.

easyinplay (commented):
@joshka asked for an adversarial review before looking at this again, so here is one. For context on where I'm standing: I maintain a Windows-first agent CLI on reedline, and I opened nushell/reedline#1132 to work around the exact bug this PR fixes. I want this merged. That is the reason I went looking for problems in it rather than just confirming it works.

Two of the findings below are measured. Reproduction notes at the end.

---

## 1. Character-key release events are discarded, and that breaks a documented guarantee

`src/event/source/windows.rs:85-124`

```rust
if self.vt_input_enabled && record.u_char != 0 && record.key_down {
    // VT path
} else if !self.vt_input_enabled || record.u_char == 0 {
    // VK path
}
// VT enabled, key_down=false, u_char!=0: intentionally
// skipped. Release events don't carry new ANSI data, and
// crossterm only reports key-press events.
```

A record with `u_char != 0` and `key_down == false` matches neither branch, so it is dropped. The comment gives two reasons and I don't think either survives.

"crossterm only reports key-press events" is not true on Windows. `src/event.rs:929-931` documents `KeyEvent::kind` as *"Only set if: … Windows: always"*, and `src/event/sys/windows/parse.rs:277-283`, which is the code this branch is bypassing, sets `Press`/`Release` from `key_down` for every key that maps to a `KeyCode`.

The second effect is subtler. Keys with `u_char == 0` (arrows, F-keys, bare modifiers) fall through to the VK path at `:111` and keep releasing normally. Character keys don't. So whether a key emits a release becomes a property of which key it is.

### Is the discarded record actually delivered?

This is the part I expected to disprove. If VT translation swallowed key-up records on its own, the branch would be dead code and the finding would evaporate. So I wrote a probe with no crossterm in it: allocate a private console, open `CONIN$`, set or clear `ENABLE_VIRTUAL_TERMINAL_INPUT`, inject records with `WriteConsoleInputW`, read them back with `PeekConsoleInputW`.

VT bit set, 6 records in, 8 out:

| injected | delivered |
|---|---|
| `'a'` press, `u_char=0061 vk=41` | `key_down=true u_char=0061`, **`vk=00 scan=00`** (rewritten as a VT byte) |
| `'a'` **release**, `u_char=0061 vk=41` | **`key_down=false u_char=0061 vk=41 scan=1E`**, verbatim |
| `F1` press, `u_char=0000` | three records, `001B 004F 0050` (`ESC O P`), all `key_down=true` |
| `F1` release, `u_char=0000` | verbatim |
| `VK_MENU` **release**, `u_char=00E9` | **`key_down=false u_char=00E9 vk=12`**, verbatim |

VT bit cleared: 6 in, 6 out, every field verbatim.

conhost's VT translation only consumes key-**down** records. Key-ups go through it untouched and keep their `u_char`. The shape this branch throws away is delivered in normal operation.

### And from a real keyboard

conhost, `Ctrl`+`V`, same terminal, same logging probe built from the same source against each tree:

| build | `Char('v')` + `CONTROL` |
|---|---|
| `origin/master` | 1 Press / **1 Release** |
| this branch | 4 Press / **0 Release** |

F-keys in that same branch session: **9 Press / 9 Release**. Real key-ups do reach crossterm under VT input. Only the character-bearing ones go missing.

What I have not done is watch a physically typed *character* key-up under VT. The injection shows the record survives translation, and the conhost session shows real key-ups arrive, but the conjunction of the two is inference, not observation. Worth saying plainly. Also worth saying: of the four terminals I tested, only conhost delivers key-up semantics at all, so that press/release contrast is only visible there.

The practical consequence is anything that tracks key-hold state. A hold-to-pan binding sees `Press` for `a` and never the matching `Release`, so the key is stuck down forever, while `Left` behaves correctly in the same app.

### It also removes Windows Alt codes

`src/event/sys/windows/parse.rs:200-204`:

```rust
// We normally ignore all key release events, but we will make an exception for an Alt key
// release if it carries a u_char value, as this indicates an Alt code.
let is_alt_code = virtual_key_code == VK_MENU && !key_event.key_down && key_event.u_char != 0;
```

`!key_down && u_char != 0` is exactly the discarded shape, and the probe row above shows the console does deliver it. At `source/windows.rs:85` the VT arm needs `key_down`; at `:111` the VK arm needs `u_char == 0`. Neither matches, `handle_key_event` is never called, and `is_alt_code` never gets a chance to fire. With VT input active, <kbd>Alt</kbd>+numpad `0233` loses the `é`.

For the same reason both `WindowsKeyEvent::Surrogate` producers become unreachable while VT is on, at `parse.rs:207` (the Alt-code branch) and `parse.rs:264` (the general `u_char` branch), since a surrogate code unit is by definition `u_char != 0`. Which is also why `legacy_surrogate` is provably always `None` in VT mode. The separate-surrogate-buffers rationale in the PR description is therefore not quite right, though the code itself is harmless.

This is a regression from master, and it contradicts the description's claim that discarding these "preserves the legacy Windows path's press-only behavior."

---

## 2. Building the event source mutates global console state, with nothing to undo it

`src/event/source/windows.rs:41`

```rust
let vt_input_enabled = try_enable_vt_input()?;
```

`WindowsEventSource` is constructed once per process from the global `EVENT_READER` (`src/event/internal.rs:10-16`). It has no `Drop`. The only code anywhere that clears `ENABLE_VIRTUAL_TERMINAL_INPUT` is `compute_disable_raw_mode` (`src/terminal/sys/windows.rs:57-64`), which is only reachable through `disable_raw_mode()`.

Reading events without raw mode is legal. `src/event.rs:18` says raw mode is needed *"for keyboard events to work properly"*, and resize/focus/mouse don't need it at all. A program that does that now permanently changes the console it was launched from.

Measured in one console, with the VT bit cleared first so the console's true original genuinely lacks it:

```
                          this branch      origin/master
start                     0x0008 clear     0x0008 clear
one event::poll, no raw   0x0208 SET       0x0008 clear
after the process exits   0x0208 SET       0x0008 clear
```

Console input mode belongs to the shared console input buffer rather than the process, so the parent shell and any later child inherit the change.

There's also no clean way for such a program to put it back. `compute_disable_raw_mode` starts from `current | NOT_RAW_MODE_MASK`, so calling `disable_raw_mode()` purely to clear the VT bit would also force-set `ENABLE_LINE_INPUT | ENABLE_ECHO_INPUT | ENABLE_PROCESSED_INPUT`, which for a never-raw program is just wrong.

New in this PR. On master, constructing the event source touched no console mode at all.

---

## 3. `vt_input_enabled` caches a bit that `disable_raw_mode()` clears

`src/event/source/windows.rs:35`, `:41`, `:85`

The flag is computed once at construction and never re-checked, while `compute_disable_raw_mode` clears the bit at the console:

```rust
enable_raw_mode()?;   // original snapshot not taken yet -> VT not set here
event::read()?;       // source constructed: VT on, vt_input_enabled = true
disable_raw_mode()?;  // VT cleared at the console
event::read()?;       // vt_input_enabled is still true
```

In that last read the source applies VT decoding to records that are no longer VT-translated.

I want to be careful about where the fault is here, because it is easy to read this as a misunderstanding of VT. Under genuine VT operation the console itself collapses `Ctrl`+`Shift`+`A` into a single `u_char=0001` record with `control_key_state = 0`. I checked that with the probe. Decoding that as `CONTROL` alone is correct and matches Unix. The problem is only the stale flag: once the console is back on the VK path it emits `u_char=0001 vk=41 control_key_state=0x18`, with `SHIFT|LEFT_CTRL` intact, and the VT path never looks at `control_key_state`. So `Ctrl`+`Shift`+`A` comes out as `Key(Char('a'), CONTROL)` where master gives `CONTROL | SHIFT`. Every character key-release is dropped in that window too, per finding 1.

Re-entering raw mode closes the window, so the suspend-for-`$EDITOR` pattern heals itself. Reading events while deliberately in cooked mode does not.

One more thing in the same area. The doc comment at `src/event/sys/windows.rs:23-27` says *"Currently, `try_enable_vt_input` is called first in `WindowsEventSource::new()`, which is correct."* For the standard startup order it isn't: `enable_raw_mode()` runs first, modifies the mode, and never calls `init_original_console_mode`, so what gets stored as "original" is `true_original & !NOT_RAW_MODE_MASK`. It happens to be harmless right now, because the only two consumers read the VT bit and the three `ENABLE_MOUSE_MODE` bits and neither group overlaps `NOT_RAW_MODE_MASK`. But the invariant the comment asserts isn't true, and whoever next reads a different bit out of that static will get a wrong answer with no indication.

---

## 4. Non-`Event` internal events are reachable on Windows now, and nothing consumes them

`src/event/source/windows.rs:108` feeds VT bytes into the shared parser, which can emit `InternalEvent::CursorPosition`, `KeyboardEnhancementFlags` and `PrimaryDeviceAttributes`. The same PR drops the Windows blanket in `EventFilter`:

```rust
-    #[cfg(windows)]
-    fn eval(&self, _: &InternalEvent) -> bool {
-        true
-    }
```

so on Windows those variants now fail the filter. `src/event/read.rs:125-137` pushes non-matching events straight back into the queue and there is no discard path anywhere, while `read.rs:48-52` linear-scans that queue on every `poll`.

I went looking for a Windows consumer and there isn't one. The three filters are constructed only in `src/terminal/sys/unix.rs:249-255` and `src/cursor/sys/unix.rs:39-52`, plus tests. `cursor::position()` on Windows goes through `ScreenBuffer` (`src/cursor/sys/windows.rs:38-45`). `supports_keyboard_enhancement()` returns `Ok(false)` (`src/terminal/sys/windows.rs:122-124`). This PR marking those filter structs `#[cfg_attr(windows, allow(dead_code))]` is the same fact from the other side.

So an app or a dependency that writes its own `ESC[6n`, which is a common size/position fallback, leaves one permanently stuck entry per reply. Memory grows, every poll gets a little slower, and nothing recovers it.

Two limits on that claim, since I'd rather state them than have them found: crossterm never sends those queries on Windows itself, so this needs an application-side probe or an unsolicited report from the terminal, and I have not confirmed that conhost forwards a DSR-CPR reply to the client instead of intercepting it. The mechanism is certain, the trigger is conditional.

---

## 5. The lone-ESC heuristic has no timing guard, and it fails both ways

`src/event/source/windows.rs:100-107`

```rust
let more_input_available = if remaining > 0 {
    true
} else {
    self.console.number_of_console_input_events()? > 0
};
```

Whether the queue happens to be non-empty right now is the only thing separating "the user pressed Esc" from "a CSI sequence starts here". No delay, no look at what follows.

Both mis-readings are reachable. Press <kbd>Esc</kbd> and then <kbd>a</kbd> close enough together that both records land in one snapshot, and the Esc disappears. I ran the parser directly to check: `parse_event(b"\x1ba", false)` returns `Key(Char('a'), ALT)`, and feeding it byte-at-a-time the way this source does, `advance(b"\x1b", more=true)` then `advance(b"a", more=false)`, yields exactly one event, `Key(Char('a'), ALT)`. Master's VK path always produced `KeyCode::Esc` for `VK_ESCAPE` and never merged it with the next key.

The mirror case is just as reachable. A genuine <kbd>Alt</kbd>+<kbd>A</kbd> whose `\x1b` happens to end a batch with the queue momentarily empty gets `more_input_available == false`, so `KeyCode::Esc` goes out immediately and the letter arrives as a bare character in the next batch. `Alt+A` degrades to `Esc`, `a`.

Matching Unix here is a defensible trade. But it is a behavior change on Windows in both directions and the PR description doesn't mention it.

Bracketed paste is safe from this, for what it's worth. Once the buffer starts with `\x1b[200~`, `parse_csi_bracketed_paste` returns `Ok(None)` regardless of `input_available` and `flush()` preserves the buffer on `Ok(None)`, so a mid-paste queue drain can't split a payload. Only a split at the opening byte of a sequence is dangerous.

---

## 6. `flush()` puts the ESC ahead of events that arrived before it

`src/event/source/windows.rs:164-166`, with `src/event/sys/parse.rs:1720-1735` (`self.internal_events.push_front(ie)`).

Prepending is right only when the ESC was the first event-producing record in the batch. I exercised the other case:

```
push_event(Mouse col=1); advance(b"\x1b", true); push_event(Mouse col=2); flush();
delivery order = [Key(Esc), Mouse(col=1), Mouse(col=2)]
```

The Esc arrives before a mouse event that physically preceded it, which matters with mouse capture on and a drag in progress.

The blast radius is small and worth writing down in the PR so the next reviewer doesn't have to derive it: `input_available` is consulted in exactly one place in the parser (`parse.rs:36-43`, the `buffer.len() == 1 && buffer[0] == 0x1B` case), so `flush()` can only ever emit a `KeyCode::Esc`.

---

## 7. `unwrap_or(false)` does the opposite of what its comment says

`src/event/sys/parse.rs:100-104`

```rust
// unwrap_or(false): if the console mode query fails (possible on Windows),
// assume raw mode is enabled so \n is treated as Ctrl+J rather than Enter.
b'\n' if !crate::terminal::is_raw_mode_enabled().unwrap_or(false) => Ok(Some(
    InternalEvent::Event(Event::Key(KeyCode::Enter.into())),
)),
```

Query fails, `unwrap_or(false)` gives `false`, `!false` is true, the guard fires, and you get `KeyCode::Enter`. That's "assume raw mode is not enabled", which is the opposite of the comment. `.unwrap_or(true)` is what the comment describes. Same thing at `src/event/sys/parse.rs:554-557`.

On master this line was `crate::terminal::sys::is_raw_mode_enabled()`, infallible on Unix, so the fallible path is new here. On Windows with raw mode on but the mode query failing, `0x0A` decodes as a bare `Enter` instead of `Ctrl+J`, which is the case issue #371's comment exists to prevent.

---

## Smaller things

The batch read can block past the caller's timeout. `src/event/source/windows.rs:79-83` reads exactly `number` records after taking the snapshot, where master read one, and `crossterm_winapi`'s own doc says `ReadConsoleInputW` *"guarantees that at least one event is read, even if it means blocking the thread"*. I hit this by accident while building the probe above: `GetNumberOfConsoleInputEvents` reported 1, another process on the shared console took the record before my read, and the read blocked indefinitely. Master only lost that race when the queue was drained completely; this loses it whenever it loses any single record.

An unterminated bracketed paste stalls all input permanently. `Parser::buffer` is cleared only on `Ok(Some(..))` or `Err(..)`, and `parse_csi_bracketed_paste` keeps returning `Ok(None)` until the terminator arrives. If it never arrives, because the terminal died mid-paste or the app sent `ESC[?2004l` while a paste was in flight, the buffer grows with every subsequent keystroke and no key event is ever delivered again. This exists on Unix too, but it's newly reachable on Windows and unbounded on both. A cap seems worth having, given that bracketed paste is what the PR is for.

`is_raw_mode_enabled()` costs a console handle open per `\n` byte on Windows. On Unix it reads a cached atomic. On Windows `src/terminal/sys/windows.rs:23-31` does `ConsoleMode::from(Handle::current_in_handle()?)`, and `current_in_handle()` is a fresh `CreateFileW("CONIN$")` with a `CloseHandle` on drop. The VT path consults it per newline byte, so a 20k-line paste is 20k open/query/close cycles rather than 20k `GetConsoleMode` calls.

The `#[allow(dead_code)]` at `src/event/sys/parse.rs:1703`, `:1719` and `:1756` is unconditional, while `src/event/filter.rs:10`, `:20` and `:37` uses `#[cfg_attr(windows, allow(dead_code))]`. Those three items are used on Windows and dead only on Unix, so `#[cfg_attr(unix, …)]` would keep the Windows-side warning alive. That matters a little more than usual because `flush()` exists to fix the lone-ESC stall, and losing its call site in some future refactor shouldn't be silent.

Very minor: `src/event/source/windows.rs:41` uses `?` on `try_enable_vt_input()`, whose own contract is "`Ok(false)` if unsupported". `.unwrap_or(false)` would match it.

---

## Things I checked that turned out fine

Writing these down so the next person doesn't spend the time again.

Unix parsing is unchanged. The rename diff of `sys/unix/parse.rs` to `sys/parse.rs` has exactly two non-test changes, both `crate::terminal::sys::is_raw_mode_enabled()` becoming `crate::terminal::is_raw_mode_enabled().unwrap_or(false)`. On Unix `terminal.rs:107-111` always returns `Ok`, so those are equivalent, and finding 7 is Windows-only. The `cfg` set inside the file is identical to master's.

The two duplicated `Parser` copies were identical on master. `git diff --no-index` between master's `mio.rs` and `tty.rs` shows zero difference across the `Parser` block, so there's no question of which copy survived. The shared version adds `push_event`/`flush` and `pub(crate)`, nothing else.

No `_ => {}` was introduced anywhere. The `#[cfg(unix)] _ => unreachable!()` arms became unconditional `_ => unreachable!()` with safety comments, and I confirmed they really are unreachable via `read.rs:98-140` plus `EventFilter::eval`.

No public API or semver change. `InternalEvent` is `pub(crate)`, and `Parser`, `decode_utf16_char` and `mod parse` are all `pub(crate)` under a `pub(crate)` `event::sys`.

Ordering of VT-parsed against VK-translated events within a batch is fine, since both `push_back` in record order. The only ordering hazard is finding 6.

Non-BMP input encodes correctly. `decode_utf16_char` plus `ch.encode_utf8(&mut [0u8; 4])` is right, and a high surrogate split across batches survives because `vt_surrogate` lives on the source.

Mouse can't double-fire via VT. `EnableMouseCapture` forces `is_ansi_code_supported()` to `false` on Windows, so crossterm never writes `ESC[?1000h` there and the terminal never sends SGR mouse reports back.

No quadratic behavior on large pastes: per-byte `parse_event` on a `\x1b[200~…` buffer is O(1).

`remaining` can't underflow, and the extra `number_of_console_input_events()` probe fires once per batch rather than once per character.

`Handle::current_in_handle()` opens `CONIN$` directly (`crossterm_winapi-0.9.1/src/handle.rs:143-155`), so redirected stdin doesn't affect event-source construction. I checked because it looked like a failure mode and it isn't.

---

## The feature does work, and this PR is what makes it work

Same probe built twice from the same source, once on this branch and once on `origin/master` (`34636cf`), same machine, same pasted payload:

| terminal | `origin/master` | this branch |
|---|---|---|
| Windows Terminal Canary 1.26.2182.0 | 659 `Key`, 0 `Paste` | `Paste` |
| Warp v0.2026.08.05 | 657 `Key`, 0 `Paste` | `Paste` |
| conhost | 337 `Key`, 0 `Paste` | `Paste` |

WezTerm nightly and Alacritty 0.17.0 also deliver `Event::Paste` on the branch, and the content is byte-exact: a normalized SHA-256 of the delivered text matches the clipboard source. More detail in my earlier comment.

---

## Reproduction

Three programs are behind the above, and I can clean any of them up and post them if that's useful:

- the raw `INPUT_RECORD` probe from finding 1, with no crossterm in it, which allocates a private console, toggles the VT bit, injects with `WriteConsoleInputW` and reads back with `PeekConsoleInputW`
- finding 2's probe, which reads the console input mode via `CONIN$`, clears `0x0200`, calls `event::poll(Duration::ZERO)` once without ever entering raw mode, and re-reads
- the key-event logger used for the conhost `Ctrl`+`V` comparison, which writes `{code, modifiers, kind}` per event to a file; build it in a checkout of this branch and in a worktree of `origin/master` and compare

The gap I couldn't close from here is a physically typed character key-up under VT input. If you want that nailed down, the raw probe with a "read the real keyboard for five seconds" arm, run in conhost, pressing `a`, then `F1`, then <kbd>Alt</kbd>+numpad `0233`, settles finding 1 and the Alt-code part of it outright.

## Not covered

I haven't tested the legacy console path (`ForceV2=0`) from your scope table, `disable_mouse_capture()` before and after the fix, large pastes, or a CRLF clipboard. Say the word on any of them and I'll run it.

eitsupi:
@easyinplay Thanks for your detailed review! I'll take a look.

eitsupi:
@easyinplay Thank you for the detailed review. I have pushed a set of changes addressing the findings at the current HEAD.

Findings 1–4, 6, and 7 are fixed in code. Finding 5 remains an intentional Unix-compatible Esc/Alt ambiguity and is now documented.
The batch-read regression is reduced to the pre-existing master race.
The unterminated bracketed-paste issue is deferred to a separate cross-platform parser follow-up.

I also added a Windows diagnostic example for the mode lifecycle, key-event kinds, and bracketed paste.
Could you please re-review the current HEAD?

easyinplay (commented):
@eitsupi Thanks for working through all of these. Six findings addressed in one push, finding 5 documented as an intentional trade rather than quietly kept, and a first-party diagnostic example on top. That last one is the part that keeps paying off: it gives the next person a way to check the mode lifecycle and key kinds without building their own probe. This PR has been waiting since February for reasons that have nothing to do with the work in it.

Re-reviewed `11e4912`. All six findings you addressed are fixed, and I confirmed each one against the code rather than taking the commit messages for it. Finding 5 reads fine as an intentional, now-documented trade.

Four new things below. One of them will fail CI and is a one-line fix, so it leads. Two of the others are consequences of the finding 1 fix rather than pre-existing problems.

I also have a correction to make to my own finding 1. The behavior I reported was real, but the mechanism I gave for the Alt-code half of it was wrong, and the raw records I captured for this pass show why.

---

## `cargo clippy -- -D warnings` fails on this HEAD

`.github/workflows/ci.yml:65`, verbatim:

```
cargo clippy --locked --all-targets --all-features -- -D warnings
```

```
error: this loop could be written as a `for` loop
    --> src/event/sys/parse.rs:1833:9
     |
1833 |         while let Some(event) = self.next() {
     |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: try: `for event in self.by_ref()`
     |
     = note: `-D clippy::while-let-on-iterator` implied by `-D warnings`

error: could not compile `crossterm` (lib) due to 1 previous error
```

That is `next_event()`, from `c12062b8`. Clean on `3ca5429` with the same toolchain, so it came with this push. The suggested `for event in self.by_ref()` builds on Windows and takes the command back to clean.

Leading with it because it is one line and the rest of this can wait for a normal round.

---

## Character key-up now works, and two things follow from how it works

Finding 1 is fixed. `source/windows.rs:103`'s `else` arm now takes every record the VT arm doesn't, and the comment says exactly what it's for. Measured in conhost on this HEAD:

```
Key Char('a') NONE Press
Key Char('a') NONE Release     +778 us
```

against 4 Press / 0 Release on the previous HEAD. Arrow and function keys are unchanged, as expected. Alt+numpad produces a character again.

Both of the following are new behavior introduced by that fix, not pre-existing. I don't think either is a reason to hold the PR, but the second one is a visible correctness bug and neither is mentioned in the description.

Raw `INPUT_RECORD`s below come from a read-only dump taken through the same `CONIN$` handle crossterm uses, after crossterm's own `enable_raw_mode()` has run. Nothing is injected and the dump changes no mode. conhost, `mode=0x000003f0`, `VT_INPUT=true`, Windows 11 26200. `cks` is `control_key_state` in hex.

### 2. Press and Release disagree for Ctrl+Shift+letter

Physically typed `Ctrl`+`Shift`+`A`:

```
down=true  vk=0x11 scan=0x1D u_char=0x0000 cks=0x28   ctrl down
down=true  vk=0x10 scan=0x2A u_char=0x0000 cks=0x38   shift down
down=true  vk=0x00 scan=0x00 u_char=0x0001 cks=0x00   VT byte, cks cleared
down=false vk=0x41 scan=0x1E u_char=0x0001 cks=0x38   key-up, verbatim
down=false vk=0x11 scan=0x1D u_char=0x0000 cks=0x30
down=false vk=0x10 scan=0x2A u_char=0x0000 cks=0x20
```

delivered as

```
Key Char('a') CONTROL         Press
Key Char('A') SHIFT | CONTROL Release
```

The key code and the modifiers both differ between the two halves of one keypress. The cause is structural rather than a slip in the routing: the VT byte is `0x01`, which has no way to encode shift, and conhost clears `control_key_state` on the record it rewrites. The key-up keeps `cks=0x38` (`NUMLOCK | SHIFT | LEFT_CTRL`) because VT translation passes key-ups through untouched. So the VT arm can only ever report `CONTROL` here while the VK arm reports `SHIFT | CONTROL`, and this will hold for every `Ctrl`+`Shift`+letter.

An application pairing press to release by `(code, modifiers)` gets a `Release` matching no `Press`, and the `Press` stays outstanding. That is the same stuck-key symptom finding 1 was about, reached a different way, so it seems worth deciding deliberately rather than inheriting.

This also closes the gap I was explicit about last time. I wrote then that I had not watched a physically typed character key-up under VT and that the conjunction was inference. It isn't any more; the trace above is a real keystroke.

### 3. Alt+numpad delivers the character twice

`Alt`+numpad `0233`:

```
down=true  vk=0x00 scan=0x00 u_char=0x001B cks=0x00   ESC, VT byte
down=true  vk=0x00 scan=0x00 u_char=0x00E9 cks=0x00   e-acute, VT byte
down=false vk=0xE7 scan=0x00 u_char=0x00E9 cks=0x22   key-up, vk=0xE7 (VK_PACKET)
down=true  vk=0x00 scan=0x00 u_char=0x00E9 cks=0x00   e-acute, VT byte again
down=false vk=0x12 scan=0x38 u_char=0x0000 cks=0x20   alt up, u_char is zero
```

delivered as

```
Key Char('é') ALT  Press
Key Char('é') ALT  Release
Key Char('é') NONE Press
```

Three events, and two of them are `Press`. A text field receiving this inserts the character twice.

Tracing them: the first two records are `ESC` followed by the character as VT bytes, which the shared parser combines into `Alt`+`é`. The `VK_PACKET` key-up is what the new `else` arm now routes to `handle_key_event`, giving the `Release` with `ALT` from `cks=0x22` (`NUMLOCK | LEFT_ALT_PRESSED`). The third record is the character arriving as a VT byte a second time, with no modifiers.

The doc comment on `examples/windows-vt-input.rs` gives the expected observation as "Alt+numpad 0233 producing a character event", singular, so I don't think this is intended.

### Correction to my finding 1

My finding 1 claimed that with VT input active, `Alt`+numpad `0233` loses the character because `is_alt_code` at `sys/windows/parse.rs:202` becomes unreachable. The loss was real on the previous HEAD, but that rationale was wrong, and the last record above is why: the `Alt` key-up carries `u_char=0x0000` under VT input. `is_alt_code` requires `u_char != 0`, so it cannot fire under VT regardless of how records are routed. It isn't being bypassed; it has nothing to act on.

What actually happens is that conhost supplies the character itself as VT bytes. On the previous HEAD those still arrived, but the discarded key-up meant nothing completed, and on this HEAD they arrive alongside a third delivery from the VK arm. I based the earlier claim on injected records rather than a typed Alt sequence, and injection doesn't reproduce what conhost does with a real Alt code.

The consequence for the fix is that the two arms can now both produce a character for the same keystroke, and only the record's own fields distinguish "this character already went out as a VT byte" from "this character exists only on the VK record".

---

## 4. An unterminated bracketed paste now withholds non-key events too

The ordering fix in `a6599bad` is right, and `test_pending_sequence_at_queue_front_blocks_later_event` covers the case it was written for. That case always resolves, because a lone `ESC` parses to `KeyCode::Esc` once `flush()` passes `more=false`.

The bracketed-paste case can't resolve. `parse_csi_bracketed_paste` returns `Ok(None)` regardless of `input_available` — deliberately, so a queue drain can't split a payload, which I noted approvingly last time — and `flush_impl` preserves the buffer on `Ok(None)`. With `buffer_event_position == 0` the new guard in `Iterator::next` then holds back everything queued behind it, and nothing can lift it.

Same reproducer against both trees:

```rust
let mut p = Parser::default();
p.advance(b"\x1b[200~hello", false);
p.push_event(InternalEvent::Event(Event::Resize(80, 24)));
p.push_event(mouse(7));
p.flush();
let drained: Vec<_> = std::iter::from_fn(|| p.next()).collect();
```

where `mouse(7)` is any `InternalEvent::Event(Event::Mouse(..))`.

| | `drained` |
|---|---|
| `3ca5429` | `[Resize(80, 24), Mouse(col 7)]` |
| `11e4912` | `[]` (`buffer.len() == 11`, `buffer_event_position == 0`) |

You already have the unterminated paste itself on the deferred list and I'm not asking to move it. The change is in what it costs when it happens: previously the application stopped receiving keys, and now it stops receiving `Event::Resize` as well, so it can't even redraw at the right size while wedged. If the eventual parser-hardening change caps the buffer, this goes away with it. It seemed worth recording next to that item rather than discovering it later.

---

## 5. Three symbols are dead on Windows now

A non-test Windows build:

```
cargo build --locked --lib --no-default-features --features windows,events
```

```
warning: function `parse_event_with_raw_mode` is never used
  --> src\event\sys\parse.rs:40:15
warning: methods `advance` and `flush` are never used
    --> src\event\sys\parse.rs:1773:19
```

Clean on `3ca5429`. Checking the three:

- `Parser::advance` (`:1773`) has no `allow` at all. Live on Unix (`source/unix/tty.rs:150`, `source/unix/mio.rs:99`), dead on Windows, which uses `advance_with_raw_mode`.
- `Parser::flush` (`:1853`) has `#[cfg_attr(unix, allow(dead_code))]`. It has no non-test caller on either platform; Windows uses `flush_with_raw_mode`.
- `parse_event_with_raw_mode` (`:40`) has `#[cfg_attr(not(windows), allow(dead_code))]` but is only called from tests, so Windows warns too.

Nothing fails on this: the warnings do show up in the Windows `test` job, which builds `--all-targets` and so compiles the non-test lib, but `-D warnings` only runs in the ubuntu clippy job. It reaches downstream Windows builds, though, which is why I'd rather mention it than not.

This is partly my fault for how I phrased the `cfg_attr` note last time. `push_event` is correct as written.

---

## A concern I raised and then withdrew

I thought the new `buffer_event_position` guard in `Iterator::next` might stall Unix, since `Parser` is shared and the Unix sources have no `flush()` call to lift it. It can't. The Unix sources only ever call `advance` and `next`, never `push_event`, so `internal_events` is populated exclusively by `insert_buffered_event`, which runs only when the buffer empties. A non-empty buffer therefore implies nothing was queued after the sequence began, and the guard can only ever refuse a pop that would have returned `None`. `cargo test --locked --lib --all-features` on Linux: 146 passed, 0 failed, 7 ignored.

Recording it because I nearly filed it.

---

## Method

Two trees at `3ca5429` and `11e4912` from the same clone, same toolchain, same machine, so every before/after above differs only in HEAD. Windows 11 26200, rustc 1.96.0; the Linux runs are the same source in WSL. Three programs behind it: the JSONL event logger from my earlier comment, the read-only raw record dump quoted above, and the two-tree parser reproducer. Happy to post any of them.

## Not covered

Same list as last time, none of it exercised on this HEAD: legacy console (`ForceV2=0`), the `disable_mouse_capture()` path, large pastes, a CRLF clipboard, WaveTerm. The re-check of bracketed paste across the terminal set is also not redone here, since nothing in this push touches the paste path I measured before. Say which of these would help and I'll run them.

@joshka For what it's worth, this looks close to done to me. Everything I raised in the adversarial pass has been dealt with, and what's left is one CI line plus two behaviors that fall out of the key-up fix and want a deliberate decision rather than more digging. I hope the review was some use in getting it to a state you can take another look at.

larsch:
I tested this PR through xi-agent on native Windows 11 (`10.0.26100`) in Windows Terminal. Bracketed paste works correctly, but enabling VT input causes a mouse-input regression: Crossterm no longer emits mouse events, so click/drag selection and wheel scrolling stop working.

The issue appears to be the interaction described in microsoft/terminal#15296. With `ENABLE_VIRTUAL_TERMINAL_INPUT` active, Windows Terminal/ConPTY delivers mouse input as VT sequences rather than Win32 `MOUSE_EVENT` records. Crossterm's Windows `EnableMouseCapture` deliberately takes only the Win32 path (`is_ansi_code_supported() == false`), so it sets `ENABLE_MOUSE_INPUT` but does not emit the VT mouse-tracking sequences.

I confirmed a local workaround: keep the existing Win32 `EnableMouseCapture`, and on Windows additionally write:

```text
CSI ? 1000 h
CSI ? 1002 h
CSI ? 1003 h
CSI ? 1015 h
CSI ? 1006 h
```

with the corresponding disable sequences in reverse order during cleanup:

```text
CSI ? 1006 l
CSI ? 1015 l
CSI ? 1003 l
CSI ? 1002 l
CSI ? 1000 l
```

With dual capture enabled, all of the following work in Windows Terminal:

- native `Event::Paste` for multiline bracketed paste;
- wheel scrolling;
- click/drag mouse selection;
- normal cleanup on exit.

I did not observe duplicated mouse events. The Win32 capture should remain enabled for conhost, which may continue delivering `MOUSE_EVENT` records.

Reproduction:

1. Enable raw mode.
2. Execute Crossterm `EnableMouseCapture` and `EnableBracketedPaste`.
3. Read events using this PR in Windows Terminal.
4. Bracketed paste works, but moving/clicking/scrolling produces no usable `Event::Mouse`.
5. Explicitly enable the VT mouse modes above; mouse events resume.

The review currently states that preventing ANSI execution for `EnableMouseCapture` means the terminal will not send SGR mouse reports and thus avoids duplicate delivery. That appears not to account for Windows Terminal switching away from Win32 mouse records when VT input is enabled. Supporting both capture mechanisms on Windows, or otherwise coordinating mouse capture with the new VT-input mode, seems necessary.

eitsupi:
Thank you both!
I'll take a look into these.

eitsupi:
@easyinplay @larsch Thank you for your feedback.
Based on it, I’ve revised the Windows input implementation and updated the PR description with the design rationale and validation results.

The latest version has been validated on both Windows Terminal and the modern Windows Console Host, including raw-mode key releases, bracketed paste, mouse input, Alt+numpad input, and console-mode restoration.

 Could you please take another look when you have a chance?

larsch:
Tested the latest PR HEAD (`94ed715`) through xi-agent on native Windows 11.

I removed xi-agent’s local VT mouse workaround and verified the upstream implementation directly in both Windows Terminal and modern Console Host (`conhost`). In both environments:

- bracketed paste still produces native paste events correctly;
- mouse-wheel scrolling works;
- click-and-drag text selection works;

This addresses the mouse regression I previously reported. I did not observe missing or duplicated mouse events. Thanks for incorporating the feedback.

elfenlieds7:
﻿Independent verification on native Windows 11 (Windows Terminal + ConPTY), testing HEAD 94ed715 through a downstream TUI app.

**Context:** we maintain an iocraft-based REPL (scode) that needs Event::Paste on Windows. crossterm's legacy Win32 input path never produces it, so we pinned iocraft's crossterm dep to this PR branch and exercised it end-to-end through a real ConPTY (via pty-expect), asserting on the VT100-rendered screen — i.e. what a user actually sees.

**Results — all green on Windows Terminal and modern conhost:**
- Multi-line bracketed paste produces a single Event::Paste (no more one-Enter-per-line). Verified with LF, CRLF, and bare-CR payloads.
- Ordinary keyboard typing keeps working while VT input is active for the paste lifetime (character keys flow through the shared ANSI parser correctly).
- Arrow keys, double-Ctrl-C exit, and general REPL redraw show no regression versus the legacy Win32 path.

**One unrelated nit** (pre-existing, not caused by this PR): 	erminal::sys::windows::tests::test_resize_winapi_20_21 fails with ERROR_INVALID_PARAMETER (os error 87) under Windows Terminal, which rejects programmatic console resize; 	erminal::tests::test_raw_mode also asserts a clean-console baseline that a piped/non-TTY harness violates. Both reproduce on master too, independent of this change — we just carry a local test-only guard that skips them when the host can't honor the operation.

Net: the shared-parser + VT-input-hybrid approach works cleanly for a real downstream consumer on Windows. Thanks for the thorough work here — would love to see this land.

PR #1038, opened 2026-02-28
Support for OSC 8 hyperlinks

### LINKED ISSUE #602: Add support for explicit hyperlinks/anchors (OSC 8)
See https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda for the specification and a list of things that already support/use this sequence.

cammeresi:
I tried to discover a preferred path for link support in the issue page, but I didn't receive any feedback.

I fully expect that you may not like this patch because it removes a couple of Copy impls, and so I have low expectations you will accept this patch as is, but I'm posting it in the hope that it can start a discussion and you can tell me the direction you'd prefer this implementation to go.

Fixes #602

### DIFF
--- a/examples/link.rs
+++ b/examples/link.rs
@@ -1,0 +1,23 @@
+//! Demonstrates OSC 8 hyperlink support.
+//!
+//! cargo run --example link
+
+use std::io;
+
+use crossterm::{
+    execute,
+    style::{EndHyperlink, Print, StartHyperlink},
+};
+
+fn main() -> io::Result<()> {
+    let mut out = io::stdout();
+
+    execute!(
+        out,
+        Print("Visit: "),
+        StartHyperlink::new("https://github.com/crossterm-rs/crossterm"),
+        Print("crossterm"),
+        EndHyperlink,
+        Print("\n"),
+    )
+}
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -56,6 +56,8 @@
 //!     [`ResetColor`](style/struct.ResetColor.html), [`SetColors`](style/struct.SetColors.html)
 //!   - Attributes - [`SetAttribute`](style/struct.SetAttribute.html), [`SetAttributes`](style/struct.SetAttributes.html),
 //!     [`PrintStyledContent`](style/struct.PrintStyledContent.html)
+//!   - Hyperlinks - [`StartHyperlink`](style/struct.StartHyperlink.html),
+//!     [`EndHyperlink`](style/struct.EndHyperlink.html)
 //! - Module [`terminal`](terminal/index.html)
 //!   - Scrolling - [`ScrollUp`](terminal/struct.ScrollUp.html),
 //!     [`ScrollDown`](terminal/struct.ScrollDown.html)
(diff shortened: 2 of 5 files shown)


joshka (changes requested):
I like the addition of `SetHyperlink` / `ResetHyperlink`. These feel like a good fit for crossterm: small, useful, and non-breaking. They expose the core OSC 8 capability in a way that composes naturally with the existing command API.

I’m much less convinced by adding `hyperlink` to `ContentStyle` and threading it through `StyledContent`, `SetStyle`, and `PrintStyledContent`.

My main concern is that this changes the semantics and cost model of the existing styling types in order to support something that feels orthogonal to styling. In particular, adding `Option<String>` to `ContentStyle` means `ContentStyle`, `StyledContent`, `SetStyle`, and `PrintStyledContent` are no longer `Copy`. That’s a fairly large knock-on effect for what is otherwise a relatively self-contained feature.

Related to that, this also introduces owned allocation into the styling path. Today `StyledContent` is generic over `Display`, so the textual content can be a `&str`, `String`, `Cow`, or some custom displayable type without forcing allocation. That feels important for terminal UI workloads, where large amounts of content are re-rendered frequently and avoiding unnecessary allocation matters. By contrast, `link()` now necessarily allocates a `String`, and that owned state then has to be carried and cloned through the styling types.

More broadly, hyperlink state doesn’t quite behave like other style attributes. It is more stateful and scoped at the terminal level, and the current approach also introduces slightly different semantics between `SetStyle` (stateful, no reset) and `PrintStyledContent` (scoped, with reset).

There is also a release management aspect here. Adding `SetHyperlink` / `ResetHyperlink` as new commands is semver-compatible, so that could go out in something like `crossterm 0.29.1`. By contrast, changing `ContentStyle` and removing `Copy` from these existing public types is a breaking change, so that would mean `crossterm 0.30.0`.

That is not just an abstract semver concern. It means downstream libraries such as Ratatui would then need to add support for the new crossterm version as well, and libraries depending on those crates can end up having to update in turn. So the practical impact is not just on crossterm itself, but on a fairly wide part of the ecosystem built on top of it.

I think there is a cleaner split available here:

* keep `SetHyperlink` / `ResetHyperlink` as standalone commands
* leave hyperlinks out of `ContentStyle`
* let users compose hyperlink behavior explicitly with the command API

For example:

```rust
queue!(
    stdout,
    SetHyperlink::new("https://example.com"),
    Print("click here"),
    ResetHyperlink,
)?;
```

That still provides the core capability, keeps the change non-breaking, and avoids baking hyperlink state into the general style model or changing the performance characteristics of existing code.

If we want a more ergonomic builder later, I’d prefer to explore that separately rather than coupling it to `ContentStyle` now.

joshka (commented):
It may be worth putting the hyperlink commands into a separate file, and re-exporting the commands from the main style. This keeps the tests by the impl.

(This is more of a personal preference thing - smaller modules)

General support of the approach. Thanks for the quick update!

joshka (changes requested):

joshka:
LGTM - thanks!

PR #1041, opened 2026-04-08
fix: open `/dev/tty` with correct permissions when querying for keyboard enhancement flags

awaken1ng:
`/dev/tty` was opened with read-only permissions, so it was always failing to write the query commands.


awaken1ng:
Oh wait, I'm blind, #1026 already exists.

PR #1045, opened 2026-04-16
WIP: Color palette querying and change detection

jake-stewart:
This PR introduces:
- querying terminal colors
- querying dark/light colorscheme
- colorscheme change notifications

`query_terminal_colors` takes a `ColorType` slice and returns RGB values by
using OSC 4 and 10–19. `query_color_scheme` uses `CSI ? 996 n` to ask whether
the terminal is in dark or light mode. `EnableColorSchemeDetection` and
`DisableColorSchemeDetection` toggle `CSI ? 2031 h/l`, and scheme changes come
through as `Event::ColorSchemeChanged`.

Both query functions put a trailing DA1 onto the request and read replies
until DA1 comes back. This matters more here than it would for
`cursor::position` because OSC color queries aren't as widely supported, so on
terminals that ignore them we'd otherwise wait the full 2000ms timeout.

Errors return `io::ErrorKind::Other` the same as `cursor::position`. Eventually
this could split into `TimedOut` and `Unsupported` (no DA1 vs DA1 with no
reply), but that applies equally to the cursor query, so I left them
consistent.

Manually tested on Kitty, Ghostty, and macOS Terminal.

https://contour-terminal.org/vt-extensions/color-palette-update-notifications/#example-source-code
https://ghostty.org/docs/vt/osc/4
https://ghostty.org/docs/vt/osc/1x

### DIFF
--- a/src/colors.rs
+++ b/src/colors.rs
@@ -1,0 +1,87 @@
+//! # Colors
+//!
+//! The `colors` module provides functionality to query the terminal for colors
+//! and color-related capabilities.
+//!
+//! * [`query_terminal_colors`] — fetch the RGB values of foreground, background,
+//!   cursor, or palette colors.
+//! * [`query_color_scheme`] — detect whether the terminal is in light or dark mode.
+//!
+//! Color scheme *change* notifications are delivered as
+//! [`Event::ColorSchemeChanged`](crate::event::Event::ColorSchemeChanged) when
+//! [`EnableColorSchemeDetection`](crate::event::EnableColorSchemeDetection) is active.
+
+pub(crate) mod sys;
+
+pub use sys::{query_color_scheme, query_terminal_colors};
+
+/// Terminal color type, used in queries and responses.
+///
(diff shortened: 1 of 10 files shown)


jake-stewart:
I am working on a query batching idea to avoid multiple round trips to the terminal once more queries are supported:

```rust
let mut batch = QueryBatch::new();
let kitty_query = batch.add(SupportsKittyGraphicsQuery);
let sixel_query = batch.add(SupportsSixelGraphicsQuery);
let xt_query = batch.add(XtVersionQuery);
let fg_query = batch.add(ColorQuery(ColorType::Foreground));
let bg_query = batch.add(ColorQuery(ColorType::Background));

let response = batch.execute()?;

let kitty_response = response.get(&kitty_query)?;
let sixel_response = response.get(&sixel_query)?;
let xt_response = response.get(&xt_query)?;
let fg_response = response.get(&fg_query)?;
let bg_response = response.get(&bg_query)?;
```

PR #1046, opened 2026-04-18
add mouse capture level

jake-stewart:
This PR introduces granular mouse tracking via `MouseCaptureLevel` (Click, Drag, Hover).

This is a breaking change as `EnableMouseCapture` now requires a `MouseCaptureLevel`.
However, this will typically be one trivial code change for users.

Examples and doc-tests are updated.

### DIFF
--- a/examples/event-poll-read.rs
+++ b/examples/event-poll-read.rs
@@ -6,7 +6,7 @@ use std::{io, time::Duration};
 
 use crossterm::{
     cursor::position,
-    event::{poll, read, DisableMouseCapture, EnableMouseCapture, Event, KeyCode},
+    event::{poll, read, DisableMouseCapture, EnableMouseCapture, Event, KeyCode, MouseCaptureLevel},
     execute,
     terminal::{disable_raw_mode, enable_raw_mode},
 };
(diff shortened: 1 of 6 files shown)


joshka:
This is the sort of "trivial change" that has a large blast radius. [search results 8k](https://github.com/search?q=EnableMouseCapture+crossterm+AND+%28NOT+is%3Afork%29+AND+%28NOT+org%3Acrossterm-rs+%29&type=code). I wonder if you'd consider a more additive change (new item perhaps or some other option?)

jake-stewart:
> This is the sort of "trivial change" that has a large blast radius

I've considered `EnableMouseCaptureLevel(MouseCaptureLevel)` and then the existing `EnableMouseCapture` remains as a shorthand for `EnableMouseCaptureLevel(MouseCaptureLevel::Hover)`.

joshka:
Thinking through this a bit more. This PR also introduces some shared state (which mode is active) that doesn't seem justified by there being an expressed need to track the modes.

I think a reasonable approach here would be more like adding each csi command as a pair and letting the user handle their choice of state. E.g. something like:

```rust
struct EnableMouseClickEvents;
struct EnableMouseHoverEvents;
struct EnableMouseDragEvents;
```

Maybe in a mouse::advanced (not sure about naming?) module? Or at the very least the docs should point the user to the main mouse capture command as the primary use case (to avoid confusion).

jake-stewart:
I put it in `event::advanced` so that it is next to `event::EnableMouseCapture` etc.

I think `event::advanced` works ok because in the future it can be used generally for advanced things.

PR #1047, opened 2026-04-18
fix tty poll fail for zero timeout

jake-stewart:
A zero timeout never enters the loop. We must check for buffered events beforehand as mio.rs does already.

### DIFF
--- a/src/event/source/unix/tty.rs
+++ b/src/event/source/unix/tty.rs
@@ -123,11 +123,11 @@ impl EventSource for UnixInternalEventSource {
             make_pollfd(&self.wake_pipe.receiver),
         ];
 
+        // check if there are buffered events from the last read
+        if let Some(event) = self.parser.next() {
+            return Ok(Some(event));
+        }
         while timeout.leftover().map_or(true, |t| !t.is_zero()) {
-            // check if there are buffered events from the last read
-            if let Some(event) = self.parser.next() {
-                return Ok(Some(event));
-            }
             match poll(&mut fds, timeout.leftover()) {
                 Err(filedescriptor::Error::Poll(e)) | Err(filedescriptor::Error::Io(e)) => {
                     match e.kind() {


joshka:
Hey, can you give more details about the problem / solution here?
Lay some context / rationale on me to help show what's happening.
Assume that I have no idea about why this path is broken.

jake-stewart:
I handle events in my TUI program in batches instead of one at a time. This lets me only draw the screen after an entire batch instead of after each event.

This is especially useful for scrolling and non-bracketed paste since these can spam tens or hundreds of events at once.

To build a batch of events, I first do a blocking [read](https://docs.rs/crossterm/latest/crossterm/event/fn.read.html). I then enter a loop, using [poll](https://docs.rs/crossterm/latest/crossterm/event/fn.poll.html) with a 0ms timeout to check whether there are remaining events, and read them if so. This lets me build up a batch of events while only blocking on the first event.

This is the [section of the code which contains the bug](https://github.com/crossterm-rs/crossterm/blob/c02a080af67c3d9f46af3abefb94b2c34744aa47/src/event/source/unix/tty.rs#L127-L130):

```rust
while timeout.leftover().map_or(true, |t| !t.is_zero()) {
    // check if there are buffered events from the last read
    if let Some(event) = self.parser.next() {
        return Ok(Some(event));
    }
    // polls for an event
}
Ok(None)
```

A 0ms timeout will never enter the while loop because the `!t.is_zero()`. It skips the polling behaviour entirely and returns a `Ok(None)`.

The fix is to move it outside the loop:

```rust
// check if there are buffered events from the last read
if let Some(event) = self.parser.next() {
    return Ok(Some(event));
}
while timeout.leftover().map_or(true, |t| !t.is_zero()) {
    // polls for an event
}
Ok(None)
```

This matches [what mio.rs does does already](https://github.com/crossterm-rs/crossterm/blob/c02a080af67c3d9f46af3abefb94b2c34744aa47/src/event/source/unix/mio.rs#L70-L72).

I believe this bug has existed undetected for a few reasons:

1. Most users are not using `use-dev-tty` and so it works correctly for them.
2. Only 0ms timeout is affected.
3. It can be hard to notice. Since polling for 0ms always returns false my program behaves like I handle each event individually. All I lose is the batching optimization.

joshka:
Semi-related (but not yet released) We added a try_read() function at the border - does that hit this path or does it skip it? https://github.com/crossterm-rs/crossterm/pull/1003

One thing I do wonder here (haven't checked in depth yet) is whether moving this out of the loop drops a stopping condition. My concern / lack of knowledge is whether there can be an extra item added to the events outside of the path where that would show up in the other loop stopping conditions. I suspect not, but the complexity of the parts involved make my feel like this is worth checking.

jake-stewart:
The user facing `read` calls `try_read` to pop an event from the queue, or
calls `poll` if none are available.

This `poll` calls a different `try_read`, the `try_read` of its `EventSource`.
There are different event sources depending on windows, unix, `use-dev-tty`.

For the mio.rs event source, `try_read` first tries to parse its buffer
for the next event. It will poll/read more bytes if this fails.

The tty.rs event source has this exact same pattern and it tries to parse
its buffer for the next event before doing any poll/read logic.

The problem is that this "parse buffer for next event" is put inside the
poll/read loop. This poll/read loop only runs when there is None timeout
(block forever) or positive timeout (block for an amount of time). With zero timeout,
this loop is never entered, so the next event is not parsed.

This fix moves the parse outside of the loop where it belongs (where mio.rs
puts it), so that when called with a zero timeout, it still parses the next event
instead of immeditately returning None despite there being potential unparsed events
ready to use.

PR #1049, opened 2026-04-24
fix(parse): clamp instead of underflow in mouse / cursor-position parsers

mattwilkinsonn:
`parse_csi_normal_mouse` panicked in debug builds (wraps to 65535 in
release) whenever the column or row byte was exactly 32 — the X10
mouse protocol's encoding of column 0 / row 0. I hit it in a TUI built
on crossterm 0.29.0:

```text
thread 'main' panicked at crossterm-0.29.0/src/event/sys/unix/parse.rs:707:14
attempt to subtract with overflow
```

The math at that line is `saturating_sub(32) - 1`, which underflows
when the input is exactly 32. The parser dispatches on byte pattern,
so the bad input also reaches it from pastes containing terminal
escape data and from subprocess output emitting VT100 `CSI M` (Delete
Line) — not just from real mouse events at the top-left corner.

Folded into a single `saturating_sub(33)`, which is equivalent for
every legal input and clamps to 0 instead of underflowing.

The same `next_parsed::<u16>(…)? - 1` pattern is in
`parse_csi_rxvt_mouse`, `parse_csi_sgr_mouse`, and
`parse_csi_cursor_position`; switched those to `saturating_sub(1)`
for consistency. The `Cb` (button) bytes in the same parsers already
use `checked_sub` + `could_not_parse_event_error` and are unchanged —
silent clamp is right for coordinates, parse-error is right for the
button identity.

Regression tests added for all four parsers. CHANGELOG bullet added
under `# Unreleased`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Unreleased
 
+## Fixed 🐛
+
+- Fix integer underflow in mouse / cursor-position parsers when coord
+  bytes encoded the protocol origin (panic in debug, wrap to 65535
+  in release). Affects `parse_csi_normal_mouse`, `parse_csi_rxvt_mouse`,
+  `parse_csi_sgr_mouse`, and `parse_csi_cursor_position`.
+
 # Version 0.29
 
 ## Added ⭐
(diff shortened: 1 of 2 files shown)


joshka (approved):

joshka:
LGTM!

PR #1050, opened 2026-04-28
Query batch

jake-stewart:
I've needed several queries and doing them sequentially is bad since each is a round trip. This PR introduces a `QueryBatch` which sends all queries in a single write and reads responses in one pass, using `DA1` as a sentinel that marks the end of the batch.

The queries themselves will be introduced in subsequent PRs so each can be reviewed independently. The example below uses those upcoming queries to show what the API looks like:

```rust
fn main() -> std::io::Result<()> {
    let mut batch = QueryBatch::new()
        .timeout(Duration::from_millis(250));

    let version = batch.add(QueryXtVersion);
    let cursor = batch.add(QueryCursorPosition);
    let window = batch.add(QueryWindowPixelSize);
    let cell = batch.add(QueryCellPixelSize);
    let kitty = batch.add(QueryKittyGraphicsSupport);
    let sixel = batch.add(QuerySixelSupport);
    let fg = batch.add(QueryColor(ColorType::Foreground));
    let bg = batch.add(QueryColor(ColorType::Background));

    enable_raw_mode()?;
    let results = batch.execute()?;
    disable_raw_mode()?;

    println!("terminal:    {:?}", results.get(&version)?);
    println!("cursor:      {:?}", results.get(&cursor)?);
    println!("window px:   {:?}", results.get(&window)?);
    println!("cell px:     {:?}", results.get(&cell)?);
    println!("kitty gfx:   {}", results.get(&kitty)?);
    println!("sixel:       {}", results.get(&sixel)?);
    println!("fg:          {:?}", results.get(&fg)?);
    println!("bg:          {:?}", results.get(&bg)?);

    Ok(())
}
```

EDIT: I've created a lot of PRs recently. If you want to move somewhere to discuss these somewhere more effectively, I am open to that.

### DIFF
--- a/src/cursor/sys/unix.rs
+++ b/src/cursor/sys/unix.rs
@@ -1,5 +1,5 @@
 use std::{
-    io::{self, Error, ErrorKind, Write},
+    io::{self, Error, ErrorKind},
     time::Duration,
 };
 
@@ -40,10 +40,8 @@ fn read_position_raw() -> io::Result<(u16, u16)> {
         let _ = internal::read(&CursorPositionFilter);
     }
 
-    // Use `ESC [ 6 n` to and retrieve the cursor position.
-    let mut stdout = io::stdout();
-    stdout.write_all(b"\x1B[6n")?;
-    stdout.flush()?;
+    // Use `ESC [ 6 n` to retrieve the cursor position.
+    crate::event::write_query(b"\x1B[6n")?;
 
     loop {
         match internal::poll(Some(Duration::from_millis(2000)), &CursorPositionFilter) {
--- a/src/event.rs
+++ b/src/event.rs
@@ -265,6 +265,19 @@ pub fn try_read() -> Option<Event> {
     }
 }
 
+/// Writes a terminal query. Avoids writing to stdout when `use-dev-tty` is enabled since it may be piped.
+#[cfg(unix)]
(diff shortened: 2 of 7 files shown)


joshka:
Hey, great idea in general. Something like this makes sense generally to start as an issue to discuss alternate ideas etc. and get feedback on.

Here, I wonder if something kind of symmetrical to the queue / execute macros might be more in keeping with how crossterm works?

```rust
let version = queue_query!(QueryXtVersion);
let cursor = queue_query!(QueryCursorPosition);
queue_execute!();
```
or 
```rust
let (version, cursor) = queue_query!(QueryXtVersion, QueryCursorPosition);
```

just an idea - not a well thought out one, but something to consider if you're willing to take some time to step back from the implementation. There might be ones with similar shapes.

jake-stewart:
We should decide whether query construction can fail since `execute_winapi` tends to return `Unsupported`, we may want an unsupported query to fail similarly at queue time. Alternatively, we can return the `Unsupported` after execution.

Should also keep in mind programmability. You may want to query multiple colors or only optionally perform queries, which makes the tuple approach less viable. 

I think this approach would work nicely. The query execute can run for all queued queries on the first `.get()`.

```rust
let version = query!(QueryXtVersion)?;
let cursor = query!(QueryCursorPosition)?;
let version = version.get()?;
let cursor = cursor.get()?;
```

PR #1052, opened 2026-04-28
Color scheme events

jake-stewart:
Adds live notifications for terminal color scheme changes (e.g. system dark/light toggle) on top of #1051.

```rust
execute!(stdout(), EnableColorSchemeDetection)?;

loop {
  if let Event::ColorSchemeChanged(scheme) = event::read()? {
      println!("scheme is now {:?}", scheme);
  }
}
```

### DIFF
--- a/src/colors.rs
+++ b/src/colors.rs
@@ -1,0 +1,195 @@
+//! # Colors
+//!
+//! Terminal capability queries for color values and the color scheme
+//! (light vs. dark mode).
+
+/// Terminal color type, used in queries and responses.
+///
+/// `Palette(n)` uses OSC 4. The remaining variants use OSC 10..=19.
+#[derive(Debug, PartialOrd, PartialEq, Hash, Clone, Copy, Eq)]
+pub enum ColorType {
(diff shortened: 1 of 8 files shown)


fritzfeger:
Hi, thanks for working on this @jake-stewart. I have a downstream TUI use case that needs both the initial color-scheme query and live change events: each terminal session follows its own appearance, temporarily unsubscribing while an editor owns the terminal and querying again on resume.

The integration currently works against a small Crossterm fork and is covered by downstream tests for lifecycle handling and rapid dark/light changes. I would prefer to migrate to the canonical API rather than open overlapping work. #1106 also appears to cover the parser robustness needed for these reports.

Is #1052 still the intended direction for color-scheme events? If so, would additional downstream testing or a focused contribution be useful, or would you prefer consumers to wait for the current series to land?

jake-stewart:
@fritzfeger I cannot say what the intended direction is because I am not associated with the project.

I tried to follow crossterm's style closely though, and it is built on top of #1050 which has generally positive feedback.

PR #1057, opened 2026-05-12
fix(event): drain tty fd to EAGAIN before returning from try_read

sinelaw:
mio registers the tty fd with `Interest::READABLE`, which on Linux is `EPOLLIN | EPOLLET` (edge-triggered). The read loop returned as soon as the parser yielded its first event, leaving any remaining bytes in the kernel buffer. The edge has already been consumed, so `epoll_wait` won't re-fire until *new* data arrives.

Symptom: a large paste (e.g. tmux `paste-buffer`, no bracketed paste) stalls partway through and only resumes on the next keypress, which creates a new edge and drains the leftovers.

Fix: drain to `WouldBlock` — or a short read, which on a tty signals the same — before returning. Events queue in the parser; subsequent `try_read` calls drain them with no extra syscalls. The typical small-input fast path is unchanged: one short read, exit.

Repro inside tmux:

```
tmux load-buffer some-long-file.md
tmux paste-buffer -t <target-pane>
```

Tested on tmux 3.4, Linux 6.18. The `use-dev-tty` path uses level-triggered `poll()` and is unaffected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Unreleased
 
+## Fixed 🐛
+
+- Drain tty fd to `WouldBlock` so large pastes don't stall under edge-triggered epoll (#1057)
+
 # Version 0.29
 
 ## Added ⭐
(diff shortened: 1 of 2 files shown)


sinelaw:
I would like to add a test for this but it requires some kind of harness that actually uses a real PTY.

PR #1058, opened 2026-05-14
fix clippy lints

xtqqczze:


examples/stderr.rs
```diff
@@ -55,10 +55,8 @@ where
         Hide                  // hide the cursor
     )?;
 
-    let mut y = 1;
-    for line in TEXT.split('\n') {
+    for (y, line) in (1..).zip(TEXT.split('\n')) {
```
joshka, line 58: nit: I think I'd normally go something like `TEXT.split('\n').enumeratet().skip(1)`, if that works

joshka (approved):

examples/stderr.rs
joshka: pro tip - it's not the same :)

PR #1059, opened 2026-05-14
Raise the MSRV to Rust 1.85

xtqqczze:
## Summary

- raise Crossterm's declared minimum supported Rust version from 1.63 to 1.85
- define the MSRV as covering the library with no default features and with all public features enabled
- add a required CI job that checks both configurations with exactly Rust 1.85.0
- record the compatibility change in the changelog

Part of #1075.

## Why Rust 1.85?

Rust 1.63 was introduced in #892 because adopting Rustix required it. The repository does not record a Debian support policy as the reason for that choice, although Rust 1.63 subsequently matched the compiler shipped by Debian Bookworm.

That compatibility point has now moved:

- [Debian Trixie, the current stable release, packages rustc 1.85.0](https://packages.debian.org/stable/rustc).
- [Rust 1.85.0 was released on February 20, 2025](https://blog.rust-lang.org/2025/02/20/Rust-1.85.0.html), about 17 months before this change.
- Crossterm's currently locked all-public-features graph already resolves `unicode-segmentation 1.13.3` through `derive_more`; that release requires Rust 1.85.

Keeping the declaration at 1.63 would therefore overstate what the current all-features dependency graph supports. Preserving 1.63 would require selecting and maintaining older compatible transitive releases rather than merely retaining working source compatibility.

Rust 1.85 is still well behind current stable Rust while being available in current Debian stable, so it provides a conservative floor without holding Crossterm to the previous Debian release's compiler indefinitely.

## What the MSRV covers

The MSRV is a promise about Crossterm as a library:

- `cargo check --locked --lib --no-default-features`
- `cargo check --locked --lib --all-features`

Both commands are now run in CI with exactly Rust 1.85.0. Development tools and examples do not independently define the library's MSRV; their dependencies may have a newer floor without changing Crossterm's consumer-facing promise. This follows Cargo's distinction between the package [`rust-version`](https://doc.rust-lang.org/cargo/reference/rust-version.html) and the toolchain used for development.

## Source changes

Changing `package.rust-version` also changes Clippy's MSRV-aware suggestions. The existing `-D clippy::all` check then requires the included `std::io::Error::other` and `Option::is_none_or` simplifications. These are the only source rewrites in this PR; they are mechanical consequences of the compatibility change rather than a broader lint cleanup.

`Cargo.lock` and all dependency requirements remain unchanged.

## Validation

- `rustup run 1.85.0 cargo check --locked --lib --no-default-features`
- `rustup run 1.85.0 cargo check --locked --lib --all-features`
- `cargo clippy --locked --all-features -- -D clippy::all`
- `cargo test --locked --all-targets --all-features -- --test-threads 1`
- `cargo test --locked --doc --all-features -- --test-threads 1`
- `cargo package --locked --allow-dirty`
- `actionlint .github/workflows/ci.yml`
- `zizmor .github/workflows/ci.yml`


joshka:
Thanks for starting this. I have brought the branch up to date with `master` and added:

- the changelog entry and explicit MSRV scope;
- Rust 1.85.0 checks for the minimal and all-public-features library configurations;
- the rationale around the current dependency graph, Debian Trixie, and the age of Rust 1.85;
- the equivalent MSRV-aware Clippy cleanups in Windows-only paths.

Changing `package.rust-version` makes Clippy enable suggestions that were unavailable at Rust 1.63, which is why the mechanical source changes are required for the existing CI rather than being a general cleanup. All seven CI jobs now pass.

xtqqczze:
We can now raise the edition to 2024: https://github.com/crossterm-rs/crossterm/pull/1080

joshka:
yup - I'm working through the stuff in #1075 in order here. Apologies that we overlapped on this on #1080 / #1081

xtqqczze:
I think it would be possible to lower the MSRV to 1.79, it just requires pinning `unicode-segmentation` to v1.12.

Issue #1060, opened 2026-05-21
Bracketed paste parser hangs event stream indefinitely if `\x1B[201~` never arrives


DannyStoll1:
## Describe the bug

When bracketed paste mode is enabled, crossterm's internal `Parser`
(`src/event/source/unix/{tty,mio}.rs`) accumulates every byte received between
the opening `ESC [ 200 ~` and closing `ESC [ 201 ~` markers into a single
in-flight `Event::Paste`. The accumulator buffer has no upper bound, neither
in time nor in bytes:

  https://github.com/crossterm-rs/crossterm/blob/0.29/src/event/sys/unix/parse.rs#L817-L822

```rust
if !buffer.ends_with(b"\x1b[201~") {
    Ok(None)
} else {
    let paste = String::from_utf8_lossy(&buffer[6..buffer.len() - 6]).to_string();
    Ok(Some(InternalEvent::Event(Event::Paste(paste))))
}
```

If the closing marker never arrives, the parser silently swallows every
subsequent byte read from the tty. The application keeps reading keystrokes from the
OS, but it never receives an `Event`. From the user's perspective, the application
appears to have stopped listening to their input.

There is currently no public API to flush or reset the parser's internal state, so there is no good recovery option. At present, applications/users are essentially limited to three options:
 - attempt a graceful shutdown (only possible if the application has a watchdog in place to detect the issue),
 - try to hack the event stream back into a valid state by emulating a `\x1B[201~` input or issuing another bracketed pasted (hard for a user to know this is what the problem is, and will emit a ton of garbage input like mouse movements), or
 - kill the process, losing any unsaved state.

This condition is reachable under normal operation:

  - **tmux pane swap**: I initially encountered this bug when the text I was pasting
    into the application triggered a tmux swap to a different pane.
    This meant the closing marker never reached the application's pane.
  - **Misbehaving terminal emulator**: Some terminal emulators (kitty,
    iTerm2 under certain configurations) can drop the close marker if focus
    is moved away mid-paste or if the host renegotiates terminal modes
    between the opening and closing markers.
  - **Adversarial input**: A program that prints `\x1B[2004h\x1B[200~` to a
    terminal hosting a `crossterm` application can wedge it by never
    sending the close marker. See "Failure modes worth flagging" below.

## To Reproduce

Minimal repro using the existing (private) parser tests:

```rust
use crossterm::event::sys::unix::parse::parse_event;
[...]

PR #1061, opened 2026-05-21
Bracketed paste limits

DannyStoll1:
Provides an opt-in mitigation for #1060.

# Summary

Adds an opt-in `BracketedPasteLimits` API that bounds inflight bracketed
pastes by wall-clock duration and/or accumulated bytes. When either bound
is exceeded, the parser hands the buffered bytes back to the caller as a new
`Event::PasteAborted { reason, buffered_bytes }` event and resumes normal
parsing.

Without this, an orphaned `\x1B[200~` will cause crossterm to swallow every
subsequent byte from the tty indefinitely. See #1060 for full discussion.

Both new limit fields default to `None`, so behavior is unchanged for existing callers.

# Changes

This PR is split into two commits to keep review manageable.

  0. ed36c4e
     Pure code motion. `src/event/source/unix/{tty,mio}.rs` each carried
     identical copies of a `Parser` struct. This deduplicates the code by extracting
     the implementation to a new shared `src/event/source/unix/parser` module.
     No effect on behavior or public API.

  1. 0695013
     The actual fix. Adds:

       - `pub struct BracketedPasteLimits { max_duration, max_bytes }`
       - `pub fn set_bracketed_paste_limits(BracketedPasteLimits)`
       - `pub fn bracketed_paste_limits() -> BracketedPasteLimits`
       - `pub enum PasteAbortReason { Timeout, SizeLimit }`
       - new `Event::PasteAborted { reason, buffered_bytes: Vec<u8> }` variant

     The limits storage lives behind two relaxed atomics (`AtomicU64` for
     duration in millis, `AtomicUsize` for bytes, with sentinel values for
     `None`). These are snapshotted once per read and threaded into the per-byte abort check.

     When an abort fires, the accumulated buffer (including the
     `ESC [ 200 ~` prefix) is moved into the event so the caller can
     salvage it; the parser's buffer is replaced with a fresh
     `Vec::with_capacity(256)` to preserve the original allocation
     behaviour for subsequent escape sequences.

# API surface

```rust
use std::time::Duration;
use crossterm::event::{set_bracketed_paste_limits, BracketedPasteLimits};

set_bracketed_paste_limits(BracketedPasteLimits {
    max_duration: Some(Duration::from_secs(5)),
    max_bytes: Some(1024 * 1024),
});
```

After this, an inflight paste that runs longer than 5s or exceeds 1 MiB
will be aborted:

```rust
match event::read()? {
    Event::PasteAborted { reason, buffered_bytes } => {
        log::warn!("paste aborted ({reason:?}); recovered {} bytes", buffered_bytes.len());
    }
    Event::Paste(s) => { /* normal complete paste */ }
    ...
}
```

# Compatibility

  * Both new limit fields default to `None`, so the parser behaves
    identically to today's unbounded accumulator unless the caller opts in.
  * `Event::PasteAborted` is a new variant gated behind the existing
    `bracketed-paste` feature flag, so it lands in the same conditional
    compilation region as `Event::Paste` and only affects callers who have
    that feature enabled.
    - Still, since `Event` is not marked `#[non_exhaustive]`, this is
      a **breaking change** and warrants a version bump.
  * No new runtime dependencies.

# Tests

5 new tests in `src/event/source/unix/parser.rs::tests`, all gated on
`feature = "bracketed-paste"`:

  - `completes_paste_within_size_limit`
  - `aborts_paste_when_size_limit_exceeded`
  - `aborts_paste_when_timeout_exceeded`
  - `parser_recovers_to_normal_input_after_abort`
  - `limits_default_unbounded_preserves_old_behaviour`

The limits are process-global, so the tests serialize on an internal
`parking_lot::Mutex` and restore the previously-installed limits on drop
to avoid cross-test interference.

# E2E reproduction

I wrote a standalone end-to-end reproduction that uses only the public
crossterm API: it opens a pty, forks, has the child enable bracketed paste
and drain `event::read()` into a log, then has the parent write `\x1B[200~`
into the master fd followed by plain keystrokes without ever sending the
close marker.

<details><summary>Reproduction code</summary>

```rust
//! End-to-end reproduction of the bracketed-paste hang in crossterm 0.29.0.
//!
//! Setup:
//!   * The parent process opens a PTY pair, forks, and the child re-execs
//!     itself with `child` as the first argument. The child's stdin / stdout /
//!     stderr are wired to the slave end of the PTY, so the child sees a real
//!     terminal and crossterm exercises the same code path it would in
//!     production.
//!   * The child enables bracketed paste, switches into raw mode, and
//!     `event::read`s in a loop for a few seconds. Every event it observes is
//!     appended to a log file at `/tmp/ct-repro-events.log` (chosen so the
//!     parent can read it back without interleaving with PTY traffic).
//!   * The parent writes the *opening* bracketed-paste marker `ESC [ 200 ~`
//!     into the master end, waits briefly, then writes a sequence of plain
//!     keystrokes (`abcdef`). It never writes the closing `ESC [ 201 ~`
//!     marker, simulating a terminal / multiplexer that dropped it.
//!   * After a few seconds the parent reaps the child, prints the contents of
//!     the log file, and reports whether any keystroke events were observed.
//!
//! Expected result on crossterm 0.29.0 (no limits): the log file is empty
//! (the child never received an event because every byte got swallowed by the
//! orphan-paste buffer).
//!
//! Expected result on a build with bracketed-paste limits installed: the log
//! file contains a `PasteAborted` event followed by `Key` events for each
//! plain keystroke.

use std::env;
use std::ffi::CString;
use std::fs::{File, OpenOptions};
use std::io::{Read, Write};
use std::os::fd::{AsRawFd, OwnedFd};
use std::os::unix::io::FromRawFd;
use std::path::Path;
use std::process::ExitCode;
use std::thread;
use std::time::{Duration, Instant};

use crossterm::event::{
    self, DisableBracketedPaste, EnableBracketedPaste, Event, KeyCode, KeyEventKind,
};
use crossterm::execute;
use crossterm::terminal::{disable_raw_mode, enable_raw_mode};
use nix::pty::{openpty, OpenptyResult};
use nix::sys::wait::{waitpid, WaitPidFlag, WaitStatus};
use nix::unistd::{dup2, execv, fork, ForkResult, Pid};

const EVENT_LOG_PATH: &str = "/tmp/ct-repro-events.log";
const CHILD_RUN_DURATION: Duration = Duration::from_secs(4);
const PARENT_WEDGE_DELAY: Duration = Duration::from_millis(250);
const PARENT_KEYS_DELAY: Duration = Duration::from_millis(500);

fn main() -> ExitCode {
    let args: Vec<String> = env::args().collect();
    if args.get(1).map(String::as_str) == Some("child") {
        match run_child() {
            Ok(()) => ExitCode::SUCCESS,
            Err(e) => {
                let _ = writeln!(std::io::stderr(), "child error: {e}");
                ExitCode::from(1)
            }
        }
    } else {
        match run_parent() {
            Ok(code) => code,
            Err(e) => {
                eprintln!("parent error: {e}");
                ExitCode::from(1)
            }
        }
    }
}

/// Parent: opens a PTY, forks, drives the master end, reports on the log.
fn run_parent() -> Result<ExitCode, Box<dyn std::error::Error>> {
    // Clean up any previous run so we never accidentally read stale events.
    let _ = std::fs::remove_file(EVENT_LOG_PATH);

    let OpenptyResult { master, slave } = openpty(None, None)?;
    let master: OwnedFd = master;
    let slave: OwnedFd = slave;

    let self_exe = env::current_exe()?;
    let exe_c = CString::new(self_exe.as_os_str().to_string_lossy().as_bytes())?;
    let arg_c = CString::new("child")?;

    // SAFETY: We only call async-signal-safe functions in the child branch
    // (dup2, execv). No allocations, no Rust-level locks acquired between fork
    // and exec.
    match unsafe { fork()? } {
        ForkResult::Child => {
            // Wire the slave end to stdin/stdout/stderr.
            let slave_raw = slave.as_raw_fd();
            dup2(slave_raw, 0)?;
            dup2(slave_raw, 1)?;
            dup2(slave_raw, 2)?;
            // Close the master / slave fds we no longer need.
            drop(master);
            drop(slave);
            execv(&exe_c, &[&exe_c, &arg_c])?;
            unreachable!()
        }
        ForkResult::Parent { child } => {
            drop(slave);
            drive_parent(master, child)
        }
    }
}

fn drive_parent(master: OwnedFd, child: Pid) -> Result<ExitCode, Box<dyn std::error::Error>> {
    // SAFETY: nix's OwnedFd owns the raw fd; converting to File transfers
    // ownership for the lifetime of `file`.
    let master_fd = master.as_raw_fd();
    let mut master_writer = unsafe { File::from_raw_fd(master_fd) };
    // Prevent the OwnedFd's drop from closing the fd a second time.
    std::mem::forget(master);

    // Drain the PTY output in a background thread so the slave's write buffer
    // never fills up and blocks the child. We discard the bytes; the
    // interesting signal is the event log.
    let drain_fd = master_writer.try_clone()?;
    let drainer = thread::spawn(move || {
        let mut f = drain_fd;
        let mut buf = [0u8; 1024];
        loop {
            match f.read(&mut buf) {
                Ok(0) | Err(_) => break,
                Ok(_) => {}
            }
        }
    });

    // Wait for the child to install its event loop before sending anything.
    thread::sleep(PARENT_WEDGE_DELAY);

    // 0. Open the paste without ever closing it.
    master_writer.write_all(b"\x1B[200~")?;
    master_writer.flush()?;

    // 1. Pretend the user types real keys after the paste was "interrupted".
    thread::sleep(PARENT_KEYS_DELAY);
    master_writer.write_all(b"abcdef")?;
    master_writer.flush()?;

    // 2. Give the child time to (try to) process the input.
    thread::sleep(CHILD_RUN_DURATION);

    // 3. Reap the child. It exits on its own after its run duration elapses;
    // if it's wedged, send SIGTERM so we don't hang forever.
    let status = waitpid(child, Some(WaitPidFlag::WNOHANG))?;
    if matches!(status, WaitStatus::StillAlive) {
        nix::sys::signal::kill(child, nix::sys::signal::Signal::SIGTERM)?;
        waitpid(child, None)?;
    }

    // 4. Closing the master end will end the drainer thread.
    drop(master_writer);
    let _ = drainer.join();

    // 5. Report.
    println!("===== ct-repro report =====");
    let log = std::fs::read_to_string(EVENT_LOG_PATH).unwrap_or_default();
    if log.is_empty() {
        println!("event log is EMPTY -- the parser swallowed every byte we sent.");
        println!("the child never observed a single Event::Key for 'abcdef'.");
        println!("this is the bug: an orphan bracketed-paste prefix wedges crossterm.");
        Ok(ExitCode::from(2))
    } else {
        println!("event log:\n{log}");
        let key_lines = log.lines().filter(|l| l.contains("Key(")).count();
        let abort_lines = log.lines().filter(|l| l.contains("PasteAborted")).count();
        println!(
            "summary: {key_lines} Key event(s), {abort_lines} PasteAborted event(s)",
        );
        Ok(ExitCode::SUCCESS)
    }
}

/// Child: enables bracketed paste, drains events into a log file.
fn run_child() -> Result<(), Box<dyn std::error::Error>> {
    let mut log = OpenOptions::new()
        .create(true)
        .append(true)
        .truncate(false)
        .open(Path::new(EVENT_LOG_PATH))?;

    enable_raw_mode()?;
    let mut stdout = std::io::stdout();
    execute!(stdout, EnableBracketedPaste)?;

    let deadline = Instant::now() + CHILD_RUN_DURATION;
    while Instant::now() < deadline {
        // Bound each poll so we re-check the deadline; otherwise a wedged
        // parser would block here forever and we'd never write our log.
        if event::poll(Duration::from_millis(100))? {
            let ev = event::read()?;
            writeln!(log, "{:?}", ev)?;
            log.flush()?;
            if let Event::Key(k) = ev {
                if k.kind == KeyEventKind::Press && k.code == KeyCode::Char('q') {
                    break;
                }
            }
        }
    }

    let _ = execute!(stdout, DisableBracketedPaste);
    let _ = disable_raw_mode();
    Ok(())
}
```
</details>

Output against the two builds:

  * crossterm 0.29.0 (c02a080): event log is empty, no `Key` events
    observed for 4 seconds, application is wedged.
  * this PR (with limits installed): 1 `PasteAborted` event (containing
    `[27, 91, 50, 48, 48, 126, 97]`, i.e. `ESC [ 200 ~ a`) followed by
    `Key('b' .. 'f')`. The parser recovered cleanly and even surfaced the
    byte the user happened to type before the abort fired.

# Verification per commit

  * `cargo test --all-features --lib` at the refactor commit: 112 passed.
  * `cargo test --all-features --lib` at this commit: 117 passed (5 new).
  * `cargo fmt --all -- --check`: clean.
  * `cargo build --all-features`: no new warnings.
  * `cargo build --no-default-features`: one pre-existing warning
    (`method 'read' is never used` on `FileDesc`) that reproduces on
    `upstream/master` and is unrelated to this PR.

# CHANGELOG

A single entry under `# Unreleased / ## Added`, following the existing
single-line bullet convention.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Unreleased
 
+## Added ⭐
+
+- Configurable bracketed-paste limits via `set_bracketed_paste_limits` and `bracketed_paste_limits`. Paste accumulation can now be bounded by wall-clock duration and/or byte count; on overflow, the inflight buffer is discarded and a new `Event::PasteAborted` is emitted. Both bounds default to `None`, so existing behaviour is unchanged unless the caller opts in. Fixes terminal hangs when the closing `\x1B[201~` marker never arrives (e.g. tmux pane swaps).
+
 # Version 0.29
 
 ## Added ⭐
--- a/src/event.rs
+++ b/src/event.rs
@@ -53,6 +53,10 @@
 //!             Event::Mouse(event) => println!("{:?}", event),
 //!             #[cfg(feature = "bracketed-paste")]
 //!             Event::Paste(data) => println!("{:?}", data),
+//!             #[cfg(feature = "bracketed-paste")]
+//!             Event::PasteAborted { reason, buffered_bytes } => {
+//!                 println!("PasteAborted: {:?}, {} bytes lost", reason, buffered_bytes.len())
+//!             }
 //!             Event::Resize(width, height) => println!("New size {}x{}", width, height),
 //!         }
 //!     }
@@ -99,6 +103,10 @@
 //!                 Event::Mouse(event) => println!("{:?}", event),
 //!                 #[cfg(feature = "bracketed-paste")]
 //!                 Event::Paste(data) => println!("Pasted {:?}", data),
+//!                 #[cfg(feature = "bracketed-paste")]
+//!                 Event::PasteAborted { reason, buffered_bytes } => {
+//!                     println!("PasteAborted: {:?}, {} bytes lost", reason, buffered_bytes.len())
+//!                 }
 //!                 Event::Resize(width, height) => println!("New size {}x{}", width, height),
 //!             }
 //!         } else {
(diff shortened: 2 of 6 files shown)


joshka:
I agree this is fixing a real problem, but I’m uneasy about locking in the public API shape here yet.

The parser wedge in #1060 seems real and worth addressing, but adding a new `Event` variant has broad downstream impact because `crossterm::event::Event` is an exhaustive public enum. That makes this feel bigger than an implementation-only fix.

I’d prefer to move the API/design discussion back to #1060 first and get feedback there on whether recovery should be surfaced as a new event at all, or whether we should aim for a non-breaking recovery design.

DannyStoll1:
> I agree this is fixing a real problem, but I’m uneasy about locking in the public API shape here yet.
> 
> The parser wedge in #1060 seems real and worth addressing, but adding a new `Event` variant has broad downstream impact because `crossterm::event::Event` is an exhaustive public enum. That makes this feel bigger than an implementation-only fix.
> 
> I’d prefer to move the API/design discussion back to #1060 first and get feedback there on whether recovery should be surfaced as a new event at all, or whether we should aim for a non-breaking recovery design.

Thanks for the careful review. Agreed that the contentious piece here is the `Event::PasteAborted` variant. I'll move the design discussion to #1060 and update this PR once the recovery surface is settled.

PR #1062, opened 2026-05-28
fix(cursor): write CSI 6n query to TTY when stdout is redirected

### LINKED ISSUE #652: Feature Request: A method for Unix systems to get the position of the cursor in stderr
I am trying to create a simple command line utility to draw a box with contents and questions in it which the user can directly input into. 

However, in an attempt to not let the box be redirected, I have decided to use eprintln!() to print out the box and its contents. This works fine when redirecting to some programs, such as cat, but on others, such as rev, crossterm::cursor::position() fails as the written characters used to find the cursor position are also redirected to the program. 

This effectively renders my utility is unusable or extremely tedious to use for scripting.

I would like to be able to get the position of the cursor in stderr, currently I have implemented a hackfix in my fork of crossterm [here](https://github.com/Vonr/crossterm/commit/ff22dec6e888257f15d127a3a29c7e16c5ee2fc4) where I changed line 32 of src/cursor/sys/unix.rs from
```rs
let mut stdout = io::stdout();
```
to
```rs
let mut stdout = io::stderr();
```

However, this is far from ideal as this breaks normal behavior and I am unable to come up with a solution that is cross-platform.

That said, I also have no intention of supporting Windows in my utility as I am deeply unfamiliar with their scripting.

I have created a fork of crossterm for personal use with a hackfix and used it in my Cargo.toml as a temporary workaround until a proper solution is implemented or found.

raphaelvigee:
## Summary

Fixes #652.

`cursor::position()` on Unix wrote the `ESC [ 6 n` Device Status Report query to stdout. When stdout is piped to another program (e.g. `prog | rev`), the escape sequence went into the pipe instead of the terminal, so the terminal never responded and the call hung until the 2s timeout. The terminal's reply also corrupted the piped output.

## Change

`write_cursor_query` picks the first available TTY for the query, in order:

1. stdout (preserved as default behavior when not redirected)
2. stderr
3. `/dev/tty`

The response is already read via `tty_fd()`, which falls back to `/dev/tty` when stdin is not a TTY, so the read side already worked — only the write side needed fixing.

`rustix::termios::isatty` is used for the TTY check (or `libc::isatty` under the `libc` feature). No new dependencies. MSRV 1.63 preserved (avoided `std::io::IsTerminal`, stable in 1.70).

## Test plan

- [x] `cargo build` (default features)
- [x] `cargo build --features libc`
- [x] `cargo build --no-default-features --features events`
- [ ] Manual: run a small binary that calls `crossterm::cursor::position()` with stdout piped to `rev` — confirm the call returns the cursor position instead of corrupting the pipe / timing out.
- [ ] Manual: same binary run normally in a terminal — confirm behavior unchanged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Unreleased
 
+## Fixed 🐛
+
+- `cursor::position()` on Unix now writes the CSI 6n query to stderr or `/dev/tty` when stdout is redirected, so it no longer corrupts piped output and works when stdout is not a TTY (#652)
+
 # Version 0.29
 
 ## Added ⭐
--- a/src/cursor/sys/unix.rs
+++ b/src/cursor/sys/unix.rs
@@ -1,4 +1,5 @@
 use std::{
+    fs::OpenOptions,
     io::{self, Error, ErrorKind, Write},
     time::Duration,
 };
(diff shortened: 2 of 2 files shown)


raphaelvigee:
I'm currently using the version 0.29 of the project, if this PR is accepted in principle i will update it to master

PR #1064, opened 2026-06-12
Support for ConEmu progress bar (OSC 9;4)

### LINKED ISSUE #1063: ConEmu progress bar (OSC 9;4) support
Support ConEmu progress bar, also known as OSC 9;4. Here is a documentation: https://learn.microsoft.com/en-us/windows/terminal/tutorials/progress-bar-sequences.

sanxiyn:
Tested on [Ghostty](https://ghostty.org/).

Fix #1063

### DIFF
--- a/examples/progress.rs
+++ b/examples/progress.rs
@@ -1,0 +1,19 @@
+//! Demonstrates ConEmu progress bar (OSC 9;4) support.
+//!
+//! cargo run --example progress
+
+use std::io;
+use std::thread;
+use std::time::Duration;
+
+use crossterm::{execute, terminal::SetProgress};
+
+fn main() -> io::Result<()> {
+    let mut out = io::stdout();
+    for i in 0..=10 {
+        let progress = i * 10;
+        execute!(out, SetProgress::Default(progress))?;
+        thread::sleep(Duration::from_secs(1));
+    }
+    execute!(out, SetProgress::Clear)
+}
--- a/src/terminal.rs
+++ b/src/terminal.rs
@@ -358,6 +358,41 @@ impl Command for Clear {
     }
 }
 
+/// A command that sets the progress
+///
+/// # Notes
+///
+/// Command must be executed/queued for execution otherwise they do nothing.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum SetProgress {
+    Clear,
+    Indeterminate,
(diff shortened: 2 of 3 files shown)


examples/progress.rs
```diff
@@ -0,0 +1,19 @@
...
+use std::thread;
+use std::time::Duration;
+
+use crossterm::{execute, terminal::SetProgress};
+
+fn main() -> io::Result<()> {
+    let mut out = io::stdout();
+    for i in 0..=10 {
+        let progress = i * 10;
+        execute!(out, SetProgress::Default(progress))?;
```
cammeresi, line 15: I would suggest that the example show all four active modes because running this example is probably 50% of the review.

src/terminal.rs
```diff
@@ -358,6 +358,41 @@ impl Command for Clear {
     }
 }
 
+/// A command that sets the progress
+///
+/// # Notes
+///
+/// Command must be executed/queued for execution otherwise they do nothing.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum SetProgress {
```
cammeresi, line 367: The variants don't seem to fully correspond to the specification (is [this one](https://conemu.github.io/en/AnsiEscapeCodes.html#ConEmu_specific_OSC) the right place to look?), and I'm not a fan of some of the names.

`Default` is IMO a poor choice because Rust has kind of taken over the word "default."  I'd suggest `Set`, `Value`, or `Info`.

`Warning` is called `Paused` in the spec.  What's wrong with `Paused`?

Also the spec says that in paused or error states, the progress number is optional, which is not allowed here.  I would make those two take `Option<u8>`.

cammeresi (changes requested):
What the example demonstrates worked in Ghostty and iTerm2, but IMO it should demonstrate everything.

src/terminal.rs
sanxiyn: The specification I am following is [Microsoft's](https://learn.microsoft.com/en-us/windows/terminal/tutorials/progress-bar-sequences), documenting Windows Terminal behavior. Variant names correspond to Microsoft's specification. Ditto whether the progress is optional.

examples/progress.rs
sanxiyn: Okay.

PR #1065, opened 2026-07-08
docs: fix typo "corresponsing" -> "corresponding" in clipboard.rs

maxtaran2010:
Fix a small typo in a doc comment in `src/clipboard.rs`.

Line 73: `corresponsing` -> `corresponding`

```rust
// Before
/// Returns a String corresponsing to the "Pc" parameter of the OSC52

// After
/// Returns a String corresponding to the "Pc" parameter of the OSC52
```


joshka:
thanks

PR #1070, opened 2026-07-20
fix: propagate EventStream poll errors

alurm:
## Summary

- Stop `EventStream`'s blocking worker from retrying indefinitely after a polling error.
- Preserve the error and wake the async consumer so that the next `poll_next` returns it.
- Suppress polling errors caused by waking the worker during normal `EventStream` shutdown.
- Add an entry to the changelog.

## Motivation

`EventStream` already exposes `io::Result<Event>`, and errors from its immediate non-blocking poll are returned to the consumer. However, errors from the blocking poll on its background thread were ignored and retried indefinitely.

That could leave the worker spinning after a terminal failure instead of reporting the error to the stream consumer.

## Related work

This is independent of #1067, which makes the Unix event source report an error when its TTY disappears. The two changes are complementary: #1067 detects the disconnected TTY, while this PR propagates that error through `EventStream`.

## Testing

The tests cover:

- not starting another poll after shutdown;
- suppressing an error when shutdown occurs during a poll;
- polling only once after a real error, storing that error, and waking the stream consumer.

Commands run:

```console
cargo fmt --all -- --check
cargo clippy --all-features -- -D clippy::all
cargo test --all-features
cargo test --no-default-features --features "use-dev-tty events event-stream bracketed-paste"
```

---

Assisted by Codex:gpt-5.6-sol.

### DIFF
--- a/src/event/stream.rs
+++ b/src/event/stream.rs
@@ -4,7 +4,7 @@ use std::{
     sync::{
         atomic::{AtomicBool, Ordering},
         mpsc::{self, SyncSender},
-        Arc,
+        Arc, Mutex,
     },
     task::{Context, Poll},
     thread,
@@ -36,6 +36,7 @@ pub struct EventStream {
     poll_internal_waker: Waker,
     stream_wake_task_executed: Arc<AtomicBool>,
     stream_wake_task_should_shutdown: Arc<AtomicBool>,
+    stream_wake_task_error: Arc<Mutex<Option<io::Error>>>,
     task_sender: SyncSender<Task>,
 }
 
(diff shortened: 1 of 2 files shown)


src/event/stream.rs
```diff
@@ -45,30 +46,48 @@ impl Default for EventStream {
...
-                    }
-
-                    if task.stream_wake_task_should_shutdown.load(Ordering::SeqCst) {
-                        break;
-                    }
-                }
-                task.stream_wake_task_executed
-                    .store(false, Ordering::SeqCst);
-                task.stream_waker.wake();
+                run_stream_wake_task(task, || internal::poll(None, &EventFilter));
```
cammeresi, line 49: In C, the language greatly encouraged callees to be above callers, so you could generally read from the bottom up.  Rust doesn't have that restriction, so some people order things top down (which I don't like, but I'm letting that go).

LLMs just do not care about human readers at all and so will spray out code in all directions, which can be hard for a human to read.  I think people should pick an order and try to force the LLM to stick to it.  If A calls B calls C, the reviewer should see an order that makes sense (A-B-C or C-B-A, not A-C-B).

(This is not an anti-LLM rant.)

PR #1073, opened 2026-07-25
fix(windows): don't feed surrogate key-up halves into the pairing

### LINKED ISSUE #1072: Windows: emoji panel input (KEYEVENTF_UNICODE) is silently dropped — surrogate key-up halves poison the pairing
## Summary

Characters inserted via the Windows emoji panel (Win+.) — or any injector using `SendInput` with `KEYEVENTF_UNICODE` — never come out of crossterm's event stream on Windows. BMP characters survive (single code unit), but anything needing a surrogate pair (emoji, many symbols) is dropped silently: no event, no error.

## Environment

crossterm 0.29.0, Windows 11, Windows Terminal + ConPTY, raw-mode TUI reading via `event::poll`/`event::read`.

## Evidence the console side is fine

A raw `ReadConsoleInputW` probe (bypassing crossterm entirely) shows the injection arriving cleanly — each UTF-16 code unit is delivered as a key-down **and** a key-up `KEY_EVENT`, with `wVirtualKeyCode = 0`:

```
KEY down=1 vk=0x00 unit=U+D83D (HIGH-SURROGATE)
KEY down=0 vk=0x00 unit=U+D83D (HIGH-SURROGATE)
KEY down=1 vk=0x00 unit=U+DE0A (LOW-SURROGATE)
KEY down=0 vk=0x00 unit=U+DE0A (LOW-SURROGATE)
```

(U+1F60A, 😊.)

## Mechanism

`parse_key_event_record` (src/event/sys/windows/parse.rs) forwards any `u_char` in `0xD800..=0xDFFF` as `WindowsKeyEvent::Surrogate` for key-down and key-up records alike, and `handle_surrogate` pairs consecutive values blindly. With the down+up interleaving above:

1. `D83D` (down) → buffered
2. `D83D` (up) → paired as (high, high) → `decode_utf16` error → discarded
3. `DE0A` (down) → buffered
4. `DE0A` (up) → paired as (low, low) → error → discarded

Every half is consumed; nothing is emitted.

## Fix

Let only key-down records carry a surrogate into the pairing, and drop key-up halves without disturbing a buffered value — the down halves then pair correctly. One guarded early-return; PR follows.

Verified end-to-end with the guard in place: emoji-panel input composes correctly and reaches a PTY child through our TUI (dogfooded via `[patch.crates-io]` in https://github.com/kt81/banto).

(The `VK_MENU` alt-code path also emits surrogates on key-up by design — that path is deliberately untouched; this is only about the general branch.)

kt81:
Fixes #1072.

The Windows emoji panel (and any `KEYEVENTF_UNICODE` injector) delivers each UTF-16 code unit of a non-BMP character as a key-down **and** a key-up record. `parse_key_event_record` forwarded the surrogate range on both, so `handle_surrogate` paired (high, high) then (low, low) — both decode errors — and the character vanished without a trace. See the issue for the raw `ReadConsoleInputW` capture demonstrating the down+up interleaving (`vk=0`).

This lets only key-down records carry a surrogate into the pairing; a key-up half returns `None` before reaching `handle_key_event`'s buffer handling, so a buffered down-half survives it and pairs with the next down-half.

Notes for review:

- The `VK_MENU` alt-code branch above also emits surrogates on key-up — **by design** (alt codes are keyup-driven) — and is deliberately untouched.
- No unit test accompanies this: `KeyEventRecord` cannot be constructed outside `crossterm_winapi` (`ControlKeyState`'s field is private), which is presumably why this file has no existing tests. Verified instead against the raw capture in the issue and end-to-end in a TUI that forwards the composed character to a PTY child (dogfooded via `[patch.crates-io]` in kt81/banto).

Happy to adjust if there's an established way to test this path.

### DIFF
--- a/src/event/sys/windows/parse.rs
+++ b/src/event/sys/windows/parse.rs
@@ -269,6 +269,17 @@ fn parse_key_event_record(key_event: &KeyEventRecord) -> Option<WindowsKeyEvent>
                     get_char_for_key(key_event).map(KeyCode::Char)
                 }
                 surrogate @ 0xD800..=0xDFFF => {
+                    if !key_event.key_down {
+                        // The Windows emoji panel (and other KEYEVENTF_UNICODE
+                        // injectors) deliver each UTF-16 code unit as a key-down
+                        // *and* a key-up record. Feeding the key-up half into the
+                        // surrogate pairing produces broken (high, high) /
+                        // (low, low) pairs that decode to nothing, so the whole
+                        // character is silently lost. Only the key-down half
+                        // carries the surrogate forward; the key-up half is
+                        // dropped here *without* disturbing a buffered value.
+                        return None;
+                    }
                     return Some(WindowsKeyEvent::Surrogate(surrogate));
                 }
                 unicode_scalar_value => {


kt81:
For reviewers: two earlier PRs attack the same bug — #857 (2023) and #1009 (2025). The difference here is scope and side: this keeps the surrogate pairing exactly as-is and only stops **key-up** halves from entering it (the alt-code path, which is key-up-driven by design, is untouched), so key-down remains the single source of characters, consistent with how every other key already behaves. Evidence chain in #1072.

PR #1074, opened 2026-07-25
feat(event): report the kitty protocol's base layout key

### LINKED ISSUE #968: Incomplete kitty progressive enhancement support and missing "base-layout-key" support
## Description

The current implementation provides only one `KeyEvent::code`.
However, Kitty offers four: `unicode-key-code`, `shifted-key`, `base-layout-key` and `text-as-codepoints`.
Currently `KeyEvent::code` is set to either `shifted-key` or `unicode-key-code`.

https://sw.kovidgoyal.net/kitty/keyboard-protocol/#key-codes
> If alternate key reporting is requested by the program running in the terminal, the terminal can send two additional Unicode codepoints, the shifted key and base layout key, separated by colons.
> The base layout key is the key corresponding to the physical key in the standard PC-101 key layout.

## Why is it necessary

Users who work with non-Latin languages often need to switch between multiple keyboard layouts. Without support for `base-layout-key`, users cannot maintain consistent shortcuts when switching layouts because the current code relies on the character's Unicode value, which varies per layout.

Creating bindings for all layouts is not a universal solution, as certain characters may appear in different positions across various layouts. This issue is actual not only for software applications, but for games as well. For instance, Shift+W or Shift+Ц should be recognized as "Shift+w" rather than "W" or "Ц". And when users switch to chat mode, game ideally shouldn't have to toggle `REPORT_ALTERNATE_KEYS` to receive text representation.

## Current implementation

Here is the current implementation using simplified pseudocode:

```rust
// Full implementation in src/event/sys/unix/parse.rs fn parse_csi_u_encoded_key_code

parse!( CSI ( unicode_key ':' shifted_key ':' base_layout_key ) ';' ( modifiers ':' event_type ) ';' ( text_as_codepoints ) 'u' );

let mut code = parse(unicode_key);
let modifiers = parse(modifiers);
let event_type = parse(event_type);

if let Some(shifted_key) = parse(shifted_key) && modifiers.contains(SHIFT) {
    let mut code = shifted_key;
    modifiers.remove(SHIFT);
}

ignore(base_layout_key);
ignore(text_as_codepoints);

return KeyEvent { code, modifiers, event_type };
```

Current implementation replaces key by its shifted key if given and removes SHIFT modifier. Even if it maybe okay for processing input as characters it leads to some inconsistency by implicitly removing SHIFT when input should be processed as scancodes.

## Kitty keys example

```
$ kitten show-key -m kitty

// Release events are omitted for example simplicity
// ЙЦУКЕН - main Cyrillic keyboard layout
// Key "qQ" on QWERTY layout or "йЙ" on ЙЦУКЕН layout
CSI 113 ; ; 113 u                   // "q" on QWERTY layout
CSI 113 : 81 ; 2 ; 113 u            // "Q" on QWERTY layout + Shift
CSI 1081 : : 133 ; ; 1081 u         // "й" on ЙЦУКЕН layout
CSI 1081 : 1049 : 113 ; 2 ; 1049 u  // "Й" on ЙЦУКЕН layout + Shift

// Key ".>" on QWERTY layout or "юЮ" on ЙЦУКЕН layout
CSI 46 ; ; 46 u                     // "." on QWERTY layout
CSI 1102 : : 46 ; ; 1102 u          // "ю" on ЙЦУКЕН layout

// Key "/?" on QWERTY layout or "." on ЙЦУКЕН layout
CSI 47 ; ; 47 u                     // "/" on QWERTY layout
CSI 46 : : 47 ; ; 46 u              // "." on ЙЦУКЕН layout

// Key "2@" on QWERTY layout or "2\"" on ЙЦУКЕН layout
CSI 50 ; ; 50 u                     // "2" on QWERTY layout
CSI 50 : 64 ; 2 ; 64 u              // "@" on QWERTY layout + Shift
CSI 50 ; ; 50 u                     // "2" on ЙЦУКЕН layout
CSI 50 : 34 ; 2 ; 34 u              // '"' on ЙЦУКЕН layout + Shift

// Key "3#" on QWERTY layout or "3№" on ЙЦУКЕН layout
CSI 51 ; ; 51 u                     // "3" on QWERTY layout
CSI 51 : 35 ; 2 ; 35 u              // "#" on QWERTY layout + Shift
CSI 51 ; ; 51 u                     // "3" on ЙЦУКЕН layout
CSI 51 : 8470 ; 2 ; 8470 u          // "№" on ЙЦУКЕН layout + Shift
```

Note that character `'.'` uses different keys in QWERTY and ЙЦУКЕН layout, so it is impossible to add a character-based shortcut that will work on any layout for same key and not conflict with other shortcuts.
When switching layouts often it is convenient that shortcuts like "Ctrl+C" or "Ctrl+." work depending on their scancode, not character that depend on layout.
If we use `base_layout_key.unwrap_or(unicode_key_code)` for shortcuts we'll be able to use same keys at any layout. 

## Expected features

Ideally, it should be possible to obtain all: "unicode-key-code", "shifted-key" and "base-layout-key" from separate `KeyEvent` fields.
Applications then can use "base-layout-key" is provided or "unicode-key-code" for shortcuts.

Also according to https://sw.kovidgoyal.net/kitty/keyboard-protocol/#text-as-code-points using "text-as-codepoints" for character input may be more correct/accurate than using "shifted-key".

Anyway ultimately, the decision about what code to use is better made at the application level, rather than at the library level.

## Possible solutions

Adding more fields to `KeyEvent` would be a significant breaking change requiring extensive modifications from developers using the library. It's challenging to implement this without such changes. Besides, this may look too protocol-specific. However, if other breaking changes are planned, it might be feasible to introduce these enhancements concurrently with those updates.
```rust
pub struct KeyEvent {
    pub code: KeyCode, // ideally should be just "unicode-key-code", but this may break backward-compatibility for apps that already rely on current behavior.
    pub chars: Vec<char>, // or better optimized container for text-as-codepoints
    pub shifted_key: Option<KeyCode>, // for "shifted-key"
    pub base_layout_code: Option<KeyCode>, // from "base-layout-key"
    pub modifiers: KeyModifiers,
    pub kind: KeyEventKind,
    pub state: KeyEventState,
}
```

A potentially better approach might involve introducing feature-gated support for more low-level, platform-dependent events. Instead of emitting a `KeyEvent` and other kitty protocol events, the library could emit a minimally parsed lower-level (with no information lost) "KittyEvent" or "RawTerminalEvent".

## Links

https://github.com/crossterm-rs/crossterm/pull/688 - adds `kind` field .
https://github.com/crossterm-rs/crossterm/pull/691 - adds fn keys and `state` field.
https://github.com/crossterm-rs/crossterm/pull/696 - adds more modifiers.
https://github.com/crossterm-rs/crossterm/pull/754 - adds `REPORT_ALTERNATE_KEYS`.

vshylov:
Closes #968.

## Problem

The kitty keyboard protocol's *report alternate keys* enhancement (`0b100`) sends up to two extra codepoints with each key: the shifted key and the **base layout key** — "the key corresponding to the physical key in the standard PC-101 key layout". The parser reads the shifted one (and only when shift is held), then drops the rest.

Without the base layout key an application has no way to match a shortcut by physical key. Under a Cyrillic, Greek, Hebrew or Georgian layout the terminal faithfully reports the layout's character, so every `Ctrl+<letter>` binding is dead — which is precisely what the enhancement exists to solve; the spec's own example is the Cyrillic `ctrl+С` → `ctrl+c` case.

## Change

* Parse both alternates positionally. Besides making the base layout key reachable at all, this fixes a latent bug: any alternate may be empty (`CSI 1076::108;5u` reports a base layout key and no shifted key), and the old `codepoints.next()` — reached only under shift — would then hand out the wrong slot.
* Expose the base layout key as `KeyEvent::base_layout_code: Option<KeyCode>`, alongside the existing enhancement-gated `kind`/`state` fields, plus a `with_base_layout_code` builder.

## The one decision worth your review

`base_layout_code` deliberately takes **no part in `PartialEq`/`Hash`**.

It describes the same key press rather than identifying it, and including it would silently break the ubiquitous

```rust
if event == KeyEvent::new(KeyCode::Char('c'), KeyModifiers::CONTROL) { … }
```

comparison for every application the moment it enables `REPORT_ALTERNATE_KEYS` — which would make the feature unusable in practice. The manual `PartialEq`/`Hash` impls destructure `KeyEvent` exhaustively, so this is an explicit `base_layout_code: _`, not an omission.

Derived `PartialOrd`/`Ord` *do* include the new field, which is inconsistent with `Eq` — but they already disagree with it today via `normalize_case`, so I left that pre-existing wart alone rather than widen the diff. Happy to align them if you would prefer.

**Scope:** I did not expose the shifted key. It is already reflected in `code` by the existing substitution, so a separate field would be redundant. #968 asks for both — say the word and I will add it.

## Compatibility

Breaking only for struct-literal construction and exhaustive destructuring of `KeyEvent`; the `new*` constructors, equality and hashing are unaffected, and `serde(default)` keeps older payloads deserializable. **No existing test needed updating** — the suite passes unchanged, which is the concrete form of that claim.

## Tests

Two new tests in the `CSI u` parser: alternates parsed (with a shifted key, without one, and with no alternates at all), and the equality/hash invariant above.

`cargo fmt --check`, `cargo clippy --all-features -- -D clippy::all` and `cargo test --all-features` are green on Linux (Ubuntu 24.04, 119 lib tests) and Windows.

One note for adopters, learned from wiring this into a TUI end to end: the base layout key should not be honoured for ASCII characters. On AZERTY the key labelled `A` sits at the US `Q` position, so it arrives as `Char('a')` with a base layout key of `q` — taking it unconditionally turns `Ctrl+A` into `Ctrl+Q`. Matching by physical key is what you want for non-ASCII layouts; for Latin ones, the label is.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,22 @@
 # Unreleased
 
+## Added ⭐
+
+- Report the Kitty keyboard protocol's *base layout key* as
+  `KeyEvent::base_layout_code`, so shortcuts can be matched by physical key
+  regardless of the active keyboard layout (#968). Set only when
+  `KeyboardEnhancementFlags::REPORT_ALTERNATE_KEYS` is enabled.
+
 ## Breaking ⚠️
 
 - Raise the minimum supported Rust version from 1.63 to 1.85.
 - Remove `IsTty` trait.
   Use the standard library's [`std::io::IsTerminal`](https://doc.rust-lang.org/std/io/trait.IsTerminal.html) trait instead,
   which provides equivalent functionality.
+- `KeyEvent` gained a `base_layout_code` field: struct-literal construction and
+  exhaustive destructuring need updating. The `KeyEvent::new*` constructors,
+  `PartialEq` and `Hash` are unaffected — the new field takes no part in
+  equality, so comparisons against manually built events keep working.
 
 ## Changed ⚙️
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,19 @@
 # Unreleased
 
+## Added ⭐
+
+- Report the Kitty keyboard protocol's *base layout key* as
+  `KeyEvent::base_layout_code`, so shortcuts can be matched by physical key
+  regardless of the active keyboard layout (#968). Set only when
+  `KeyboardEnhancementFlags::REPORT_ALTERNATE_KEYS` is enabled.
+
+## Breaking ⚠️
+
+- `KeyEvent` gained a `base_layout_code` field: struct-literal construction and
+  exhaustive destructuring need updating. The `KeyEvent::new*` constructors,
+  `PartialEq` and `Hash` are unaffected — the new field takes no part in
+  equality, so comparisons against manually built events keep working.
+
 ## Fixed 🐛
 
 - Fix integer underflow in mouse / cursor-position parsers when coord
--- a/src/event.rs
+++ b/src/event.rs
@@ -933,6 +933,24 @@ pub struct KeyEvent {
     /// Only set if [`KeyboardEnhancementFlags::DISAMBIGUATE_ESCAPE_CODES`] has been enabled with
     /// [`PushKeyboardEnhancementFlags`].
     pub state: KeyEventState,
+    /// The key at the same physical position on a standard PC-101 keyboard,
+    /// independent of the active keyboard layout.
+    ///
+    /// This is what makes layout-independent shortcuts possible: with a Cyrillic
+    /// layout active, the physical `C` key reports `KeyCode::Char('с')` in
+    /// [`code`](Self::code) and `Some(KeyCode::Char('c'))` here, so an
+    /// application can match `Ctrl+C` regardless of the layout the user types in.
+    ///
+    /// Only set if [`KeyboardEnhancementFlags::REPORT_ALTERNATE_KEYS`] has been
+    /// enabled with [`PushKeyboardEnhancementFlags`], and only on Unix — the
+    /// Windows console API does not report it.
+    ///
+    /// This field deliberately takes no part in [`PartialEq`]/[`Hash`], so that
+    /// comparing against a manually constructed event — the common
+    /// `event == KeyEvent::new(KeyCode::Char('c'), KeyModifiers::CONTROL)`
+    /// pattern — keeps working once the enhancement is enabled.
+    #[cfg_attr(feature = "serde", serde(default))]
+    pub base_layout_code: Option<KeyCode>,
 }
 
 impl KeyEvent {
@@ -942,6 +960,7 @@ impl KeyEvent {
             modifiers,
             kind: KeyEventKind::Press,
             state: KeyEventState::empty(),
+            base_layout_code: None,
         }
     }
 
@@ -955,6 +974,7 @@ impl KeyEvent {
             modifiers,
             kind,
             state: KeyEventState::empty(),
+            base_layout_code: None,
         }
     }
 
@@ -969,7 +989,15 @@ impl KeyEvent {
             modifiers,
             kind,
             state,
+            base_layout_code: None,
         }
+    }
+
+    /// Returns the event with [`base_layout_code`](Self::base_layout_code) set.
+    #[must_use = "this returns a new event instead of modifying the original"]
+    pub fn with_base_layout_code(mut self, base_layout_code: Option<KeyCode>) -> KeyEvent {
+        self.base_layout_code = base_layout_code;
+        self
     }
 
     // modifies the KeyEvent,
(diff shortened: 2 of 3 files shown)


vshylov:
Rebased onto current `master` — the edition 2024 migration and the MSRV bump both landed after this was opened. Only `CHANGELOG.md` conflicted (`src/event.rs` and the `CSI u` parser auto-merged); I folded the `KeyEvent` note into the existing `Breaking` section rather than adding a second one.

Re-ran the current CI set locally on Ubuntu 24.04, all green:

- `cargo clippy --locked --all-targets --all-features -- -D warnings`
- `cargo doc --locked --all-features --no-deps` with `RUSTDOCFLAGS=-D warnings`
- `cargo test --locked --all-targets --all-features -- --test-threads 1` — 123 passed
- `cargo test --locked --doc --all-features` — 51 + 2 passed
- `cargo check --locked --lib` on 1.85.0, both `--no-default-features` and `--all-features`

(The lone `method read is never used` warning in the MSRV run reproduces on plain `master`, so it is not from this change.)

The workflow run is sitting in `action_required` — could someone approve it when you get the chance?

Also, re #1102: nothing here needs more than the 1.63 baseline this was originally written against, so lowering the MSRV would not affect this PR either way.

Issue #1075, opened 2026-07-29
Tracking: modernize and harden Crossterm CI


joshka:
## Summary

This issue is the plan of record for bringing Crossterm’s CI up to a level that can safely support required checks and, later, GitHub-based releases. It is related to the release discussion in [#1023](https://github.com/crossterm-rs/crossterm/issues/1023), but publishing and trusted publishing remain separate follow-up work.

The work will land as small PRs. Each PR should link this issue, and this checklist will be updated as work lands.

### 1. Harden the existing workflow

- [x] Rename the existing workflow to `ci.yml` while preserving approximately its current coverage.
- [x] Pin `actions/checkout` to `3d3c42e5aac5ba805825da76410c181273ba90b1` (`v7.0.1`) with `persist-credentials: false`.
- [x] Replace `hecrj/setup-rust-action@master` with `dtolnay/rust-toolchain@2c7215f132e9ebf062739d9130488b56d53c060c`, a durable commit from the action’s `master` history.
- [x] Set explicit `contents: read` permissions and grant no write permissions or secrets.
- [x] Add concurrency cancellation, job timeouts, `fail-fast: false`, deliberate runner versions, and `workflow_dispatch`.
- [x] Add grouped monthly Dependabot updates for pinned GitHub Actions.
- [x] Validate the workflow with Actionlint and Zizmor.

Completed in [#1076](https://github.com/crossterm-rs/crossterm/pull/1076).

The initial Rust installer pin used a generated toolchain-branch commit. [#1083](https://github.com/crossterm-rs/crossterm/pull/1083) corrects it to `master` history because the workflows supply the `toolchain` input explicitly and unreachable generated commits may eventually be garbage-collected.

This PR changed workflow execution and security, not Rust policy or test coverage.

### 2. Establish the lockfile baseline

- [x] Begin tracking the repository-root `Cargo.lock` generated from the existing manifests.
- [x] Establish a resolution baseline that does not contain the identified vulnerable or yanked `bytes 1.6.0` and `futures-util 0.3.30` releases.
- [x] Confirm that obsolete transitive dependencies observed during the audit, such as `adler`, are absent from the baseline.
- [x] Keep manifest requirements unchanged unless a newer minimum is genuinely required.
- [x] Include a dependency-tree summary and run the full existing test suite.

Completed in [#1078](https://github.com/crossterm-rs/crossterm/pull/1078).

[...]

PR #1082, opened 2026-07-29
Update edition to 2024 in interactive-demo

xtqqczze:
Contributes to https://github.com/crossterm-rs/crossterm/issues/1075

### DIFF
--- a/examples/interactive-demo/src/macros.rs
+++ b/examples/interactive-demo/src/macros.rs
@@ -6,7 +6,7 @@ macro_rules! run_tests {
         ),*
         $(,)?
     ) => {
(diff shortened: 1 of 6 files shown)


examples/interactive-demo/src/macros.rs
```diff
@@ -1,6 +1,6 @@
 macro_rules! run_tests {
     (
-        $dst:expr,
+        $dst:expr_2021,
```
joshka: ditch this - it's usually unnecessary otherwise evaluate it and state what the problem is that it solves

joshka (approved):

joshka:
Thanks

PR #1083, opened 2026-07-29
Establish the required CI gate

joshka:
## Summary

- split stable validation into focused format, Clippy, documentation, test, doctest, MSRV, feature, and package jobs
- add one stable `ci-success` aggregate intended for branch protection
- restore tests for the selected default, Serde, event-stream, minimal Unix, and supported Windows feature configurations
- add an advisory beta-Clippy check so incoming stable lints can be handled before they break every pull request
- move beta and nightly test compatibility checks to a weekly and manually dispatched workflow
- document the local commands and the design constraints behind the CI policy

This is the next implementation slice from #1075. It does not add caching, secrets, release permissions, or publishing authority.

## Coverage

Stable tests run with all targets and features on Linux, macOS, and Windows. Tests remain single-threaded because Crossterm has tests that manipulate process-global terminal and environment state.

The workflow also checks:

- Rust 1.85 with no default features and all public features
- default, Serde, and event-stream unit tests on Unix
- the selected minimal Unix feature configurations
- the supported `windows` and `windows,events` configurations on Windows
- the publishable package on Linux and Windows

A bare Windows `--no-default-features` build is not included as a passing check: Crossterm deliberately rejects Windows builds without the `windows` feature in `src/lib.rs`. Testing the two supported configurations records the actual invariant without making an expected failure part of the required gate.

## CI design

Required checks use stable Rust and cover deterministic properties directly relevant to a releasable crate. Cross-platform all-feature tests catch operating-system differences, while the smaller feature matrix tests deliberate compatibility points on representative platforms to keep feedback timely.

This is not an exhaustive feature powerset. [`cargo-hack`](https://github.com/taiki-e/cargo-hack) supports `--each-feature` and `--feature-powerset`, but adopting it would mean choosing a broader feature-compatibility policy, paying the additional runtime, and maintaining another pinned tool. That is better considered with the auxiliary-tool integrity work in the next tracking-issue slice rather than introduced implicitly here.

Stable Clippy is required. Beta Clippy also runs on pull requests with `continue-on-error`: it is an early warning for lints likely to arrive in the next stable toolchain, but an upstream beta change must not block merges. The job is deliberately excluded from `ci-success`, labels itself advisory, and emits a warning explaining failures. This follows the motivation in [Clippy's CI guidance](https://doc.rust-lang.org/clippy/continuous_integration/index.html), while choosing beta over nightly to focus on the next stable release with less churn.

Beta and nightly test compatibility remain in a separate weekly and manually dispatched workflow. They are visible signals without becoming ordinary pull-request gates. The weekly schedule runs away from the top of the hour because GitHub notes that scheduled workflows may be delayed during high-load periods.

Caching is intentionally deferred until job timings show that its additional state and maintenance are worthwhile.

## Branch-protection design

`ci-success` uses `always()` so it still reports a definitive failure when one of its dependencies fails instead of becoming a skipped required check, as described in [GitHub's required-check troubleshooting documentation](https://docs.github.com/en/pull-requests/how-tos/merge-and-close-pull-requests/troubleshooting-required-status-checks).

The aggregate has a unique, stable name so branch protection can require it without depending on every matrix job name. GitHub recommends unique job names when required checks may be produced by multiple workflows: [About protected branches](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-protected-branches/about-protected-branches).

Pull-request workflows are read-only, persist no checkout credentials, and have no secrets or publishing authority.

Third-party actions are pinned to full commit SHAs. The Rust installer uses a commit from its durable `master` history because the action requires that pin shape when workflows provide the `toolchain` input explicitly.

## Validation

- `actionlint` 1.7.12
- `zizmor` 1.26.1
- formatting, strict stable and beta Clippy, Rustdoc warnings, doctests, and the all-target/all-feature test suite
- Rust 1.85 checks
- every selected Unix feature-test configuration
- package verification

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,24 +9,157 @@ on:
       - master
   workflow_dispatch:
 
-# Keep validation read-only. Publishing will use a separate workflow with narrowly scoped permissions.
+# Keep pull request validation read-only. Publishing belongs in a separate workflow.
 # https://docs.github.com/en/actions/tutorials/authenticate-with-github_token
 permissions:
   contents: read
 
+# Actions use full commit SHAs. The rust-toolchain pin comes from master history because the action
+# requires a durable master commit when the `toolchain` input is supplied explicitly.
+# https://github.com/dtolnay/rust-toolchain#choice-of-full-length-commit-sha
+
 # A pull request keeps the same number as commits are added, so newer commits cancel stale runs.
-# Push and manual runs fall back to the ref. The workflow name prevents cross-workflow cancellation.
+# Push and manual runs fall back to the ref.
 # https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-workflow-concurrency
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
   cancel-in-progress: true
 
 jobs:
+  format:
+    name: format
+    runs-on: ubuntu-24.04
+    timeout-minutes: 10
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
+
+      - name: Install stable Rust
+        uses: dtolnay/rust-toolchain@2c7215f132e9ebf062739d9130488b56d53c060c
+        with:
+          toolchain: stable
+          components: rustfmt
+
+      - name: Check formatting
+        run: cargo fmt --all -- --check
+
+  clippy:
+    name: clippy
+    runs-on: ubuntu-24.04
+    timeout-minutes: 15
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -9,24 +9,153 @@ on:
...
+          persist-credentials: false
+
+      - name: Install stable Rust
+        uses: dtolnay/rust-toolchain@4cda84d5c5c54efe2404f9d843567869ab1699d4
+        with:
+          toolchain: stable
+          components: clippy
+
+      - name: Run Clippy
+        run: cargo clippy --locked --all-targets --all-features -- -D warnings
```
xtqqczze, line 65:
```suggestion
        run: cargo clippy --locked --workspace --all-targets --all-features -- -D warnings
```

etc

PR #1085, opened 2026-07-29
Add local CI command shortcuts

joshka:
## Summary

- add a cross-platform `justfile` with named recipes for Crossterm's local checks
- make `just ci` run the required checks and the deliberate Unix or Windows feature configurations
- keep beta Clippy available as a separate advisory recipe
- document how to install Just and discover the available recipes

## Why

The CI suite now has enough distinct commands that copying them from the workflow or contribution guide is error-prone. These recipes provide memorable entry points without replacing the GitHub Actions workflows as the source of truth for required policy.

The `features` recipe uses Just's operating-system attributes so Unix and Windows contributors run the feature configurations supported on their platform. Individual recipes remain available when a contributor only needs to repeat one check.

The local `package` recipe passes `--allow-dirty` because its purpose is to validate package contents while changes are still under review. GitHub CI continues to run `cargo package --locked` without that option and therefore retains the VCS-cleanliness check.

This is the Justfile follow-up recorded in [#1075](https://github.com/crossterm-rs/crossterm/issues/1075).

## Validation

- `just --fmt --check`
- `just --list`
- `just --dry-run ci`
- `just ci` on macOS

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -67,6 +67,11 @@ Search for `#![deny(...)]` in the code:
 
 ## Local checks
 
+The [`justfile`](../justfile) provides memorable shortcuts for the local checks. Install
+[`just`](https://just.systems/man/en/packages.html), then run `just` to list the available recipes
+or `just ci` to run the required checks for the current operating system. The GitHub Actions
+workflows remain the source of truth for required CI policy.
+
 Run the required checks before opening a pull request:
 
 ```sh
--- a/justfile
+++ b/justfile
@@ -1,0 +1,76 @@
+# These recipes are shortcuts for local development. The GitHub Actions workflows remain the
+# source of truth for required CI policy.
+
+manifest := read("Cargo.toml")
+
+raw_msrv := replace_regex(manifest, '(?ms)\A.*?^[ \t]*rust-version[ \t]*=[ \t]*"([^"]+)".*\z', '$1')
+
+msrv_version := if raw_msrv == manifest { error("Cargo.toml must define rust-version") } else { raw_msrv }
+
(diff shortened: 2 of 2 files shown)


xtqqczze:
Could we avoid hardcoding the MSRV here and instead read it from Cargo.toml?

```sh
msrv=$(cargo metadata --no-deps --format-version 1 \
      | jq -r '.packages[0].rust_version')
```

That way the workflow stays in sync if the project’s `rust-version` changes.

joshka:
Good catch. I’ve changed the Justfile to derive the MSRV from `Cargo.toml`.

I used Just’s built-in file reading and regex support rather than `cargo metadata | jq`, which avoids adding `jq` as another cross-platform prerequisite. It also fails with a clear error if `rust-version` is missing.

I verified that it resolves to `1.85.0` and that both MSRV checks pass.

xtqqczze:
I’m not particularly keen on using regex for this, as it would break if `rust-version` is changed to be inherited from the workspace, for example:

```toml
rust-version.workspace = true
```

But I guess this is the best solution for now, unless we install a dependency: (`cargo install --locked jaq`)?

joshka:
Sometimes simple is best. I'm fine with taking the pragmatic approach here and not being overly defensive about things that can break that are unlikely to. The workspace version would still be defined at the root Cargo.toml file, so wouldn't cause that problem.

I did think about an install just task, but I mainly didn't want the hassle of working that out for win/linux/macos to do the right thing in each case. Docs are likely good enough for that.

xtqqczze:
[`cargo-hack`](https://github.com/taiki-e/cargo-hack) has a `--rust-version` option.

PR #794, opened 2023-07-11
Remove `IsTty` now that `std::io::IsTerminal` has stabalized.

gibbz00:
Wasn't sure if the example/is_tty should remain, simply adapted it for now.

PR #1086, opened 2026-07-29
Remove IsTty trait

xtqqczze:
Closes https://github.com/crossterm-rs/crossterm/pull/794


joshka:
It's worth noting in the changelog why and what the replacement is - obv. the built in has been around for a long time, but it's handy generally to ensure there's a replacement mentioned in docs that people would notice.

joshka:
thanks

xtqqczze:
This is slightly complicated by the fact that `IsTty` was not a sealed trait (see https://github.com/uutils/coreutils/pull/13638), meaning downstream users could have provided their own implementations. However, I think removing it is still the right decision.

Issue #996, opened 2025-07-06
Make kqueue work with tty device on macOS


zhscn:
# TL;DR

`kqueue()` does not work with `/dev/tty` on macOS, but it works with `STDIN_FILENO` inherited from the parent shell process and the specific controlling terminal of the current process (e.g. `/dev/ttys000`). This behavior can be tested using this code:

<details>
<summary>test.c</summary>

```c
#define BSD_SOURCE

#include <errno.h>
#include <fcntl.h>
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <sys/event.h>
#include <sys/stat.h>
#include <sys/sysctl.h>
#include <unistd.h>

void test_kqueue_on_fd(int fd, const char* name) {
  struct stat st;
  fstat(fd, &st);

  printf("Testing %s (fd=%d):\n", name, fd);
  printf("  Device type: ");
  if (S_ISCHR(st.st_mode)) {
    printf("character device (major=%d, minor=%d)\n", major(st.st_rdev),
           minor(st.st_rdev));
  } else if (S_ISFIFO(st.st_mode)) {
    printf("FIFO\n");
  } else {
    printf("other\n");
  }

  int kq = kqueue();
  if (kq == -1) {
    perror("kqueue");
    return;
  }

  struct kevent ev;
  EV_SET(&ev, fd, EVFILT_READ, EV_ADD, 0, 0, NULL);

  errno = 0;
  int ret = kevent(kq, &ev, 1, NULL, 0, NULL);

  printf("  kqueue ret: %d, ", ret);
  if (ret == -1) {
    printf("error: %s (errno=%d)\n\n", strerror(errno), errno);
  } else {
    printf("success!\n\n");
  }

  close(kq);
}

int main() {
  // 1. test STDIN_FILENO
  test_kqueue_on_fd(STDIN_FILENO, "STDIN_FILENO");

  // 2. test /dev/tty
  int tty_fd = open("/dev/tty", O_RDWR);
  if (tty_fd == -1) {
    perror("open /dev/tty");
  } else {
    test_kqueue_on_fd(tty_fd, "/dev/tty");
    close(tty_fd);
  }

  // 3. test controlling terminal of the current process
  pid_t pid = getpid();
  struct kinfo_proc proc_info;
  size_t len = sizeof(proc_info);
  int mib[4] = {CTL_KERN, KERN_PROC, KERN_PROC_PID, pid};
[...]

PR #1087, opened 2026-07-29
Remove the libc backend

### LINKED ISSUE #905: Cannot build with `use-dev-tty` feature
**Describe the bug**

When enabling with `use-dev-tty` feature, this crate does not build because it is missing the `rustix/process` feature:

```
error[E0433]: failed to resolve: could not find `process` in `rustix`
   --> src/event/source/unix/tty.rs:74:40
    |
74  |                 pipe::register(rustix::process::Signal::Winch as i32, sender)?;
    |                                        ^^^^^^^ could not find `process` in `rustix`
    |
note: found an item that was configured out
   --> /home/gyscos/.cargo/registry/src/index.crates.io-6f17d22bba15001f/rustix-0.38.34/src/lib.rs:240:9
    |
240 | pub mod process;
    |         ^^^^^^^
    = note: the item is gated behind the `process` feature
```

**To Reproduce**
Steps to reproduce the behavior (from this repository):
```
cargo check --features use-dev-tty
```

**Expected behavior**
Successful compilation.

**Workaround**
Manually adding the `process` feature of the `rustix` crate makes this crate compile again.

**OS**
 - Linux

**Terminal/Console**
 - Doesn't matter

### LINKED ISSUE #935: Cannot build crossterm 28.1 with use-dev-tty + event-stream features
When enabling both use-dev-tty feature and event-stream feature, latest crossterm 28.1 fails to build. This is a complete example of how to reproduce the build error:
```
~ ❯ cargo init crosstermtest && cd crosstermtest
    Creating binary (application) package
note: see more `Cargo.toml` keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
~/crosstermtest master* ❯ cargo add crossterm --features use-dev-tty,event-stream
    Updating crates.io index
      Adding crossterm v0.28.1 to dependencies
             Features:
             + bracketed-paste
             + event-stream
             + events
             + filedescriptor
             + use-dev-tty
             + windows
             - libc
             - serde
    Updating crates.io index
     Locking 44 packages to latest compatible versions
      Adding hermit-abi v0.3.9 (latest: v0.4.0)
      Adding linux-raw-sys v0.4.14 (latest: v0.6.5)
      Adding wasi v0.11.0+wasi-snapshot-preview1 (latest: v0.13.3+wasi-0.2.2)
      Adding windows-sys v0.52.0 (latest: v0.59.0)
~/crosstermtest master* ❯ cat Cargo.toml | grep crossterm
name = "crosstermtest"
crossterm = { version = "0.28.1", features = ["use-dev-tty", "event-stream"] }
~/crosstermtest master* ❯ cargo build
   Compiling libc v0.2.159
   Compiling proc-macro2 v1.0.87
   ... (cut)
   Compiling filedescriptor v0.8.2
   Compiling crossterm v0.28.1
error[E0599]: no variant or associated item named `new` found for enum `FileDesc` in the current scope
   --> /home/da/.cargo/registry/src/index.crates.io-6f17d22bba15001f/crossterm-0.28.1/src/event/source/unix/tty.rs:188:36
    |
188 |                 let fd = FileDesc::new(self.wake_pipe.receiver.as_raw_fd(), false);
    |                                    ^^^ variant or associated item not found in `FileDesc<'_>`
    |
   ::: /home/da/.cargo/registry/src/index.crates.io-6f17d22bba15001f/crossterm-0.28.1/src/terminal/sys/file_descriptor.rs:30:1
    |
30  | pub enum FileDesc<'a> {
    | --------------------- variant or associated item `new` not found for this enum
    |
    = help: items from traits can only be used if the trait is implemented and in scope
    = note: the following trait defines an item `new`, perhaps you need to implement it:
            candidate #1: `parking_lot_core::thread_parker::ThreadParkerT`

For more information about this error, try `rustc --explain E0599`.
error: could not compile `crossterm` (lib) due to 1 previous error
```

joshka:
## Summary

- Remove the optional `libc` dependency and its implicit Cargo feature.
- Use rustix as the only Unix terminal backend.
- Remove the duplicate file-descriptor, terminal-mode, window-size, and signal paths.
- Update the changelog and dependency documentation.

## Compatibility and deprecation context

Rustix became the default backend in [#892](https://github.com/crossterm-rs/crossterm/pull/892), closing [#847](https://github.com/crossterm-rs/crossterm/issues/847). The libc implementation was deliberately retained as a temporary compatibility escape hatch while the new backend gained real-world coverage. The original discussion explicitly proposed removing that escape hatch in the following breaking release if the migration went well ([#878 comment](https://github.com/crossterm-rs/crossterm/pull/878#issuecomment-2111901696)).

The Cargo feature was not formally annotated as deprecated, but it has served that temporary fallback role since rustix became the default. This PR completes that migration in the next breaking release.

The known rustix-specific Crossterm regressions have been resolved:

- WSL and Android rejected `TCGETS2`; rustix added a legacy `TCGETS` fallback and Crossterm adopted the fixed release in [#926](https://github.com/crossterm-rs/crossterm/pull/926), resolving [#912](https://github.com/crossterm-rs/crossterm/issues/912).
- The `use-dev-tty` feature initially omitted rustix's `process` feature; [#906](https://github.com/crossterm-rs/crossterm/pull/906) resolved [#905](https://github.com/crossterm-rs/crossterm/issues/905).
- The `use-dev-tty + event-stream` combination initially failed to compile, with libc serving as a workaround; [#955](https://github.com/crossterm-rs/crossterm/pull/955) resolved [#935](https://github.com/crossterm-rs/crossterm/issues/935) and added that combination to CI.

The only meaningful library found forwarding `crossterm/libc` is Reedline. It added the feature in [reedline#1014](https://github.com/nushell/reedline/pull/1014) for emulators that did not implement `TCGETS2`/`TCSETS2` or returned a nonstandard error instead of allowing rustix's fallback. That motivating compatibility gap has since closed:

- Rustix has fallen back from `TCGETS2`/`TCSETS2` to the legacy ioctls for standard unsupported-call errors since [rustix#1147](https://github.com/bytecodealliance/rustix/pull/1147).
- CheerpX 1.2.7 implemented the missing calls, resolving [WebVM #214](https://github.com/leaningtech/webvm/issues/214#issuecomment-3907814619).
- QEMU implemented termios2 support in [qemu@e9a8a10](https://github.com/qemu/qemu/commit/e9a8a10e84c1bf6e2e8be000e4dd5c83ba0d8470), and the corresponding [Ubuntu QEMU bug](https://bugs.launchpad.net/ubuntu/+source/qemu/+bug/2133804) is Fix Released, including Jammy and Noble.

The residual compatibility change is intentional: obsolete emulator builds that neither implement the newer ioctls nor return a fallback-compatible error can no longer select Crossterm's libc backend. Downstreams forwarding or enabling `crossterm/libc` must remove that feature when updating to the next breaking Crossterm release. Existing Crossterm 0.29 dependency resolutions are unaffected.

This removes Crossterm's direct libc backend; it does not promise that the `libc` crate disappears transitively from every dependency graph.

## Edge-case audit

I reviewed every Crossterm issue and pull request mentioning libc or rustix. The remaining open Unix terminal issues concern TTY selection, redirected standard streams, macOS `/dev/tty` polling, and competing readers—not differences between the libc and rustix wrappers. In particular, [#396](https://github.com/crossterm-rs/crossterm/issues/396), [#692](https://github.com/crossterm-rs/crossterm/issues/692), [#828](https://github.com/crossterm-rs/crossterm/issues/828), [#996](https://github.com/crossterm-rs/crossterm/issues/996), and [#957](https://github.com/crossterm-rs/crossterm/pull/957) do not require retaining the duplicate backend.

Open pull requests that currently contain parallel libc/rustix paths will need to rebase and retain only their rustix implementation, but they do not expose an unresolved compatibility blocker.

## Validation

- Formatting, Clippy with all targets/features, docs, and doctests
- Full tests with all targets/features
- Rust 1.85 checks with default and all features
- Supported Unix feature combinations, including `events + event-stream + use-dev-tty`
- Package verification and dependency policy
- Workflow validation with actionlint and zizmor

The required GitHub matrix will additionally run the suite on Linux, Windows, and macOS.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Remove `IsTty` trait.
   Use the standard library's [`std::io::IsTerminal`](https://doc.rust-lang.org/std/io/trait.IsTerminal.html) trait instead,
   which provides equivalent functionality.
+- Remove the `libc` feature and always use rustix for Unix terminal operations.
 
 ## Changed ⚙️
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -210,7 +210,6 @@ dependencies = [
  "futures",
  "futures-core",
  "futures-timer",
- "libc",
  "mio",
  "parking_lot",
  "rustix",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -72,9 +72,6 @@ winapi = { version = "0.3.9", optional = true, features = ["winuser", "winerror"
 # UNIX dependencies
 [target.'cfg(unix)'.dependencies]
 filedescriptor = { version = "0.8", optional = true }
-# Default to using rustix for UNIX systems, but provide an option to use libc for backwards
-# compatibility.
-libc = { version = "0.2", default-features = false, optional = true }
 mio = { version = "1.0", features = ["os-poll"], optional = true }
 rustix = { version = "1", default-features = false, features = ["std", "stdio", "termios"] }
 signal-hook = { version = "0.3.17", optional = true }
--- a/README.md
+++ b/README.md
@@ -166,7 +166,7 @@ This can disable `mio` / `signal-hook` / `signal-hook-mio` dependencies.
 |:---------------|:---------------------------------------------------------------------------------|:--------------------------------------|
 | `bitflags`     | `KeyModifiers`, those are differ based on input.                                 | always                                |
 | `parking_lot`  | locking `RwLock`s with a timeout, const mutexes.                                 | always                                |
-| `libc`         | UNIX terminal_size/raw modes/set_title and several other low level functionality. | optional (`events` feature), UNIX only |
+| `rustix`       | UNIX terminal size/raw modes and other low-level functionality.                   | UNIX only                              |
 | `Mio`          | event readiness polling, waking up poller                                        | optional (`events` feature), UNIX only |
 | `signal-hook`  | signal-hook is used to handle terminal resize SIGNAL with Mio.                   |  optional (`events` feature),UNIX only |
 | `winapi`       | Used for low-level windows system calls which ANSI codes can't replace           | windows only                          |
--- a/src/event/source/unix/tty.rs
+++ b/src/event/source/unix/tty.rs
@@ -1,8 +1,5 @@
-#[cfg(feature = "libc")]
-use std::os::unix::prelude::AsRawFd;
 use std::{collections::VecDeque, io, os::unix::net::UnixStream, time::Duration};
 
-#[cfg(not(feature = "libc"))]
 use rustix::fd::{AsFd, AsRawFd};
 
 use signal_hook::low_level::pipe;
@@ -68,9 +65,6 @@ impl UnixInternalEventSource {
             winch_signal_receiver: {
                 let (receiver, sender) = nonblocking_unix_pair()?;
                 // Unregistering is unnecessary because EventSource is a singleton
-                #[cfg(feature = "libc")]
-                pipe::register(libc::SIGWINCH, sender)?;
-                #[cfg(not(feature = "libc"))]
                 pipe::register(rustix::process::Signal::WINCH.as_raw(), sender)?;
                 receiver
             },
@@ -163,9 +157,6 @@ impl EventSource for UnixInternalEventSource {
                 }
             }
             if fds[1].revents & POLLIN != 0 {
-                #[cfg(feature = "libc")]
-                let fd = FileDesc::new(self.winch_signal_receiver.as_raw_fd(), false);
-                #[cfg(not(feature = "libc"))]
                 let fd = FileDesc::Borrowed(self.winch_signal_receiver.as_fd());
                 // drain the pipe
                 while read_complete(&fd, &mut [0; 1024])? != 0 {}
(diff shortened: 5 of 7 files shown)


xtqqczze:
`rustix` does in fact support a libc backend via `--cfg=rustix_use_libc`. It may be worth documenting this option in case users need it for `LD_PRELOAD`-related use cases.

orhun (approved):
yay

xtqqczze:
It's probably worth keeping the `libc` feature, since `rustix` has a `use-libc` feature which uses a `libc` backend internally. So we can remove the backend from `crossterm` and use something like:

```toml
[features]
libc = ["rustix/use-libc"]
```

joshka:
> `rustix` does in fact support a libc backend via `--cfg=rustix_use_libc`. It may be worth documenting this option in case users need it for `LD_PRELOAD`-related use cases.

I did a github search of things using our libc feature - there's not enough to make it worth worrying about IMO. Can you find any use cases for this that matter?

xtqqczze:
I see this was discussed in https://github.com/crossterm-rs/crossterm/pull/1087.

Would it make sense to merge this with a `libc = ["rustix/use-libc"]` feature for now? We can always remove it before the next release.

xtqqczze:
> I did a github search of things using our libc feature - there's not enough to make it worth worrying about IMO. Can you find any use cases for this that matter?

In theory, removing the feature could remove an upgrade path for dependent crates currently using crossterm 0.27 (which doesn't use rustix).

joshka:
Thanks. I checked this more closely, and I don’t think retaining the feature provides meaningful compatibility.

Crossterm 0.27 had no `libc` feature. Libc was unconditional, so there is no Cargo feature upgrade path to preserve. Users who need `LD_PRELOAD` can select rustix’s libc backend with `--cfg=rustix_use_libc`, even when rustix is a transitive dependency.

The feature also wouldn’t preserve the original Reedline/VM workaround, since rustix’s libc backend still tries `TCGETS2` first. The affected VMs have since added support for it.

My main reason for keeping the libc implementation was the risk of switching fully to rustix when it was relatively unfamiliar. The releases since then have given us enough confidence that it works. I’m happy to reevaluate or revert if that proves wrong, but my preference is to remove the feature cleanly.

PR #1102, opened 2026-08-03
Reduce MSRV to 1.75

xtqqczze:
https://github.com/crossterm-rs/crossterm/pull/1059 unnecessarily raised the declared MSRV. This reduces it to Rust 1.75 while keeping the dependencies and implementation compatible.

Also remove an unnecessary target-platform-specific `windows` feature and associated `compile_error!`, simplifying the configuration without affecting functionality.

MSRV CI workflow now verifies builds against minimal dependency versions.


xtqqczze:
```console
$ cargo minimal-versions check -q --rust-version
info: running `cargo update -Z minimal-versions`
...
info: running `cargo hack check -q --rust-version`
info: running `rustup run 1.75 cargo check -q` on crossterm (1/1)
error[E0658]: mutable references are not allowed in constant functions
  --> src/style/styled_content.rs:51:28
   |
51 |     pub const fn style_mut(&mut self) -> &mut ContentStyle {
   |                            ^^^^^^^^^
   |
   = note: see issue #57349 <https://github.com/rust-lang/rust/issues/57349> for more information
```

Blame: https://github.com/crossterm-rs/crossterm/commit/34636cf8149c1e7252f778e24ad1d0ade0f1a1ca

joshka:
I don't think this is a good idea without a rationale that meaningfully says 1.75 is a better version to collapse on. Two concerns:

Rust 1.85 is the threshold for the 2024 edition, which is where the formatting changes come in. It's supported by debian stable and ubuntu, which is one of the general concerns over choosing an MSRV that works. It was released Feb 2025, so it's reasonably long enough for most use cases.

Pinning a dependency to a specific version makes us responsible for understanding every future change and how it impacts whether we should update. This adds maintenance burden.

Because these two negatives outweigh any obvious direct benefits of choosing 1.75 that I'm aware of, I'd tend to say this is not worth doing. Are there arguments for this that I'm missing? Do you have a specific concrete use case that you can't update your compiler, or is this just a general conservative take?

PR #1111, opened 2026-08-19
feat: add `MouseEventKind::get_button()`

Bishops-exe:


src/event.rs
```diff
@@ -799,6 +799,18 @@ pub enum MouseEventKind {
...
 }
 
+impl MouseEventKind {
+    pub fn get_button(&self) -> Option<MouseButton> {
+        match self {
+            MouseEventKind::Down(button)
+            | MouseEventKind::Up(button)
+            | MouseEventKind::Drag(button) => Some(*button),
+
+            _ => None,
```
orhun, line 812:
```suggestion
            | MouseEventKind::Drag(button) => Some(*button),
            _ => None,
```

src/event.rs
```diff
@@ -799,6 +799,18 @@ pub enum MouseEventKind {
     ScrollRight,
 }
 
+impl MouseEventKind {
+    pub fn get_button(&self) -> Option<MouseButton> {
```
orhun: Would be nice to add a doc comment

orhun (approved):

src/event.rs
```diff
@@ -799,6 +799,21 @@ pub enum MouseEventKind {
     ScrollRight,
 }
 
+impl MouseEventKind {
+    /// Returns the mouse button connected with the event.
+    ///
+    /// This will return `Some(MouseButton)` for `Down`, `Up`, and `Drag` events.
+    /// For other events, like `Moved` or `Scroll...`, it returns `None`.
+    pub fn get_button(&self) -> Option<MouseButton> {
```
joshka: naming is non idiomatic rust. should probably be just `button()`. We do have `as_...()` methods in other types (Event, KeyEvent, KeyCode, KeyEventKind), but I think this isn't so much a conversion and is more a give me the associated button to this event if one exists type method, so probably shouldn't have an `as_` prefix. Not 100% sure on this, but I'm definite about not having a `get_` prefix.

joshka (changes requested):
100% support on the idea - it's a good one, less so on the name.

src/event.rs
Bishops-exe: I agree with that it should not have been a `get_...` function. but from looking into crossterm further i have seen that other functions with similar behavior have been named with `as_...`

joshka: yes, I know - I added those... :) and I don't think as_ is correct for this use (but not 100% certain), so probably just button() here. Look at https://rust-lang.github.io/api-guidelines/naming.html#c-conv and https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter for some guidance on this.

Bishops-exe: ok i fixed it

PR #1112, opened 2026-08-20
fix(cursor): clamp Windows legacy-console cursor moves at the buffer edge

### LINKED ISSUE #1094: Windows legacy console: `move_up`/`move_left` underflow when count exceeds the cursor position
## Summary

`cursor::sys::windows::move_up` and `move_left` subtract the caller's `count` from the current
cursor row/column with a bare `u16` subtraction and no floor. If `count` exceeds the current row
or column the subtraction underflows: **panic in a debug build, wrap to ~65535 in a release
build.**

This is the legacy-console path only (reached through `execute_winapi`, i.e. when `supports_ansi`
is false), so it is Windows-only and does not affect the ANSI path.

## The sites

`src/cursor/sys/windows.rs` on the default branch:

```rust
pub(crate) fn move_up(count: u16) -> std::io::Result<()> {
    let (column, row) = position()?;
    move_to(column, row - count)?;          // :59  -- underflows when count > row
    Ok(())
}

pub(crate) fn move_left(count: u16) -> std::io::Result<()> {
    let (column, row) = position()?;
    move_to(column - count, row)?;          // :77  -- underflows when count > column
    Ok(())
}
```

`MoveUp` and `MoveLeft` in `src/cursor.rs` reach these through `execute_winapi` with the caller's
`count` unchecked, so `queue!(w, MoveUp(5))` with the cursor on row 2 is sufficient to trigger it.

In release the wrapped value then feeds `move_to`, which casts to `i16`, sees a negative, and
returns an `Argument Out of Range` error — so the release outcome degrades to an error rather than
a bogus cursor move. That is better than a silent wrong position, but it is still not the intended
behaviour, and debug builds panic outright.

## Why this reads as an oversight rather than a decision

The analogous case is guarded two impls away in the same file. `MoveToNextLine` and
`MoveToPreviousLine` both test their count before calling into the sys layer, while `MoveUp`,
`MoveDown`, `MoveLeft` and `MoveRight` call theirs unguarded. The author was thinking about the
edge in one place and not the neighbouring one.

## Severity, honestly bounded

Not memory-unsafe: `u16` subtraction panics or wraps, it does not read out of bounds. The trigger
is a caller passing a `count` larger than the current position, which is an ordinary thing for
application code to do near the top or left edge of the screen. No untrusted input is involved.

## Suggested fix

`saturating_sub` matches what the crate already chose for the coordinate parsers in
`src/event/sys/unix/parse.rs`, and clamping at the edge is the natural terminal behaviour:

```rust
    move_to(column, row.saturating_sub(count))?;   // move_up
    move_to(column.saturating_sub(count), row)?;   // move_left
```

`move_down` and `move_right` add rather than subtract, so they are not affected by this, though a
matching `saturating_add` would be consistent.

## Scope note — the parser half of this report is already fixed

This finding originally covered eight bare `- 1` subtractions in four coordinate parsers in
`src/event/sys/unix/parse.rs`. **Checked against the default branch before filing: those are all
fixed** — the parsers now use `saturating_sub(1)`, `parse_csi_normal_mouse` folds the protocol
offset into a single `saturating_sub(33)`, and there is a regression test
(`test_parse_csi_normal_mouse_zero_coords_clamp`) naming the old behaviour. Nothing is being
re-reported there; this issue is narrowed to the two Windows sites, which are unchanged.

## Provenance

Found by a line-by-line read of crossterm 0.28.1 during a `cargo vet` supply-chain certification,
then re-checked against the default branch immediately before filing. Not observed on a live legacy
Windows console — the underflow is reasoned from the source, and the release-path behaviour is
derived from `move_to`'s `i16` cast rather than measured.

hdimer:
Closes #1094.

## The bug

`move_up`, `move_left` and `move_to_previous_line` in `src/cursor/sys/windows.rs`
subtract the caller's `count` from the current row/column with a bare `u16`
subtraction and no floor:

```rust
move_to(column, row - count)?;      // move_up
move_to(column - count, row)?;      // move_left
move_to(0, row - count)?;           // move_to_previous_line
```

When `count` exceeds the current row/column that underflows: `attempt to subtract
with overflow` in a debug build, or a wrap to a large `u16` in release. This is the
legacy-console path only, reached through `execute_winapi` when `supports_ansi` is
false, so it does not affect the ANSI path.

`MoveUp(n)` with `n` greater than the current row is not really a caller error, it
is the ordinary "get back to the top of the block I just drew" idiom, and it does
the right thing everywhere except here.

## Scope: three sites, not the two in the issue

#1094 names `move_up` and `move_left` and says `move_to_previous_line` is already
guarded. It isn't. `MoveToPreviousLine::execute_winapi` in `src/cursor.rs` only
checks `if self.0 != 0`, which skips the zero case, not the `count > row` case, so
`sys::move_to_previous_line` underflows the same way. All three are fixed here.

One correction to the issue while I'm at it: the release-mode outcome is not
uniformly the `Argument Out of Range` error it describes. That happens for counts
in `1..=32768`, where the wrapped value casts to a negative `i16` and trips the
guard in `ScreenBufferCursor::move_to`. A larger count wraps to a *positive*
`i16`, sails past the guard, and fails later against the real buffer bounds with a
different OS error. Either way it is wrong, but the error you get varies.

## Why clamp

`execute_winapi` exists to emulate the ANSI sequence on legacy conhost, so the ANSI
branch is the spec. `MoveUp`/`MoveLeft`/`MoveToPreviousLine` emit CUU/CUB/CPL, which
terminals clamp at the top row and first column. Returning an error would make the
WinAPI path diverge from the branch it is imitating, and doing nothing would be
wrong on its own terms (moving up 5 from row 2 should land on row 0, not stay on
row 2). `saturating_sub` is also the primitive the crate already picked for this
exact bug class in `src/event/sys/unix/parse.rs`.

## What is deliberately not in here

`move_down`, `move_right` and `move_to_next_line` add rather than subtract, so they
can overflow rather than underflow. I left them alone on purpose: `saturating_add`
would clamp at `u16::MAX`, which the `as i16` cast turns straight back into an
error, so it converts a panic into a different failure rather than fixing anything.
A real fix there needs the screen buffer bounds and a decision about whether to
clamp to the buffer or the window, which is a separate question from this one.

## Tests

The three existing tests moved back by the exact count (`move_to(0, 2)` then
`move_up(2)`), so they landed on zero without ever crossing it and never touched
the bug. Each now also overshoots. The overshoot deliberately starts from a
non-zero position rather than from the origin: starting at zero, "clamp at the
edge" and "do nothing" are indistinguishable, so such a test would still pass if
the fix were later replaced with `checked_sub(count).unwrap_or(row)`. Starting two
rows down and asking for five pins the clamp itself.

## Verification

This is `cfg(windows)` code needing a real console handle, and I am on macOS, so
the three tests here run on your Windows CI, not on my machine. What I did check
locally:

- `cargo clippy --locked --target x86_64-pc-windows-msvc --all-targets --all-features -- -D warnings`, clean. I confirmed `--all-targets` really does type-check the `#[cfg(test)]` module for that target by injecting a deliberate type error into one of the new test lines and watching it fail.
- `rustup run 1.85.0 cargo check --locked --lib --all-features --target x86_64-pc-windows-msvc`, clean, plus both host MSRV configurations.
- Host side: `cargo fmt --check`, host clippy, `cargo test --locked --all-targets --all-features -- --test-threads 1` (121 passed), doc tests, `cargo doc` with `RUSTDOCFLAGS=-D warnings`, `cargo package`.
- To get an actual red/green cycle out of unrunnable code, I mirrored the three
  functions over a stub cursor and ran the new assertions against three
  implementations: the current bare subtraction fails all three (panic in debug,
  error in release), `checked_sub(..).unwrap_or(..)` fails all three, and
  `saturating_sub` passes in both profiles.

I did not run `cargo deny`, `actionlint` or `zizmor`; this touches no dependency
and no workflow file.

I used an AI assistant while working on this. The reasoning above, the scope
correction and the test design are mine and I have checked them; the diff is small
enough to read in a minute either way.
---
Disclosure: this change was prepared with AI assistance (Claude Code). The repro and tests described above were run before it was opened.

### DIFF
--- a/src/cursor/sys/windows.rs
+++ b/src/cursor/sys/windows.rs
@@ -32,6 +32,15 @@ pub fn parse_relative_y(y: i16) -> std::io::Result<i16> {
     }
 }
 
+/// Returns the last addressable cell of the screen buffer (column, row).
+///
+/// Relative moves clamp here rather than at the window edge: `SetConsoleCursorPosition` takes
+/// screen buffer coordinates, and the buffer can be larger than the window.
+fn last_cell() -> std::io::Result<(u16, u16)> {
+    let (width, height): (u16, u16) = ScreenBuffer::current()?.info()?.buffer_size().into();
+    Ok((width.saturating_sub(1), height.saturating_sub(1)))
+}
+
 /// Returns the cursor position (column, row).
 ///
 /// The top left cell is represented `0,0`.
@@ -62,13 +71,15 @@ pub(crate) fn move_up(count: u16) -> std::io::Result<()> {
 
 pub(crate) fn move_right(count: u16) -> std::io::Result<()> {
     let (column, row) = position()?;
-    move_to(column + count, row)?;
+    let (last_column, _) = last_cell()?;
+    move_to(column.saturating_add(count).min(last_column), row)?;
     Ok(())
 }
 
 pub(crate) fn move_down(count: u16) -> std::io::Result<()> {
     let (column, row) = position()?;
-    move_to(column, row + count)?;
+    let (_, last_row) = last_cell()?;
+    move_to(column, row.saturating_add(count).min(last_row))?;
     Ok(())
 }
 
@@ -92,7 +103,8 @@ pub(crate) fn move_to_row(new_row: u16) -> std::io::Result<()> {
 
 pub(crate) fn move_to_next_line(count: u16) -> std::io::Result<()> {
     let (_, row) = position()?;
-    move_to(0, row + count)?;
+    let (_, last_row) = last_cell()?;
+    move_to(0, row.saturating_add(count).min(last_row))?;
     Ok(())
 }
 
@@ -236,6 +248,11 @@ mod tests {
         let (saved_x, saved_y) = position().unwrap();
         move_right(1).unwrap();
         assert_eq!(position().unwrap(), (saved_x + 1, saved_y));
+
+        // Overshooting lands on the last column of the buffer. Its exact index depends on the
+        // machine, so this asserts the move happened rather than a fixed position.
+        move_right(u16::MAX).unwrap();
+        assert!(position().unwrap().0 > saved_x + 1);
     }
 
     #[test]
@@ -284,6 +301,11 @@ mod tests {
         move_to_next_line(2).unwrap();
 
         assert_eq!(position().unwrap(), (0, 4));
+
+        // Overshooting lands on the last row of the buffer. Its exact index depends on the
+        // machine, so this asserts the move happened rather than a fixed position.
+        move_to_next_line(u16::MAX).unwrap();
+        assert!(position().unwrap().1 > 4);
     }
 
     #[test]
@@ -338,6 +360,11 @@ mod tests {
         move_down(2).unwrap();
 
         assert_eq!(position().unwrap(), (0, 2));
+
+        // Overshooting lands on the last row of the buffer. Its exact index depends on the
+        // machine, so this asserts the move happened rather than a fixed position.
+        move_down(u16::MAX).unwrap();
+        assert!(position().unwrap().1 > 2);
     }
 
     #[test]
--- a/src/cursor/sys/windows.rs
+++ b/src/cursor/sys/windows.rs
@@ -56,7 +56,7 @@ pub(crate) fn move_to(column: u16, row: u16) -> std::io::Result<()> {
 
 pub(crate) fn move_up(count: u16) -> std::io::Result<()> {
     let (column, row) = position()?;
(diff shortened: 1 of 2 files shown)


src/cursor/sys/windows.rs
joshka: This seems like it would similarly overflow - same move_down.

joshka (commented):
Thanks for the PR. I understand there's concerns about i16 casts and that's why you skipped the move_right/down fixes. I think it's worth getting this right in a single PR rather than two PRs or skipping the overflow condition here.

Would you mind taking a look and expanding on this?

PR #1113, opened 2026-08-20
Replace derive_more event helpers with a local macro

joshka:
## Summary

This removes the `derive_more` dependency and feature used to provide the event enum `is_*` methods.

The public methods and their behavior remain unchanged. They are now generated by a small private `macro_rules!` macro, while payload-sensitive predicates remain handwritten.

This follows the direction discussed in [PR #949](https://github.com/crossterm-rs/crossterm/pull/949), where the tradeoff between boilerplate and an additional proc-macro dependency was discussed. [PR #970](https://github.com/crossterm-rs/crossterm/pull/970) subsequently made the dependency optional, but it was still enabled by default.

## Why

The generated methods are all simple variant checks, so requiring downstream users to compile `derive_more` and its proc-macro dependencies is disproportionate to the behavior being added.

The local macro keeps the declarations concise and consistent:

```rust
impl_is_variant! {
    Event {
        FocusGained => is_focus_gained,
        Key(..) => is_key,
        Mouse(..) => is_mouse,
    }
}
```

The variants previously marked `is_variant(ignore)`—`KeyCode::F`, `Char`, `Media`, and `Modifier`—are intentionally omitted from the macro. They already have handwritten payload-aware methods: `is_function_key`, `is_char`, `is_media_key`, and `is_modifier`. This avoids duplicate or conflicting methods while preserving their value-sensitive behavior.

## Compile-time comparison

These were clean, isolated builds on the same machine and toolchain:

| Benchmark | Local macro | `derive_more` | Difference |
| --- | ---: | ---: | ---: |
| Clean crossterm `cargo build` | 2.41s | 3.41s | +1.00s |
| Fresh consumer `cargo check` | 2.12s | 3.00s | +0.88s |

The `derive_more` build additionally compiled proc-macro dependencies including `syn`, `quote`, `convert_case`, `derive_more-impl`, and `derive_more`.

For context, [PR #949](https://github.com/crossterm-rs/crossterm/pull/949) measured the original fresh-project comparison at 1.851s versus 3.539s, and a clean M2 crate build at 2.016s versus 2.688s. Exact timings vary by toolchain and machine, but the direction is consistent.

## Validation

- `cargo fmt --check`
- `cargo test --all-targets --locked`
- `cargo test --no-default-features --all-targets --locked`
- `cargo test --all-features --all-targets --locked`
- `cargo clippy --all-targets --all-features --locked -- -D warnings`
- `cargo doc --no-deps --all-features --locked`

cc @archseer @the-mikedavis

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 - Remove `IsTty` trait.
   Use the standard library's [`std::io::IsTerminal`](https://doc.rust-lang.org/std/io/trait.IsTerminal.html) trait instead,
   which provides equivalent functionality.
+- Remove the `derive-more` feature and `derive_more` dependency. Event variant helper
+  methods are always available.
 
 ## Changed ⚙️
 
@@ -33,7 +35,6 @@
 - Derive standard traits for "SetCursorStyle" (#909)
 - Add query_keyboard_enhancement_flags to read enabled flags (#958)
 - Add is_* and as_* methods to the event enums (#949)
-- Add a feature flag for derive_more impls (#970)
 - Update rustix to 1.0 (#982)
 
 ## Breaking ⚠️
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -183,15 +183,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "convert_case"
-version = "0.10.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -33,7 +35,6 @@
 - Derive standard traits for "SetCursorStyle" (#909)
 - Add query_keyboard_enhancement_flags to read enabled flags (#958)
 - Add is_* and as_* methods to the event enums (#949)
-- Add a feature flag for derive_more impls (#970)
```
cammeresi, line 36: I think this line was deleted by mistake.

